PR #110, opened 2012-11-01
Add encoding comment to source files

Zearin:
This is a minor thing.  It’s a convention, not a rule (obviously)—but it’s also a best practice.

Many text editors look for this kind of comment as the first line—or second line, if there is a shebang (the `#!/usr/bin/env python`, or whatever else)—and make use of this to determine the file’s encoding.


Diwasdrpm:
...

PR #289, opened 2014-12-23
Add optional context when creating a status for a commit

andycasey:
Hi,

This pull request allows optional context to be given when creating a status for a commit.

From https://developer.github.com/v3/repos/statuses/#create-a-status the ``context`` is a string label to differentiate the current status from the status of other systems. The default is "default", which means any status created through ``v1.x`` will always have a context of "default". Even though the API specifies the default context to be "default", I have matched the code based for existing optional arguments by using ``github.GithubObject.NotSet``.

Cheers,
Andy

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,7 +17,7 @@ What's new?
 Version 1.25.2 (October 7th, 2014)
 ----------------------------------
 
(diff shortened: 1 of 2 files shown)


JordanMilne:
:+1:

jacquev6:
Please see #297.

andycasey:
:smile: :beers:

PR #290, opened 2014-12-27
Change InputGitAuthor to reflect that time is an optional parameter.

ndahlquist:
Admittedly, the documentation at https://developer.github.com/v3/git/commits/ is rather ambiguous (this change is bordering on undefined behavior). Despite that, this change works and is pretty convenient for me, so I thought I'd put it out there if you're interested.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@
 /MANIFEST
 /PyGithub.egg-info/
 /.coverage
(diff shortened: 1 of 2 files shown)


jacquev6:
Please see #297.

jzelinskie:
lgtm

PR #292, opened 2015-01-30
Added sort option to get_pulls.

jondb:
Valid option for get pulls: https://developer.github.com/v3/pulls/


jacquev6:
Please see #297.

PR #294, opened 2015-02-03
Allow access to preview versions of the server-side API

dpocock:
Developers can preview / test their code against new versions of the Github APIs by setting an Accept header

https://developer.github.com/changes/2014-12-08-organization-permissions-api-preview/

This change allows PyGithub users to try the new APIs by setting api_preview=True when constructing the Github object


jacquev6:
Please see #297.

Issue #172, opened 2013-06-08
can not get access to github API through proxy


linvictor88:
My environment is within in a local network with a http proxy, I wonder whether there is a simple method to get access the github API. Thank you very much for your help in advance

PR #309, opened 2015-04-15
Connect through a proxy with authentication

beugley:
Issue 172 is closed (https://github.com/PyGithub/PyGithub/issues/172), but was never resolved.  Since it's quite old, you may not be interested in this change.  But if you are, this change will allow PyGithub to connect through a proxy that requires authentication.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -326,12 +326,10 @@ class Requester:
         ## set.
         ## http_proxy: http://user:password@proxy_host:proxy_port
         ##
-        try:
-           proxy_uri = os.getenv('http_proxy')
-        except:
-           proxy_uri = None
-
-        if (proxy_uri != None):
+        proxy_uri = os.getenv('http_proxy')
+        if proxy_uri is None:
+            proxy_uri = os.getenv('HTTP_PROXY')
+        if proxy_uri is not None:
             url = urlparse.urlparse(proxy_uri)
             conn = self.__connectionClass(url.hostname, url.port, **kwds)
             headers = {}
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -326,9 +326,7 @@ class Requester:
         ## set.
         ## http_proxy: http://user:password@proxy_host:proxy_port
         ##
-        proxy_uri = os.getenv('http_proxy')
-        if proxy_uri is None:
-            proxy_uri = os.getenv('HTTP_PROXY')
+        proxy_uri = os.getenv('http_proxy') or os.getenv('HTTP_PROXY')
         if proxy_uri is not None:
             url = urlparse.urlparse(proxy_uri)
             conn = self.__connectionClass(url.hostname, url.port, **kwds)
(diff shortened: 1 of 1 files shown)


github/Requester.py
```diff
@@ -319,7 +320,29 @@ def __createConnection(self):
...
+        ## Connect through a proxy server with authentication, if http_proxy
+        ## set.
+        ## http_proxy: http://user:password@proxy_host:proxy_port
+        ##
+        try:
+           proxy_uri = os.getenv('http_proxy')
+        except:
+           proxy_uri = None
+
+        if (proxy_uri != None):
```
jzelinskie:
```python
if proxy_uri is not None:
```

github/Requester.py
```diff
@@ -319,7 +320,29 @@ def __createConnection(self):
...
             kwds["timeout"] = self.__timeout  # Did not exist before Python2.6
-        return self.__connectionClass(self.__hostname, self.__port, **kwds)
+
+        ##
+        ## Connect through a proxy server with authentication, if http_proxy
+        ## set.
+        ## http_proxy: http://user:password@proxy_host:proxy_port
+        ##
+        try:
+           proxy_uri = os.getenv('http_proxy')
```
jzelinskie: is this standard?

github/Requester.py
beugley: No, it's not standard.  I will change it to use "is not".

jzelinskie: [According to the knowledge card on Google](https://www.google.com/search?q=http_proxy), it is standard. Can we get it in all caps? Like this: `HTTP_PROXY`

github/Requester.py
beugley: http_proxy is respected by many apps as a way to specify a proxy server, I just used the existing convention.

github/Requester.py
beugley: On Linux, it's always in lower case.  How about if I add code to check for both lower and upper case env vars?

beugley: Please check again.  The code now uses http_proxy, then HTTP_PROXY.  If neither is set, it then connects without proxy.

github/Requester.py
```diff
@@ -319,7 +320,27 @@ def __createConnection(self):
...
         if atLeastPython26:  # pragma no branch (Branch useful only with Python 2.5)
             kwds["timeout"] = self.__timeout  # Did not exist before Python2.6
-        return self.__connectionClass(self.__hostname, self.__port, **kwds)
+
+        ##
+        ## Connect through a proxy server with authentication, if http_proxy
+        ## set.
+        ## http_proxy: http://user:password@proxy_host:proxy_port
+        ##
+        proxy_uri = os.getenv('http_proxy')
```
jzelinskie:
```python
proxy_uri = os.getenv('http_proxy') or os.getenv('HTTP_PROXY')
```

beugley: Done.

jzelinskie:
lgtm

PR #284, opened 2014-11-10
Add team membership add call

xorel:
This should fix https://github.com/jacquev6/PyGithub/issues/280

PR #313, opened 2015-04-25
Add add_membership to Team

domoritz:
Ref #284

### DIFF
--- a/github/Team.py
+++ b/github/Team.py
@@ -121,6 +121,18 @@ class Team(github.GithubObject.CompletableGithubObject):
             self.url + "/members/" + member._identity
         )
 
+    def add_membership(self, member):
+        """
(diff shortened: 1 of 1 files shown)


jzelinskie:
Isn't this already merged into the develop? We just need to merge the two and only work from master, because it seems like people just naturally submit their changes there and not develop.

domoritz:
Hmm, I didn't see the dev branch. Btw I noticed that the master branch has a more recent commit. Is master the latest version on pip?

jzelinskie:
No, it isn't. The maintainers on this project got shifted and inherited a bunch of PRs for all different branches. The project was supposed to be using the "git flow" development model, but we ended up merging a bunch of the PRs to actually make some progress. We really need to unify both branches and just maintain one for simplicities sake.

PR #327, opened 2015-07-14
Bugfix in issue.remove_from_labels

travisreed-wf:
If a label name is passed in with a space (such as "help wanted") then issue.remove_from_labels raises a 400 error.

This is because the url sent to github includes a space.


jzelinskie:
Awesome find! Can you write a simple test for this?

travisreed-wf:
@jzelinskie I tried updating the test on Issue.testAddAndRemoveLabelsWithStringArguments(self)

but I get a 404 from github

I have included user creds under my root directory in a file called GithubCredenitials.
This is the command I am running:
`python -m github.tests Issue.testAddAndRemoveLabelsWithStringArguments --record --auth_with_token`

I have committed my test with the hope that you can take a look. I am assuming that it is an auth problem.

Issue #310, opened 2015-04-17
Support for updated members list endpoint


bgreeley:
Add support for the filter and role parameters.

https://developer.github.com/v3/orgs/members/#members-list

PR #334, opened 2015-08-18
Add support for filter/role options in Organization.get_members()

sbesson:
This PR should address https://github.com/PyGithub/PyGithub/issues/310

@jzelinskie: what is your policy for the Organization tests under `github/test/ReplayData`. Do you want to migrate to `PyGithub` as the organization used to generate the API responses?

### DIFF
--- a/github/Organization.py
+++ b/github/Organization.py
@@ -469,20 +469,22 @@ class Organization(github.GithubObject.CompletableGithubObject):
             url_parameters
         )
 
-    def get_members(self, filter=github.GithubObject.NotSet,
+    def get_members(self, filter_=github.GithubObject.NotSet,
                     role=github.GithubObject.NotSet):
         """
         :calls: `GET /orgs/:org/members <http://developer.github.com/v3/orgs/members>`_
-        :param filter: string
+        :param filter_: string
         :param role: string
         :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.NamedUser.NamedUser`
         """
-        assert filter is github.GithubObject.NotSet or isinstance(filter, (str, unicode)), filter
-        assert role is github.GithubObject.NotSet or isinstance(role, (str, unicode)), role
+        assert (filter_ is github.GithubObject.NotSet or
+                isinstance(filter_, (str, unicode))), filter_
+        assert (role is github.GithubObject.NotSet or
+                isinstance(role, (str, unicode))), role
 
-        url_parameters = dict()
-        if filter is not github.GithubObject.NotSet:
-            url_parameters["filter"] = filter
+        url_parameters = {}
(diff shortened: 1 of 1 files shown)


jzelinskie:
This change looks good! :+1:

I really don't mind if the ReplayData uses a different org than PyGithub, but if you feel like changing them around, I wouldn't object.

sbesson:
Using PyGithub would certainly be my preferred choice overall. The main issue is that due to a lack permissions, I cannot rewrite replaydata for some tests in `record` mode using the `PyGithub` repository. On the other hand `BeaverSoftware` is no longer a public organization on GitHub and new tests cannot be added using the current infrastructure.

For now, a solution could be to migrate the public tests to using `PyGithub` and leaving the ones requiring authorization in an `OrganizationLegacy` unit test class unless you have a better solution.

jzelinskie:
There really isn't much of a benefit to doing so besides visual consistency, right? I'm fine with people contributing whatever ReplayData they can in order to build their tests. It would be a little crazy to force everything to use this org; it doesn't lend well to contribution if someone needs tests for an API call with write access.

sbesson:
This makes sense. Let me know if you need anything else for this PR then. No more commit pending on my side.

github/Organization.py
```diff
@@ -469,16 +469,27 @@ def get_issues(self, filter=github.GithubObject.NotSet, state=github.GithubObjec
...
         """
         :calls: `GET /orgs/:org/members <http://developer.github.com/v3/orgs/members>`_
+        :param filter: string
+        :param role: string
         :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.NamedUser.NamedUser`
         """
+        assert filter is github.GithubObject.NotSet or isinstance(filter, (str, unicode)), filter
+        assert role is github.GithubObject.NotSet or isinstance(role, (str, unicode)), role
+
+        url_parameters = dict()
```
jzelinskie: I think I'd prefer `{}` to `dict()`.

github/Organization.py
```diff
@@ -469,16 +469,27 @@ def get_issues(self, filter=github.GithubObject.NotSet, state=github.GithubObjec
             url_parameters
         )
 
-    def get_members(self):
+    def get_members(self, filter=github.GithubObject.NotSet,
```
jzelinskie: `filter` is a python built-in function, it sucks but it would be preferable if we could use another word >_<

sbesson:
Pushed. I opted for a trailing underscore to the `filter_` keyword to follow [PEP8 naming conventions](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles).

jzelinskie:
LGTM

Issue #335, opened 2015-08-27
Issue.remove_from_labels fails when the label passed is a string with whitespace characters


klnrajareddy:
remove_from_labels should escape the spaces and other special characters in the label name, if passed as a string.

klnrajareddy:
Please see if the submitted pull request https://github.com/PyGithub/PyGithub/pull/336 helps.

PR #336, opened 2015-08-27
Fixing https://github.com/PyGithub/PyGithub/issues/335.

klnrajareddy:
Issue.remove_from_labels fails when the label passed is a string with whitespace characters


jzelinskie:
LGTM

Issue #297, opened 2015-02-22
Maintain PyGithub - looking for volunteers


jacquev6:
Hello everyone,

I've obviously neglected PyGithub in the last few months and my current priorities are hardly compatible with doing a good job on PyGithub.

I'm looking for volunteer(s) to take over PyGithub. I would transfer ownership of the main PyGithub repository to a GitHub organization, and give admin permissions on the PyPI package as well.

Please comment on this issue if you're interested.

nhomar:
You can count on us.
We use a lot your library and I was facing some little issues which can help to move forward faster this cool and well done library.

Our organization is @vauxoo you can look for our opensource work here in github.

Regards.

jzelinskie:
I wouldn't mind helping out, but I wouldn't want to be the sole maintainer. We're dependent on this library at CoreOS (specifically for Quay.io).

thiderman:
I don't think I can actually help out because of the same reasons, but big kudos for taking this step and making sure that your project lives on! :+1:

jacquev6:
Thanks for your replies!

@nhomar @jzelinskie I've transferred the repository to @PyGithub and I've invited you to @PyGithub/members team. If you give me your PyPI user names, I'll give you permissions to upload new versions of https://pypi.python.org/pypi/PyGithub.

Let me know if you have any question or if something doesn't work as expected.

@thiderman Thanks for the kudos.

nhomar:
Hello @jacquev6 

My PyPi user is nhomar as well.

Thanks, let's put in our roadmap improve and finish all what we can... 

Just FYI. We are integrating odoo with git-flow and we planned use your lib, now let's put hand in work...

regards.

jzelinskie:
I'm jzelinskie @ pypi. Thanks again for the effort!

jacquev6:
You now both have the "Maintainer" role on PyPI; this should allow you to publish new versions of the package.

Thank *you* for taking over ! I'll try to be responsive if you have questions during this transition phase.

nhomar:
@jzelinskie @jacquev6 

Hello all.

I really appreciate if you extend a little the sources of the methodology and some minimal rules to avoid misunderstandings in the way we develop.

For example:

I see you use git-flow to manage names of repositories, but I can not see the plans of things already achieve in develop branch, which is tthe difference between one and another (master and develop).

[...]

PR #337, opened 2015-09-08
README: Remove searching maintainer note

sils1297:
There are obviously volunteers maintaining this project and this note is scaring people away.

Fixes https://github.com/PyGithub/PyGithub/issues/297


jzelinskie:
There are a few volunteers, but we're always looking for more. I agree that this is scaring people away, but maybe it can be reworded into something positive.

sils1297:
@jzelinskie maybe you want to add a "We're looking for help" paragraph then with an email adress or so to contact.

Issue #278, opened 2014-10-06
Iterating through repo.get_tags() throws NoneType TypeError for repositories with lots of tags


erichaase:
For repositories with lots of tags (git, linux, etc.), iterating through _repo.get_tags()_ throws a NoneType _TypeError_. Please see the following for the steps to recreate:

```
eric@white:~/autoport> pip freeze|ack pygithub
PyGithub==1.25.0

eric@white:~/autoport> python
Python 2.7.3 (default, Feb 27 2014, 20:00:17) 
[GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from github import Github
>>> tags = Github("...").get_repo(2325298).get_tags()
>>> for tag in tags:
...     print tag
... 
<github.Tag.Tag object at 0xb6e1848c>
<github.Tag.Tag object at 0xb6e18e8c>
...
<github.Tag.Tag object at 0xb6d14b0c>
<github.Tag.Tag object at 0xb6d14c0c>
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/github/PaginatedList.py", line 48, in __iter__
    newElements = self._grow()
  File "/usr/local/lib/python2.7/dist-packages/github/PaginatedList.py", line 60, in _grow
    newElements = self._fetchNextPage()
  File "/usr/local/lib/python2.7/dist-packages/github/PaginatedList.py", line 177, in _fetchNextPage
    for element in data
  File "/usr/local/lib/python2.7/dist-packages/github/GithubObject.py", line 74, in __init__
    self._storeAndUseAttributes(headers, attributes)
  File "/usr/local/lib/python2.7/dist-packages/github/GithubObject.py", line 86, in _storeAndUseAttributes
    self._useAttributes(attributes)
  File "/usr/local/lib/python2.7/dist-packages/github/Tag.py", line 73, in _useAttributes
    if "commit" in attributes:  # pragma no branch
TypeError: argument of type 'NoneType' is not iterable
```

PR #341, opened 2015-09-25
Don't try to parse "data" if it's None

mattpep:
I suspect this is related to the cause of
https://github.com/PyGithub/PyGithub/issues/278#issuecomment-58292338 in
that the response from the github API contains null for some elements.

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -160,7 +160,7 @@ class PaginatedList(PaginatedListBase):
         data = data if data else []
 
         self.__nextUrl = None
-        if len(data) > 0:
+        if data and len(data) > 0:
             links = self.__parseLinkHeader(headers)
             if self._reversed:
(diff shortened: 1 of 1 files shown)


mattpep:
With more time I'd like to capture the response from github to find out which elements are missing, and perhaps even work out in what situations they are absent. Two things of note:

1. The versioned release of the parent repo is quite old - there is ~1 year of commits 
That repo has some discussion (in https://github.com/PyGithub/PyGithub/issues/297) seeking another maintainer(s). It means that in the short term we're not likely to get our work merged into that repo, and so should consider using our fork.

2. This fix sits atop all of that work - most of that is catching other null de-referencing issues such as this so we'll be bitten less often by other partially-missing responses.

mattpep:
Sorry, raised against wrong repo.

Issue #345, opened 2015-10-12
Support application/vnd.github.v3.star+json


danvk:
See https://developer.github.com/v3/activity/starring/#list-stargazers

When requesting a list of stargazers, you can set a custom media type to find out when each user starred your repo. It would be great if pygithub supported this.

danvk:
Here's some code:

```python
class Stargazer(github.GithubObject.NonCompletableGithubObject):
    @property
    def starred_at(self):
        """
        :type: datetime.datetime
        """
        return self._starred_at.value

    @property
    def user(self):
        """
        :type: :class:`github.NamedUser`
        """
        return self._user.value

    def _initAttributes(self):
        self._starred_at = github.GithubObject.NotSet
        self._user = github.GithubObject.NotSet
        self._url = github.GithubObject.NotSet

    def _useAttributes(self, attributes):
        if 'starred_at' in attributes:
            self._starred_at = self._makeDatetimeAttribute(attributes['starred_at'])
        if 'user' in attributes:
            self._user = self._makeClassAttribute(github.NamedUser.NamedUser, attributes['user'])


def get_stargazers(repo):
    return github.PaginatedList.PaginatedList(
        Stargazer,
        repo._requester,
        repo.url + "/stargazers",
        None,
        # XXX: pygithub doesn't support this:
        headers={
            'Accept': 'application/vnd.github.v3.star+json'
        }
    )
```

and then you have to modify `PaginatedList` to accept a `headers` keyword arg which it passes through the call to `self.__requester.requestJsonAndCheck`.

danvk:
(I'd be happy to turn this into a pull request if there's interest and adding a `headers` parameter seems like the way to go.)

PR #346, opened 2015-10-14
Thread headers through PaginatedList

danvk:
See #345 

This would allow the use of endpoints which return different content depending on headers, e.g. `/stargazers`.

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -106,7 +106,7 @@ class PaginatedList(PaginatedListBase):
         some_other_repos = user.get_repos().get_page(3)
     """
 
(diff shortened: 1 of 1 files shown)


jzelinskie:
Should we take headers directly or abstract this as another parameter type you pass such that the user doesn't need to know that it's implemented via headers?

danvk:
Most users won't interact with `PaginatedList` directly. Adding this parameter will allow the creation of new methods like `get_stargazers(with_dates=True)` that set headers without the user needing to know or care.

jzelinskie:
Good point. LGTM

PR #352, opened 2015-11-03
Updating doc link for Github Commit Statuses

babineaum:
It seems that this new link should be the appropriate link to docs for Commit Statuses. The old one just seemed to be a place holder "todo"


jzelinskie:
LGTM

PR #354, opened 2015-11-09
Add creator to get_issues

jayfk:
The `issues` endpoint of a repository has an optional parameter `creator` that is added here.

https://developer.github.com/v3/issues/#list-issues-for-a-repository


jzelinskie:
LGTM

PR #372, opened 2016-01-07
Add branch protection endpoint

khornberg:
Branch protection is in developer preview thus `get_branch` is duplicated to explicitly call the protected branch endpoint.

I'm not sure if I've placed all of the tests in the correct file.


jzelinskie:

khornberg:
So this broke the build. I'm not exactly sure what cause it. To change protected branches, the user must have the correct permissions. The record files are for my fork. Would that have caused the fail? I also didn't include some `tests/...setup...` files. Would that have done it?

jzelinskie:
Yeah, I've emailed the owner of the repository to try and fix Travis so that I can actually see the build failures. It looks like the failures are being caused by [some strings in the test data being in your namespace rather than the jacquev6 namespace](https://travis-ci.org/PyGithub/PyGithub/jobs/100904663#L637).

PR #378, opened 2016-01-14
Handle rate limit: retry after requested delay

jrouquie:
This suggestion requires to update tests/RateLimiting.py, as the call doesn't fail anymore (that's the whole point of this PR).

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -41,6 +41,7 @@ import sys
 import Consts
 import re
 import os
+import time
 
 atLeastPython26 = sys.hexversion >= 0x02060000
 atLeastPython3 = sys.hexversion >= 0x03000000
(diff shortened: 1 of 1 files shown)


justfortherec:
Has this been picked up? It would be immensely useful to have this inside the library (instead of in my wrapper class). 

What needs to be done to get this or something similar integrated?

sfdye:
We will take a shot at `requests`, with #664 being merged.

Also see #757 for possible duplicate.

mfonville:
I think this would be very useful to implement, and especially to also add support for the `Retry-After` header, to also avoid the anti-abuse limits that are in place next to the regular rate-limit, as mentioned in https://github.com/PyGithub/PyGithub/issues/493 and is documented at https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits

allevin:
@sfdye #757 is not really a duplicate, but is similar .  This PR seems to be for handing issues related to rate limits and attempted retries.  Whereas #757 is for failed comms due to a timeout (and/or other failure types).  The resolution for both may be similar.  However, I would like to see better higher level controls over the time between tries, and the number of attempts.

jrouquie:
Thanks for the distinction with #757.
About higher level controls over the time between tries, and the number of attempts, I agree it would enhance the current pull request. I think the current feature would already benefit some users if merged before waiting for more features. I needed the feature "basic rate limit handling" two years ago, implemented it, and contributed it back in the current #378. I don't need the feature anymore. Maybe if someone needs those higher level controls, she could implement it and contribute it in a new pull request? (Sadly, now that the PR has waited for two years, the main branch has evolved and there are merge conflicts).

BTW, I'm not sure why someone would want to wait longer than the delay requested by github, but if someone needs it and wants to integrate it, fine with me.

sfdye:
@jrouquie Thanks for your original PR. I am closing it for now as it was two years old. If someone wants to take anther shot, feel free to make a new PR.

PR #381, opened 2016-01-20
Expose target_commitish parameter

yannispanousis:
In https://developer.github.com/v3/repos/releases/#create-a-release there is a parameter called `target_commitish` which makes it possible to tag objects other than the current head of the default branch of the repository.

Currently we are trying to create GitHub releases with PyGithub but it is in a broken state since all our releases so far are incorrectly tagged.

We've tested and confirmed that using this parameter fixes the problem.

Could we add this please?

Thanks in advance!

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -746,11 +746,11 @@ class Repository(github.GithubObject.CompletableGithubObject):
         )
         return github.GitRef.GitRef(self._requester, headers, data, completed=True)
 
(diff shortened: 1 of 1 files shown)


jzelinskie:
The precedent for missing parameters is to use `github.GithubObject.NotSet` instead of `None`, if you make this change and add a test, I'll merge the code.

Thanks!

sfdye:
Close in favor of #625

PR #383, opened 2016-01-29
Replace httplib with requests

nricklin:
Here's a fairly simple set of tweaks to replace httplib with requests.

Why?  Two main reasons:
- Requests is WAY better (http://docs.python-requests.org/en/latest/)
- I wanted to use pyvcr or httpretty as a quick and simple way to unit test my scripts that used PyGithub, neither of which worked out of the box.  They both work great with requests.


setup.py
```diff
@@ -69,6 +69,9 @@
             =======================
 
             See http://jacquev6.github.com/PyGithub"""),
+        install_requires=[
+          "requests==2.9.1"
```
virtuald: Should you really ==2.91? Presumably this code isn't doing anything special that would work in older versions of requests.

virtuald:
One complaint (and perhaps I just missed something) is that if you're going to use requests, why not use a session object? That will automatically use HTTP keepalive and should improve performance when making multiple calls out to github since you won't have to make new TCP connections each time.

nricklin:
That's probably a good idea.  If I get a chance soon I will:

- [ ] use requests session object
- [ ] properly deal with HTTPS vs HTTP ( I kinda ignored it in this PR )

setup.py
nricklin: I just like fixing versions so nothing breaks on me unexpectedly.  Probably fine to remove the version alltogether.

jacquev6:
There is a huge downside to replacing 'httplib' by 'requests': the tests. When I first authored PyGithub, I developed a record/replay test framework. Porting this to 'requests' seemed like a huge task and that's why I never did it.
I'm all for 'requests', but I'm even more all for automated testing so I feel like this issue should be addressed at the same time as the migration.

nricklin:
Won't it just work?  Part of this PR "mocks" the class of httplib manipulated by the tests.

jakirkham:
Wouldn't this be easy to determine from the CI whether this works? Or is there something we're missing?

Note: I know nothing about the test suite here, but would love to learn. 😄

jakirkham:
Another thought on testing would be to use [betamax]( https://github.com/sigmavirus24/betamax ) or [VCR.py]( https://github.com/kevin1024/vcrpy ), which provide similar recorded test functionality for `requests`. It also might lighten the maintenance load here after the port.

jayfk:
I don't see that happening. You'd have to gather replay data for the whole test suite. Just take a look at the data directory [here](https://github.com/PyGithub/PyGithub/tree/master/github/tests/ReplayData). This will be A LOT of work to recreate.

You'd probably also need to rewrite a lot of tests, because they are all tight to @jacquev6 repos and there's no easy way to recreate them.

arpit1997:
I agrees to @jacquev6  that there is a major drawback while porting from httplib to [reqeusts](https://github.com/kennethreitz/requests). WIth httplib major part of testing is done and porting to reqeusts library will create more need of new tests. No doubt that [reqeusts](https://github.com/kennethreitz/requests). better but a new library need new tests and it is a very big task for repo owner and users of this wrapper too.

sfdye:
Closing this in favor of #664

PR #386, opened 2016-02-08
add params to NamedUser.get_starred

webmasterraj:
This adds a way to pass parameters to `NamedUser.get_starred`. Github [accepts](https://developer.github.com/v3/activity/starring/#parameters)) parameters on that endpoint for sorting repositories by created or updated, and for ordering direction. 

This is particularly useful when a user has thousands of starred repos, and you don't want to make API calls on every PaginatedList. Instead, you can pass parameters to decide how to sort and order the returned repos.

### DIFF
--- a/github/NamedUser.py
+++ b/github/NamedUser.py
@@ -467,7 +467,7 @@ class NamedUser(github.GithubObject.CompletableGithubObject):
             url_parameters
         )
 
-    def get_starred(self):
+    def get_starred(self, params=None):
         """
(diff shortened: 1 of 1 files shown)


RSully:
Other methods in this project name the parameters specifically (e.g. `sort`, `direction`), so this PR isn't consistent with that. 

However, I would personally love an extra argument for all methods that accepts arbitrary parameters.

sfdye:
I agree with @RSully this PR is not consistent for parameters and the indentation is off thus tests are failing. I am afraid I have to close it but feel free to open a new one if you want to give it another shot!

PR #395, opened 2016-04-04
Fix for pickling exception when deserializing GithubException.

jlafleche:
See: https://bugs.python.org/issue9400

"Because CalledProcessError extends PyExc_BaseException, which defines a `__reduce__` method, that special method cause the pickle load to call the exception type's `__init__` method with packed `self.args` as arguments. So if a subclass of "Exception" needs to behave correctly in pickling, it should make `self.args` meet its `__init__` method's function signature. "

**Before:**
``` python
>>> from github import GithubException
>>> import pickle
>>> pickle.loads(pickle.dumps(GithubException('a', 'b')))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1382, in loads
    return Unpickler(file).load()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 858, in load
    dispatch[key](self)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pickle.py", line 1133, in load_reduce
    value = func(*args)
TypeError: __init__() takes exactly 3 arguments (1 given)
```

**After:**
```python
>>> from github import GithubException
>>> import pickle
>>> pickle.loads(pickle.dumps(GithubException('a', 'b')))
GithubException('a', 'b')
```

### DIFF
--- a/github/GithubException.py
+++ b/github/GithubException.py
@@ -36,6 +36,7 @@ class GithubException(Exception):
         Exception.__init__(self)
         self.__status = status
         self.__data = data
+        self.args = [status, data]
 
     @property
     def status(self):
--- a/github/tests/Exceptions.py
+++ b/github/tests/Exceptions.py
@@ -25,6 +25,7 @@
 
 import github
 import sys
+import pickle
 
 import Framework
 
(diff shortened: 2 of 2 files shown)


jzelinskie:

PR #400, opened 2016-05-17
Added support to consider No proxy settings

kumarvaradarajulu:
This change is to add support to no_proxy settings. 

Currently all requests are routed to proxy is a proxy is specified in `http_proxy` or `https_proxy`. But for Github enterprise servers such as ours we have dedicated servers on intranet hence the request should bypass proxy.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -51,6 +51,7 @@ else:  # pragma no cover (Covered by all tests with Python 2.5)
     import simplejson as json  # pragma no cover (Covered by all tests with Python 2.5)
 
 import GithubException
+from Utils import should_bypass_proxies
 
 
 class Requester:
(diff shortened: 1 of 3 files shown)


kumarvaradarajulu:
@jzelinskie Can you review and merge it.

jzelinskie:
I wonder if it would be more flexible if we allowed a proxy to be provided in the constructor of the `GitHub` object rather than attempting to satisfy all uses cases within our library. This way applications could implement whatever policies they want and simply pass that policy onto us. This would in practice end up being similar to using the [`requests` library](http://docs.python-requests.org/en/master/user/advanced/#proxies).

kumarvaradarajulu:
Much of the code that I picked was from [requests] (https://github.com/kennethreitz/requests/blob/4865f0f9bf30682239aa2672dad5c488d9c9b949/requests/sessions.py#L228). Even though `requests` accepts `proxies`, that is to override environment proxies, but the fail safe code considers en settings. In this library http_proxy is being considered & used, and honoring no_proxy settings along with it must be a binding behaviour according to me.

jzelinskie:
Gotcha! It seems like this is our best bet to properly support these environment variables. We haven't created a "utils" module yet, so I'd hesitate to do it for something as closely coupled to the Requester as this. Can you just add all the code into the requester file and then rename the test file to be for Requester and add it to `AllTests.py`.

seeker89:
Hi guys.

I'm currently hitting the same problem with the lack of support for `no_proxy`. Any reason this was never merged in ?

PR #412, opened 2016-06-28
Add previous_filename property to File

jeffreymelvin-wf:
GitHub API uses the status "renamed" for the new file when a filename is changed rather than a separate "Added" and "Removed" file. The original filename is provided for such files, but is not currently exposed via an attribute of a github.File.File.


jzelinskie:
LGTM. Thanks!

PR #413, opened 2016-06-30
fix wrong expectance on requestJsonAndCheck() returning {} if no data

proyvind:
When there's no data, requestJsonAndCheck(), or rather
__structuredFromJson() returns None rather than {}.
As both None & {} evaluates to false, 'if not data:' will give the
correct behaviour regardless of returning {} or None.

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -1797,7 +1797,7 @@ class Repository(github.GithubObject.CompletableGithubObject):
             "GET",
             self.url + "/stats/contributors"
         )
-        if data == {}:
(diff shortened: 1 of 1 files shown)


jzelinskie:
Nice find. I bet there are some other files doing the same thing. We should probably fix them all at once.

proyvind:
I tried searching through code for other places, wasn't able to come up with any, so I think this is unique for these functions only. :)

jzelinskie:
LGTM! Thanks!

PR #417, opened 2016-07-05
Document how one gets an AuthenticatedUser object

edunham:
I just spent an embarrassing number of minutes trying to figure out why calling `get_user('edunham')` was returning a `NamedUser` object and how I was supposed to get an `AuthenticatedUser` instead. 

It seems that I'm also [not the first](http://stackoverflow.com/questions/28675121/how-to-create-a-new-repository-with-pygithub/38212843) to run into the issue of "where does an AuthenticatedUser object come from?". 

The docs for AuthenticatedUser were the first place I looked for the solution, so they seem like a sensible place to leave hints for future people encountering the same frustrations.


jzelinskie:
LGTM. Thanks!

PR #419, opened 2016-07-08
fix a couple typos to be clearer

dx-pbuckley:
I found the section on PRs wasn't completely clear - the problem I'm attempting to solve is to change the grammar and make it clearer what the project wants from PRs. My approach was to modify the text to use a bulleted list.

I also found different case used for `GitHubCredentials.py` and `Githubcredentials.py` and went with the lowercase `h` since I confirmed that worked in my local testing.

I thought it would be helpful to mention that `password` is still required in `GithubCredentials.py` even when using an `--auth-with-token` token, I got an error when I left it out (`AttributeError: 'module' object has no attribute 'password'`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,8 +16,12 @@ If for any reason you are not able to do that, open your issue anyway and a main
 
 ## Pull Requests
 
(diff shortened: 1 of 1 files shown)


jzelinskie:
Awesome! Thanks for the contribution! LGTM!

PR #420, opened 2016-07-12
Update doc url

dx-pbuckley:
I noticed that the URL for pygithub was pointing to the prior jacquev6 URL, which just redirects to pygithub.github.io. Not a big issue, I know, but seemed trivial to change it everywhere. I updated it everywhere save one place, the github/tests/ReplayData/ExposeAllAttributes.testAllClasses.txt - I'm not sure how to generate that file (the tests are not working/passing for me locally even on master, I get some variations of `FAILED (failures=63, errors=328)` or `FAILED (failures=174, errors=188)` and I haven't dug into it beyond following the basic instructions in CONTRIBUTING.md which haven't fixed it my local environment. Seems maybe all tests can only run locally for users with write access to github.com/jacquev6 org? NB that the pygithub.github.io URL is long enough that it made sense to put on a separate line given the current char width of the copyright notice. I welcome any additional advice on how I can run all the tests locally and/or how to generate that ExposeAllAttributes.testAllClasses replaydata (I couldn't find a single matching test named that, assuming I need to run them all).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,8 @@
 # Copyright 2013 AKFish <akfish@gmail.com>                                     #
 # Copyright 2013 Vincent Jacques <vincent@vincent-jacques.net>                 #
 #                                                                              #
-# This file is part of PyGithub. http://jacquev6.github.com/PyGithub/          #
+# This file is part of PyGithub.                                               #
+# http://pygithub.github.io/PyGithub/v1/index.html                             #
 #                                                                              #
 # PyGithub is free software: you can redistribute it and/or modify it under    #
 # the terms of the GNU Lesser General Public License as published by the Free  #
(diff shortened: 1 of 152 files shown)


jzelinskie:
The first commit looks fine. We don't really need to replace any of the replaydata unless you've added new behavior that you want to test. TravisCI claims that this passes all of the tests, so I'm not sure why you're getting failures locally.

dx-pbuckley:
thanks @jzelinskie - I wasn't sure if the tests needed the homepage URL identical between the source and the replaydata, but if they're not dependent in that way I'll remove that commit to the replaydata *.txt files and push again.

jzelinskie:
LGTM. Thanks!

PR #421, opened 2016-07-14
changed "permission" to "privacy" in create_team

mgelbart:
`permission` has been deprecated in the GitHub API, and is being replaced with a `privacy` option; see https://developer.github.com/v3/orgs/teams/#create-team

I simply changed the word `permission` to `privacy` everywhere in `create_team()` so that users can use the newer parameter.

### DIFF
--- a/github/Organization.py
+++ b/github/Organization.py
@@ -28,6 +28,7 @@
 # ##############################################################################
 
 import datetime
+import warnings
 
 import github.GithubObject
 import github.PaginatedList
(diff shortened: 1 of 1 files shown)


jzelinskie:
Thanks for the PR!

You'll have to generate new tests against a GitHub server to make the CI pass. We have a short doc on testing found [here](https://github.com/PyGithub/PyGithub/blob/master/CONTRIBUTING.md#automated-tests).

jayfk:
Shouldn't we leave the `permission` parameter in there and raise a deprecation warning? This way it won't break peoples code and they have a indicator what's wrong.

mgelbart:
Thanks for the feedback. Following @jayfk's suggestion I added `permission` back in and added a deprecation warning using `warnings`. The tests now pass.

More info: it seems that deprecation warnings are ignored by default in more recent versions of Python, so this warning doesn't do much unfortunately. Also, I continued passing `permission` to GitHub to minimize the change. I'm not sure if that's what you intended or if GitHub even uses that parameter anymore.

jayfk:
This LGTM!

zergab:
I think the test `Organization.testCreateTeamWithAllArguments` should be updated to include the `privacy` parameter. This would also mean updating the appropriate ReplayData file.

Issue #399, opened 2016-05-03
Add support for adding/updating organization membership


sidewinder12s:
Add support for this endpoint. 

https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership

ckmason:
I'd really like to see this in PyGithub too...

alertedsnake:
So.. this seems to work fine, can we get it merged in?  It's been months...

moghya:
how to invite someone with just email id using PyGithub ?

tjboring:
Hmm, what's the status of this issue? I see the following PRs that appear to address it:

- https://github.com/PyGithub/PyGithub/pull/424
- https://github.com/PyGithub/PyGithub/pull/726

Neither of the above have unittests.

But these 2 commits referenced above add the tests:
- https://github.com/jbweston/PyGithub/commit/c1ce18a24632e8d4b0752b19ab3fec9806e54593
- https://github.com/jwplayer/PyGithub/commit/3fb0e82d71bb112f3d4632e6b5883fd98d680866

sfdye:
@tjboring You are right. Would you like to glue the pieces together so we can merge this change?

tjboring:
@sfdye Sure. Let me figure out the best way to proceed. I'll update this thread with the steps once I figure it out.

tjboring:
I'm open to other suggestions, but what I've come up with is this:

- create a new fork
- copy/paste the changes from PRs 424 and 726
- write tests and get them to pass
- create new PR

tjboring:
@sfdye I'm getting a 404 trying to create the replay data:

```
$ python -m github.tests Organization.testAddMembers --record --auth_with_token
GET /orgs/BeaverSoftware None {'Authorization': 'token XXXXXX', 'User-Agent': 'PyGithub/Python'} => 404
```

sfdye:
@tjboring As I mentioned in another thread, the `BeaverSoftware` organization does not seem to exist anymore. 

Maybe you can open a PR to merge the commits first and I will take a look?

tjboring:
I figured out a workaround. PR forthcoming.

tjboring:
@sfdye https://github.com/PyGithub/PyGithub/pull/741

sfdye:
I will make a release today.

PR #424, opened 2016-07-18
added add_to_members to organization

armedgorillas:
#399

Currently, there is no implementation of the add to (nonpublic) members. I copied the add_public_members function and lightly edited it to match the endpoint.

### DIFF
--- a/github/Organization.py
+++ b/github/Organization.py
@@ -288,6 +288,23 @@ class Organization(github.GithubObject.CompletableGithubObject):
             self.url + "/public_members/" + public_member._identity
         )
 
+    def add_to_members(self, member, role=github.GithubObject.NotSet):
+        """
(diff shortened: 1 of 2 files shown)


ckmason:
I'd like to see this make it into code base too...

jzelinskie:
Looks good. Could you add a test that covers this API usage?

armedgorillas:
Can do, but I might need some guidance. Since a test would involve adding a member to an org, do we have a testing org set up that I should add members to?

sfdye:
@jzelinskie It seems that the `BeaverSoftware` org in the test no long exists. What was the background for this? Any idea?

tjboring:
i'm looking to use PyGithub in a new project and need this functionality. I'm happy to pick this up and push it across the finish line.

sfdye:
@tjboring That would be awesome!

Issue #353, opened 2015-11-06
get_repo() doesn't raise 404


klmitch:
I have code that is expecting get_repo() to raise an UnknownObjectException if it calls get_repo() on a repository that doesn't exist.  With the 1.26.0 release, this is no longer happening; get_repo() returns an object, then subsequent (unprotected) code is getting the UnknownObjectException when it attempts to refer to an attribute on that object (in this case, full_name).

klmitch:
Is there any movement on this?  This is a very serious bug that prevents me from using the latest version of PyGithub…

acsrujan:
I'm getting 404 on get_repo("repo_not_present") as intended.
```
github_instance = Github("acsrujan", "password")
user = github_instance.get_user()
user.get_repo("repo_not_present")
---------------------------------------------------------------------------
UnknownObjectException                    Traceback (most recent call last)
<ipython-input-22-50a1bf2f4078> in <module>()
----> 1 user.get_repo("repo_not_present")
/Users/acsrujan/projects/github-in-github/test-gig/lib/python2.7/site-packages/github/AuthenticatedUser.pyc in get_repo(self, name)
    843         headers, data = self._requester.requestJsonAndCheck(
    844             "GET",
--> 845             "/repos/" + self.login + "/" + name
    846         )
    847         return github.Repository.Repository(self._requester, headers, data, completed=True)

/Users/acsrujan/projects/github-in-github/test-gig/lib/python2.7/site-packages/github/Requester.pyc in requestJsonAndCheck(self, verb, url, parameters, headers, input, cnx)
    169
    170     def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
--> 171         return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
    172
    173     def requestMultipartAndCheck(self, verb, url, parameters=None, headers=None, input=None):

/Users/acsrujan/projects/github-in-github/test-gig/lib/python2.7/site-packages/github/Requester.pyc in __check(self, status, responseHeaders, output)
    177         output = self.__structuredFromJson(output)
    178         if status >= 400:
--> 179             raise self.__createException(status, responseHeaders, output)
    180         return responseHeaders, output
    181

UnknownObjectException: 404 {u'documentation_url': u'https://developer.github.com/v3', u'message': u'Not Found'}
```

klmitch:
[...]

PR #425, opened 2016-07-19
Changed default laziness of get_repo() to False

AlyxMoon:
A potential fix for #353 

The optional argument "lazy" in the function determines whether it actually checks for the existence of the repo. Users can get around this by simply including "False" as a second argument when calling it. 

Alternatively, this fix makes the default behavior of the function to check for the repo and return 404 if it doesn't exist, as it used to be.

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -190,7 +190,7 @@ class Github(object):
         )
         return github.Organization.Organization(self.__requester, headers, data, completed=True)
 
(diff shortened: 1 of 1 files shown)


sfdye:
I am +1 for this. Laziness should be disabled by default and only enabled explicitly by request. But would you minding fixing the tests?

Link #305

sfdye:
@jasonwhite what do you think?

sfdye:
Closed in favor of #882

PR #426, opened 2016-07-21
Adding "since" param to Issue.get_comments()

babineaum:
Needed to have this for something I was working on, figured I would submit a PR.

We are developing our own in house status check to make sure that a Peer Reviewed comment (ie, ":shipit:" comment) could not be placed on a PR before the latest commit. Thus requiring the owner of the PR to make sure they were "completely" finished with the PR before getting a Peer Review.

We needed to be able to collect all the comments AFTER the latest commit. This is now available via the "since" param. 

This now adds the since param, defaulting to None. 

I followed the same way the "get_issues()" is configured inside the Repository.py 

Added a test to make sure it handles the correct messaging


jzelinskie:
Awesome! This PR looks perfect! Thanks for the contribution!

babineaum:
Thanks @jzelinskie might do some more work soon since im working on status checks and this since param seems to be a lot of new places.

Issue #358, opened 2015-11-21
CommitStatus has no context


sils1297:
I can now set the context but not get it apparently.

makuk66:
For example:

```
from github import Github

github = Github('user','password')
repo = github.get_repo('docker/docker')
pull = repo.get_pull(25076)
commit = repo.get_commit(pull.head.sha)
statuses = commit.get_statuses()

for status in statuses:
    if status.context == 'docker/dco-signed':
        print "context={}, state={}, description={}.".format(status.context, status.state, status.description)
```

prints:

```
$ python no-context.py
Traceback (most recent call last):
  File "no-context.py", line 10, in <module>
    if status.context == 'docker/dco-signed':
AttributeError: 'CommitStatus' object has no attribute 'context'
```

instead of:

```
context=docker/dco-signed, state=success, description=All commits signed.
```

PR #428, opened 2016-07-26
Patch issue 358 status context

makuk66:
This change adds a context attribute to the CommitStatus.

This addresses https://github.com/PyGithub/PyGithub/issues/358.

I've added the context as a attribute test in the CommitStatus test-case and modified the replay data accordingly. Tests pass:

```
$ python -m github.tests 
..............................................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 414 tests in 4.334s

OK
```

The example repro in the issue also works.


jzelinskie:
LGTM

Issue #415, opened 2016-07-02
github.PaginatedList.PaginatedList totalCount return None


GrapeBaBa:
My question is why invoking totalCount directly return None, but iterate will get the number.

``` python
            repo_commits = repo.get_commits()
            repo_total = 0
            for _ in repo_commits:
                repo_total = repo_total + 1
```

``` python
            repo_total = repo.get_commits().totalCount
```

GrapeBaBa:
I paste my code
```python
import json
import os

import github
from github import Github

# First create a Github instance:

g = Github("grapebaba", "heatonn1",per_page=1000)


def main():
    '''
    Use small data for this application
    :return:
    '''
    with open(os.path.join(os.path.expanduser("~"),'recruitbot_data.txt'),'w') as f:
        for user in g.search_users("type:user")[0:10000]:
            user_dict = {}
            user_dict['username'] = user.login
            user_dict['id'] = user.id
            user_dict['profile_url'] = user.html_url
            user_dict['location'] = user.location
            user_dict['followers'] = user.followers
            user_dict['private_gists'] = user.private_gists
            user_dict['public_gists'] = user.public_gists
            user_dict['name'] = user.name
            user_dict['company'] = user.company
            user_dict['blog_url'] = user.blog
            user_dict['email'] = user.email
            user_dict['id'] = user.id
            user_dict['contributions']={}
            for repo in user.get_watched():
                try:
                    repo_total = 0
                    for _ in repo.get_commits():
                        repo_total = repo_total + 1
                    if repo.get_stats_contributors() is not None:
                        for contributor in repo.get_stats_contributors():
                            if contributor is not None and contributor.author.id == user_dict['id']:
                                user_dict['contributions'][repo.name]={}
                                user_dict['contributions'][repo.name]['contributor_commits']=contributor.total
                                user_dict['contributions'][repo.name]['repo_commits']=repo_total
                                user_dict['contributions'][repo.name]['language']=repo.language
                                user_dict['contributions'][repo.name]['stars']=repo.stargazers_count
                                print user_dict
                                break
                except github.GithubException as e:
                    print e

            f.write(json.dumps(user_dict)+"\n")

if __name__ == '__main__':
    main()

```

GrapeBaBa:
I have another issue, sometimes I will get a SSL error
```shell
Traceback (most recent call last):
  File "/tmp/collector.py", line 54, in <module>
    main()
  File "/tmp/collector.py", line 36, in main
[...]

PR #433, opened 2016-08-09
Implement __len__ on the PaginatedList

pgmillon:
See #415

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -50,6 +50,9 @@ class PaginatedListBase:
             for element in newElements:
                 yield element
 
+    def __len__(self):
+        return len(self.__elements)
+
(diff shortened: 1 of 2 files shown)


RSully:
I feel like this is misleading, as if you iterate over a PaginatedList you're going to get more entries than the length of elements, since it will automatically fetch more pages.

jzelinskie:
@RSully `page_length`?

RSully:
Eh that sounds like the length of the page itself. I might propose cached_length or element_count?

pgmillon:
If I get this right, the issue is that:
> if you iterate over a PaginatedList you're going to get more entries than the length of elements
Let me make you another proposal with the next commit.

PR #436, opened 2016-08-10
change file content encoding to support unicode(like chinese), py2

GHfan:
The case is that: my file is written in Chinese which is encoded in unicode, and when i create or update a file the progress is broken because the b64encode gives me UnicodeEncodeError.

The solution is simple, in the python2, encode the file content if it is instance of unicode type using 'utf-8'.


jzelinskie:
LGTM! Thanks for the PR!

PR #437, opened 2016-08-11
Releases based on travis 

jayfk:
This PR allows travis to push new PyPi releases based on git tags.

Furthermore, I've revamped the `manage.sh` script a little:

- removed the `check` and `test` function calls (that's now done by travis)
- updated the `readme` function to automatically create a changelog based on the commits between the previous release and now
- removed the twitter script

In order to issue a new release, one of the maintainers has to run `sh manage.sh publish` locally. This bumps the version, creates a new git tag, updates the changelog and pushes the changes to origin. Travis then picks up the job and runs the tests. If the tests are passing, travis creates a new release on PyPi.

Of course, this can all be done manually. All that's really needed is an updated version in `setup.py` and a tagged commit.

Everything **should** work as expected, but since there are multiple layers involved (github, travis, PyPi) there was no way to fully test it.

\*fingers crossed\*


jzelinskie:
LGTM! Thanks! If you want to cut a release after this is merged, go for it!

jayfk:
Everything worked so far, except that the newly created user wasn't able to push to PyPi.

```
HTTPError: 401 Client Error: You must be identified to edit package information for url: https://pypi.python.org/pypi
```

I see that the user is listed as *Package Index Maintainer*, could that be the problem?

jzelinskie:
Ok, I added it as an owner. I assumed that the maintainer role could add new packages.

jayfk:
This didn't do the trick :/

I'm afraid we are affected by this bug here: https://github.com/travis-ci/dpl/issues/377

jzelinskie:
@jayfk do you need help changing the password?

babineaum:
ahh was wondering about this, going to watch this.

jayfk:
> do you need help changing the password?

I just needed to get some sleep :)

After changing the password, travis successfully deployed a new release. I had to create a new tag, because there was no way to re-run the test with the old password.

PR #449, opened 2016-09-15
GitHub integrations support

jayfk:
This PR adds support for the all new [GitHub Integrations](https://developer.github.com/early-access/integrations/authentication/).

Since the `GithubIntegration` is the main entrypoint to obtain access tokens, I've created it alongside the `Github` main class.

I wasn't able to recycle much of the functionality in `Requester` because the integration is to coupled with the `Github` main class. Instead, I've used the _wonderful_ httplib.
### Usage

``` python
from github import GithubIntegration, Github

# obtained during integration creation
private_key = "some private key"
integration_id = 123

# obtained when a user installs the integration (webhook)
installation_id = 456

# get a token
integration = GithubIntegration(integration_id, private_key)
auth = integration.get_access_token(installation_id)

gh = GitHub(auth.token)
repo = gh.get_repo("user/whatever")
```
### Tests

None, yet. Integrations are in early access right now. I think it's safe to add tests once the API is stable.
### Added dependencies

In order to create the JSON web token, this PR adds [python-jose](https://github.com/mpdavis/python-jose). Another candidate is [pyjwt](https://github.com/jpadilla/pyjwt), but that isn't installable on Google Compute Engine.

### DIFF
--- a/github/InstallationAuthorization.py
+++ b/github/InstallationAuthorization.py
@@ -2,9 +2,7 @@
 
 # ########################## Copyrights and license ############################
 #                                                                              #
-# Copyright 2012 Vincent Jacques <vincent@vincent-jacques.net>                 #
-# Copyright 2012 Zearin <zearin@gonk.net>                                      #
-# Copyright 2013 Vincent Jacques <vincent@vincent-jacques.net>                 #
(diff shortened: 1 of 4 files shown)


jayfk:
This works, let's merge it in.

PR #403, opened 2016-05-27
Follow redirects in the case of a 301 status code

mat128:
PyGithub does not follow HTTP redirects. This issue is visible when a repository is moved on GitHub. GitHub returns a proper 301 with a new location for the resource, but httplib does not follow redirects on it's own.

This PR ensures it follow redirects when a 301 is returned and a location header is provided in the response.

I was unable to add a test case for it as it does not seem possible to have multiple exchanges while using ReplayingConnection.

Here is an example curl (test_redirect was renamed to test_redirect_new_name)
```
~$ curl -v https://api.github.com/repos/mat128/test_redirect
* Adding handle: conn: 0x7fb7a9804000
* Adding handle: send: 0
* Adding handle: recv: 0
* Curl_addHandleToPipeline: length: 1
* - Conn 0 (0x7fb7a9804000) send_pipe: 1, recv_pipe: 0
* About to connect() to api.github.com port 443 (#0)
*   Trying 192.30.252.126...
* Connected to api.github.com (192.30.252.126) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
* Server certificate: *.github.com
* Server certificate: DigiCert SHA2 High Assurance Server CA
* Server certificate: DigiCert High Assurance EV Root CA
> GET /repos/mat128/test_redirect HTTP/1.1
> User-Agent: curl/7.30.0
> Host: api.github.com
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
* Server GitHub.com is not blacklisted
< Server: GitHub.com
< Date: Fri, 27 May 2016 13:28:20 GMT
< Content-Type: application/json; charset=utf-8
< Content-Length: 168
< Status: 301 Moved Permanently
< X-RateLimit-Limit: 60
< X-RateLimit-Remaining: 55
< X-RateLimit-Reset: 1464359212
< Location: https://api.github.com/repositories/59836255
< X-GitHub-Media-Type: github.v3
< Access-Control-Expose-Headers: ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval
< Access-Control-Allow-Origin: *
< Content-Security-Policy: default-src 'none'
< Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
< X-Content-Type-Options: nosniff
[...]

PR #451, opened 2016-09-16
Fix status check

f1nality:
Fixes redirect status check as stated here:
https://github.com/PyGithub/PyGithub/pull/403#issuecomment-244069423


jzelinskie:
LGTM! Thanks!

PR #460, opened 2016-09-23
Replace deprecated "permission" with "privacy" in create_team.

zergab:
The "Create team" endpoint in the teams API allows to specify `privacy ` as a parameter while the `permissions` parameter is deprecated. See https://developer.github.com/v3/orgs/teams/#create-team


zergab:
Realized there was another PR for this already.

Issue #446, opened 2016-09-12
Follow PEP8


mingrammer:
Some codes (variables, methods) don't follow PEP8.

Why the codes don't follow PEP8?

arpit1997:
I ran some inspections and found some PEP8 code style violation like: 

- Function name should be lowercase 

- variable in function name should be lowercase

- CamelCase variable imported as constant

while we ignore line length limit for all files

mingrammer:
Yes those are

Are you contributor of here? @arpit1997

arpit1997:
No I am not a contributor of here but if this issue is correct and need to be solved then i can try to solve it  @mingrammer

mingrammer:
Oh, welcome! @arpit1997. Are you willing to resolve the issues with me?

arpit1997:
i will first try to resolve with some of my tools including pycharm, PEP8 command line etc and we can discuss it later on for further corrections

mingrammer:
79 lines limits applied?

arpit1997:
Sorry i did not understand clearly what you are saying. Can you please explain?

mingrammer:
Ah, sorry.
Um, your PR seems like to resolve the problems related to line limits of PEP8 (79 lines). But, I can not see the fixed codes for camelCase related problems.

arpit1997:
I saw that mistake just now. Thanks for for your advice, i will try to resolve this by tomorrow or in 2-3 days

mingrammer:
Ye, thanks for your efforts.

PR #461, opened 2016-09-24
Now following PEP8 code style in whole code base

arpit1997:
fixed issue #446


akumria:
This PR can be closed since it both conflicts and does not take into account that E501 (line-length) is being ignored.

arpit1997:
This was a very late reply :smiley:

PR #463, opened 2016-09-25
Add info

f1nality:
Response headers have useful info (https://developer.github.com/v3/#rate-limiting)


sfdye:
Rate limit info is already available:
http://pygithub.readthedocs.io/en/latest/github.html#github.MainClass.Github.get_rate_limit

Issue #405, opened 2016-05-30
Add support for read only deploy keys


birkholz:
Deploy keys have a field "read_only", which cannot currently be set or retrieved.
https://developer.github.com/v3/repos/keys/

sfdye:
+1

jasonwhite:
Fixed in #570.

PR #467, opened 2016-10-06
Add support for read_only keys.

mattclegg:
This fixes issue https://github.com/PyGithub/PyGithub/issues/405

The new setting is to 'False' by default to preserve existing functionality.

Updated tests to show usage.


jasonwhite:
Merged #570 in favor of this one.

Issue #380, opened 2016-01-20
Expose target_commitish parameter when creating GitHub releases


yannispanousis:
In https://developer.github.com/v3/repos/releases/#create-a-release there is a parameter called `target_commitish` which makes it possible to tag objects other than the current head of the default branch of the repository.

Currently we are trying to create GitHub releases with PyGithub but it is in a broken state since all our releases so far are incorrectly tagged.

We've tested and confirmed that using this parameter fixes the problem. The patch that works is very simple, see below.

Could we add this please?

Thanks in advance!

```
From 10ffec64efcd1b5983c5d9c1a4df91ef86c6d774 Mon Sep 17 00:00:00 2001
From: Yannis <yannis@lystable.com>
Date: Wed, 20 Jan 2016 13:51:23 +0000
Subject: [PATCH] Expose target_commitish parameter in create_git_release

See https://developer.github.com/v3/repos/releases/#create-a-release
---
 github/Repository.py | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/github/Repository.py b/github/Repository.py
index 3803548..d483905 100644
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -746,11 +746,11 @@ class Repository(github.GithubObject.CompletableGithubObject):
         )
         return github.GitRef.GitRef(self._requester, headers, data, completed=True)
 
-    def create_git_tag_and_release(self, tag, tag_message, release_name, release_message, object, type, tagger=github.GithubObject.NotSet, draft=False, prerelease=False):
+    def create_git_tag_and_release(self, tag, tag_message, release_name, release_message, object, type, tagger=github.GithubObject.NotSet, target_commitish=None, draft=False, prerelease=False):
         self.create_git_tag(tag, tag_message, object, type, tagger)
         return self.create_git_release(tag, release_name, release_message, draft, prerelease)
 
-    def create_git_release(self, tag, name, message, draft=False, prerelease=False):
+    def create_git_release(self, tag, name, message, target_commitish=None, draft=False, prerelease=False):
         assert isinstance(tag, (str, unicode)), tag
         assert isinstance(name, (str, unicode)), name
         assert isinstance(message, (str, unicode)), message
@@ -763,6 +763,8 @@ class Repository(github.GithubObject.CompletableGithubObject):
             "draft": draft,
             "prerelease": prerelease,
         }
+        if target_commitish:
+            post_parameters["target_commitish"] = target_commitish
[...]

PR #473, opened 2016-10-14
adding target_commitish parameter to methods for creating a release

ben-whitney:
Adding `target_commitish` as an optional parameter to methods for creating a release and adding tests for those methods.  Resolves issue https://github.com/PyGithub/PyGithub/issues/380 whose previous pull request was closed.

This addresses the issues found in an older pull request (https://github.com/PyGithub/PyGithub/pull/381) that was closed.  It wasn't merged due to a lack of tests and the use of `None` as opposed to `github.GithubObject.NotSet`.  Both of those issues have been resolved.  Additionally, one bug was resolved where `target_commitish` was not being passed to the `create_git_release` method from the `create_git_tag_and_release` method.


ben-whitney:
The failures seem to imply that the tests have to work with login/password for authentication but I use 2-factor authentication which required me to use a token. Please advise on what I can do to fix the tests. Thanks.

sfdye:
@ben-whitney Thanks for your PR, but I believe this feature is already merged in by from PR #625.

PR #476, opened 2016-10-17
Removed unused imports

jwilk:
Since 1d9ad14fa918, the setup script no longer uses `subprocess`, `shutil` and `os` modules.


jzelinskie:
Nice find. Thanks!

PR #479, opened 2016-10-21
add html_url to GitRelease

ben-whitney:
Adding html_url to GitRelease as its available in the json return but currently not available.  Updating an existing test to account for this addition as well.


jzelinskie:
LGTM! Thanks!

PR #491, opened 2016-11-25
Support team membership roles

mgorny:
Support filtering members by role in `get_members()` and adding membership with a role in `add_membership()`.


sfdye:
Merged in favor of #638 where `add_membership` is already implemented.

sfdye:
Thank you very much for your contribution @mgorny

Issue #494, opened 2016-11-29
GithubObject get__repr__ chokes on non-ascii values


sjcorbett:
Example using Python 2.7:
```
from github import Github
g = Github(user, token)
r = g.get_repo("apache/brooklyn-server")
r.get_pull(465)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/sam/.virtualenvs/brooklyn-changelog/lib/python2.7/site-packages/github/PullRequest.py", line 48, in __repr__
    return self.get__repr__({"number": self._number.value, "title": self._title.value})
  File "/Users/sam/.virtualenvs/brooklyn-changelog/lib/python2.7/site-packages/github/GithubObject.py", line 226, in get__repr__
    params=", ".join(list(format_params(params)))
  File "/Users/sam/.virtualenvs/brooklyn-changelog/lib/python2.7/site-packages/github/GithubObject.py", line 223, in format_params
    yield '{k}="{v}"'.format(k=k, v=v) if isinstance(v, (str, unicode)) else '{k}={v}'.format(k=k, v=v)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2026' in position 69: ordinal not in range(128)
```

The character in question is '…'

PR #495, opened 2016-11-29
Encode GithubObject repr values in utf-8

sjcorbett:
Fixes https://github.com/PyGithub/PyGithub/issues/494.


jayfk:
This won't work on Python 3.

sjcorbett:
@jayfk second time lucky.

jayfk:
This works on Python 2 and 3 👍

PR #393, opened 2016-03-27
add_to_collaborators: add permission param

Moisi:
Signed-off-by: Ohad Moisi ohad.moisi@tandemg.com

PR #497, opened 2016-11-30
Add permissions parameter to add_to_collaborators

dave-tucker:
See the Github API Reference [here](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator)

This change will only apply if the optional `permissions` parameter is
provided to `add_to_collaborators`. If not, it will behave as before

This is similar to the implementation in #393 but differs as explained above.
Happy to close if the other implementation is preferred, but either way I would very much like to see this feature merged!


dave-tucker:
Closing due to lack of activity

PR #500, opened 2016-12-05
Add details of repo type for get_repos documentation

MichaelPereira:
Taken from https://developer.github.com/v3/repos/#list-organization-repositories, it would save the trip to the github API for whoever is reading the documentation.


MichaelPereira:
thanks @nhomar !

PR #506, opened 2016-12-19
Fix README instructions

jamesbroadhead:
repo.edit takes a mandatory first argument "name". Thus, it currently results in:
```
$ python github_mechanize_forkers.py 
Axelrod
  File "github_mechanize_forkers.py", line 11, in <module>
    repo.edit(has_wiki=False)
TypeError: edit() takes at least 2 arguments (2 given)
```

Fix the instructions by removing the "edit" call entirely -- basic instructions like this should prefer read-only operations

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,6 @@ g = Github("user", "password")
 # Then play with your Github objects:
 for repo in g.get_user().get_repos():
     print repo.name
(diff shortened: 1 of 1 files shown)


jzelinskie:
LGTM. Thank you!

PR #522, opened 2017-02-06
added tarball_url, zipball_url, prerelease and draft property

infinity0n3:
Added few more properties supported by v3 API for releases


infinity0n3:
Sure. No problem to extend GitRelease test case with new attributs.
Only problem is that in the tests PyGithub repo is used and github returns an empty releases response even though it does have 56 releases.

Reproduce the empty result by visiting
[PyGithub releases using github api](https://api.github.com/repos/PyGithub/PyGithub/releases)

jdno:
@nhomar I added some tests for the attributes. Is that what you had in mind?

jdno:
Can we get this merged?

sfdye:
Thanks for your PR @infinity0n3 

Merged @jdno

PR #530, opened 2017-02-22
Update the RepositoryKey class

lra:
# The problems I attempt to solve

- `RepositoryKey.created_at` is available on the Github API, but not in
PyGithub
- `RepositoryKey.edit()` is not supported anymore by the Github API,
see https://developer.github.com/v3/repos/keys/#edit-a-deploy-key

# How I went about solving the problem

I added myself a new deploy key to one of my repositories, and recorded
the existing tests using my account.
I could not keep the existing test data as the specified key does not
exist anymore, I added a comment for the next contributor to help him
figure out how to record new tests.


sfdye:
I cloned this branch locally and tests are passing 👍 I wonder why travis didn't get triggered.

Issue #183, opened 2013-07-30
Question: Is PyGitHub Thread Safe?


ptwobrussell:
Is PyGitHub thread safe? For example, if I were to use a thread pool to repeatedly make calls a function like get_followers() which is calling http://developer.github.com/v3/users/followers/ under the hood, is there any danger that you know of within 1) PyGitHub itself and 2) with the way that GitHub's API would respond?

dzieciou:
An alternative would be two create multiple instances of `GitHub` class. I guess instances would be isolated enough from each other.

PR #532, opened 2017-02-24
[docs] warn about using threads

svenevs:
Please make explicit that this library is not thread-safe.

The documentation for this project is excellent, and for the most part I was able to get everything up and running *very* quickly!  However, an unfortunate amount of time was just spent debugging my own program only to realize what the actual problem is.

It makes sense why this library is not thread-safe, and I'm in no way asking for this feature!  But I seem to be [the second person](https://github.com/PyGithub/PyGithub/issues/183) to have done this.  As that user indicates, us expecting PyGithub to be threadsafe was a foolish thing to do.

At the same time, given the nature of the API and what it does, it feels natural to assume that it would be thread-safe.

A simple warning will probably save a lot of other people a lot of time :wink:

Thanks for PyGithub!

### DIFF
--- a/doc/introduction.rst
+++ b/doc/introduction.rst
@@ -8,6 +8,10 @@ Should you have any question, any remark, or if you find a bug,
 or if there is something you can do with the API but not with PyGithub,
 please `open an issue <https://github.com/PyGithub/PyGithub/issues>`__.
 
+.. warning::
+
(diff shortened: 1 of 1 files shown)


svenevs:
Oh dear.  **Very** sorry, I started using this project and ended up on a different (but similarly named) project.  I have not tested yours with threads, so this may not be relevant.

PR #557, opened 2017-04-19
add "integration" install option with minimal docs

### LINKED ISSUE #554: GitHub Integration raises "NotImplementedError Algorithm not supported"
We have working github integration code using PyGithub v1.32 that does essentially:

```python
integration = github.GithubIntegration(settings.GITHUB_INTEGRATION_ID, settings.GITHUB_INTEGRATION_PRIVATE_PEM)
inst_token = integration.get_access_token(installation_id).token
```
After upgrading to v1.34 this code raises "NotImplementedError Algorithm not supported"

I suspect it has to do with the [switch to pyjwt from python-jose](https://github.com/PyGithub/PyGithub/commit/d447eb13b9f4688a4c981ca03b1b3111fb299142)

paulortman:
Would close #554 

Tested PyGithub installation locally in a blank virtualenvs (both 2.7 and 3.6) by `pip install -e .['integrations']`  and it successfully builds and installs the cryptography package as expected.

### DIFF
--- a/doc/introduction.rst
+++ b/doc/introduction.rst
@@ -31,6 +31,9 @@ This package is in the `Python Package Index
 be enough.  You can also clone it on `Github
 <http://github.com/PyGithub/PyGithub>`__.
 
+If you wish to use GitHub Integrations, you'll want to be sure to install the
+'integrations' option: ``pip install PyGithub['integrations']``
+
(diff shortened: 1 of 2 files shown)


jayfk:
Thanks 👍

PR #562, opened 2017-04-26
Add NamedUser permissions attribute for Repository.get_collaborators()

a13m:
permissions is valid as per https://developer.github.com/v3/repos/collaborators/#response


sfdye:
@a13m Thanks for your PR, but this attribute is already in the master.

Issue #559, opened 2017-04-21
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host


moghya:
Traceback (most recent call last):
  File "C:/Users/moghya_s/Desktop/toGithub/test.py", line 11, in <module>
    test_upToGithub()
  File "C:/Users/moghya_s/Desktop/toGithub/test.py", line 8, in test_upToGithub
    user.cforceUpload('codeforcesUsername','codeforcesPassword')
  File "C:\Users\moghya_s\Desktop\toGithub\upload\uploader.py", line 45, in cforceUpload
    self.repo.create_file('/cforce/'+abc['contestId']+'/'+abc['problemName']+'/'+abc['id']+'.'+abc['ext'],'cforce solution for '+abc['problemName']+' with submissionId '+abc['id']+' is added.',abc['code'])
  File "C:\Program Files\Anaconda3\lib\site-packages\github\Repository.py", line 1310, in create_file
    input=put_parameters
  File "C:\Program Files\Anaconda3\lib\site-packages\github\Requester.py", line 172, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
  File "C:\Program Files\Anaconda3\lib\site-packages\github\Requester.py", line 213, in requestJson
    return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
  File "C:\Program Files\Anaconda3\lib\site-packages\github\Requester.py", line 252, in __requestEncode
    status, responseHeaders, output = self.__requestRaw(cnx, verb, url, requestHeaders, encoded_input)
  File "C:\Program Files\Anaconda3\lib\site-packages\github\Requester.py", line 283, in __requestRaw
    output = response.read()
  File "C:\Program Files\Anaconda3\lib\http\client.py", line 458, in read
    s = self.fp.read()
  File "C:\Program Files\Anaconda3\lib\socket.py", line 575, in readinto
    return self._sock.recv_into(b)
  File "C:\Program Files\Anaconda3\lib\ssl.py", line 929, in recv_into
    return self.read(nbytes, buffer)
  File "C:\Program Files\Anaconda3\lib\ssl.py", line 791, in read
    return self._sslobj.read(len, buffer)
  File "C:\Program Files\Anaconda3\lib\ssl.py", line 575, in read
    v = self._sslobj.read(len, buffer)
[...]

PR #567, opened 2017-04-29
resolved spaces in directory name in path

moghya:
as GitHub doesn't allow spaces in directory name GitHub API was raising an exception and hence those spaces are replaced by '-' character.
Regular expression is used and lambda function to replace in all directory names is implemented

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -30,6 +30,7 @@
 # along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
 #                                                                              #
 # ##############################################################################
+import re   #imported for replacing the spaces in path of create_file method
 import sys
 import urllib
 import datetime
@@ -1303,7 +1304,9 @@ class Repository(github.GithubObject.CompletableGithubObject):
             put_parameters["author"] = author._identity
         if committer is not github.GithubObject.NotSet:
             put_parameters["committer"] = committer._identity
-
+        
+        #for replacing one or more spaces in directory names specified in  path with  '-' 
+        path =  '/'.join([re.sub(' +','-',directory) for directory in path.split('/')[0:-1]]+[path.split('/')[-1]])
         headers, data = self._requester.requestJsonAndCheck(
             "PUT",
             self.url + "/contents" + path,


Enteleform:
@moghya 

I just ran into the same issue you mentioned in #559, and have found that the issue is caused by other characters as well.

Specifically:
```python
[
  (' ', '%20'),
  ('`', '%60'),
  ('@', '%40'),
  ('#', '%23'),
  ('$', '%24'),
  ('%', '%25'),
  ('^', '%5E'),
  ('&', '%26'),
  ('=', '%3D'),
  ('+', '%2B'),
  ('[', '%5B'),
  (']', '%5D'),
  ('{', '%7B'),
  ('}', '%7D'),
  (';', '%3B'),
  (',', '%2C'),
]
```

&nbsp;  
`urllib.parse` seems like it should handle all of the above cases, which would the resolve the encoding issues:
```python
import urllib

def encode_path(path:str):
  return urllib.parse.quote(path)

def decode_path(path:str):
  return urllib.parse.unquote(path)
```

github/Repository.py
```diff
@@ -1303,7 +1304,9 @@ def create_file(self, path, message, content,
             put_parameters["author"] = author._identity
         if committer is not github.GithubObject.NotSet:
             put_parameters["committer"] = committer._identity
-
+        
+        #for replacing one or more spaces in directory names specified in  path with  '-' 
+        path =  '/'.join([re.sub(' +','-',directory) for directory in path.split('/')[0:-1]]+[path.split('/')[-1]])
```
htgoebel: - This is hard to understand. What exactly does this code do?
- It looks like you are replacing spaces in all path parts but the last one. But why not in the last part?
- Why are you replace *multiple* spaces with a *single* dash?
- Also you are splitting the whole string twice.

htgoebel:
ACk, this looks like the URL should be encoded properly using urlencode, if `requests.requestJsonAndCheck()` does not do this.

github/Repository.py
moghya: Replacing spaces with dash as github does .
Github doesn't allow spaces in directory names. It automatically converts spaces into dashes.

Not converting spaces in last part as last part would be filename and spaces in filename are allowed.


Yea, splitting string twice should not he done.
My bad there.

sfdye:
What's the behavior of uploading a file to github via web if the filename contains space? Without some unit tests, it's very hard for me to merge this.

PR #593, opened 2017-07-10
fixing date format for milestone creation

behrisch:
This change fixes the faulty time format in milestone creation.

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -972,7 +972,7 @@ class Repository(github.GithubObject.CompletableGithubObject):
         if description is not github.GithubObject.NotSet:
             post_parameters["description"] = description
         if due_on is not github.GithubObject.NotSet:
-            post_parameters["due_on"] = due_on.strftime("%Y-%m-%d")
+            post_parameters["due_on"] = due_on.strftime("%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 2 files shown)


behrisch:
Currently milestone creation fails when giving a due date because github expects the date in ISO 8601 format, see
https://developer.github.com/v3/issues/milestones/#create-a-milestone
This patch fixes it.

btorresgil:
Need this fix for a project

jasonwhite:
Sorry this sat for so long. This project is in desperate need of more maintainers.

btorresgil:
Thanks for merging, though with some testing it looks like this isn't a complete fix.  The problem is it always passes UTC to the API.  So for example, if you're in the United States and pass a `date` object like `date(2018, 2, 4)`, then GitHub will interpret that as 2018-02-04T00:00:00Z which is actually 2018-02-03 in the United States.  So the due date in the milestone will be the day before the one you passed in.  This is incorrect behavior.

Since GitHub API needs expects a date/time in ISO 8601 format, it would be best to pass in a `datetime` object and call `.isoformat()` on the object, than to bother with `date` objects.  This will also be more compatible with enhanced datetime libraries like [pendulum](https://pendulum.eustace.io/).

I created a Pull Request to correct this:  https://github.com/PyGithub/PyGithub/pull/698

Issue #487, opened 2016-11-18
PaginatedList totalCount raises AttributeError


jakirkham:
Ran into this rather odd issue locally. Not quite sure what is going on, but figured it deserved an issue as I saw it repeatedly.

```python
>>> l
<github.PaginatedList.PaginatedList instance at 0x...>
>>> l[0]
>>> IssueComment(user=NamedUser(login="..."), id=...)
>>> l.totalCount
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-34-ba47f831ae47> in <module>()
----> 1 l.totalCount

/zopt/conda2/lib/python2.7/site-packages/github/PaginatedList.pyc in totalCount(self)
    125     def totalCount(self):
    126         if not self.__totalCount:
--> 127             self._grow()
    128 
    129         return self.__totalCount

/zopt/conda2/lib/python2.7/site-packages/github/PaginatedList.pyc in _grow(self)
     59 
     60     def _grow(self):
---> 61         newElements = self._fetchNextPage()
     62         self.__elements += newElements
     63         return newElements

/zopt/conda2/lib/python2.7/site-packages/github/PaginatedList.pyc in _fetchNextPage(self)
    160             self.__nextUrl,
    161             parameters=self.__nextParams,
--> 162             headers=self.__headers
    163         )
    164         data = data if data else []

/zopt/conda2/lib/python2.7/site-packages/github/Requester.pyc in requestJsonAndCheck(self, verb, url, parameters, headers, input, cnx)
    170 
    171     def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
--> 172         return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
    173 
    174     def requestMultipartAndCheck(self, verb, url, parameters=None, headers=None, input=None):

/zopt/conda2/lib/python2.7/site-packages/github/Requester.pyc in requestJson(self, verb, url, parameters, headers, input, cnx)
    211             return "application/json", json.dumps(input)
    212 
--> 213         return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
    214 
    215     def requestMultipart(self, verb, url, parameters=None, headers=None, input=None):

/zopt/conda2/lib/python2.7/site-packages/github/Requester.pyc in __requestEncode(self, cnx, verb, url, parameters, requestHeaders, input, encode)
[...]

PR #596, opened 2017-07-12
fix: totalCount always returns None

Tommos0:
#487
#415

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -60,6 +60,8 @@ class PaginatedListBase:
     def _grow(self):
         newElements = self._fetchNextPage()
         self.__elements += newElements
+        if not self._couldGrow():
+            self._totalCount = len(self.__elements)
         return newElements
 
     class _Slice:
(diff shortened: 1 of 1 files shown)


Tommos0:
https://github.com/PyGithub/PyGithub/issues/487
https://github.com/PyGithub/PyGithub/issues/415

Tommos0:
Committed solution for totalCount will request more pages until it is exhausted to get the total count. There is a better way to get it in a single request (GitHub API returns a link to the last page in the header) - we can use this by setting `per_page = 1`. I'm using this as a workaround at the moment:

```python
def total_count(paginated_list):
    pars = paginated_list._PaginatedList__nextParams.copy()
    pars.update({"per_page" : 1, "page" : 1}) 
    headers, data = paginated_list._PaginatedList__requester.requestJsonAndCheck(
        "GET",
        paginated_list._PaginatedList__firstUrl,
        parameters=pars,
        headers=paginated_list._PaginatedList__headers
    )
    return int(re.match(r'.*[&\?]page=(\d*).*?rel="last"', headers['link']).group(1))
```

k1-hedayati:
I'm having issue with empty `totalCount`, please accept this merge

koglinjg:
To add to @Tommos0's workaround. I changed `return 1` to `return len(data)` to account for cases where the total count is `0`

sfdye:
```
if not 'link' in headers:
	return 1
```

Should it `return 0` instead?

It seems that when the result is empty, the `Link` header will disappear.

sfdye:
@Tommos0 Maybe a new PR for the more efficient total count?

koglinjg:
@sfdye: I noticed that the `Link` header was not present if 0 or 1 results were returned. The data list will either be empty or have one result. So I think when the `link` header is not there just doing `len(data)` is the right thing to do.

sfdye:
@koglinjg you are right.

sfdye:
I'd prefer the `Link` header approach with setting `per_page` to 1. If someone can make a new PR for it I am more than happy to review.

PR #602, opened 2017-07-27
added role parameter to add_membership function

suranc:
The add_membership function did not accept the optional role parameter, to add a member as a team maintainer.  I added a keyword parameter called "role" that will optionally accept a string value for the desired role.


sfdye:
Closed in favor of #638

PR #623, opened 2017-09-05
Added missing parameters for repo creation

MrBlaise:
There were some missing parameters from the repo_create method

The missing ones are the following:

- has_projects
- license_template
- allow_squash_merge
- allow_merge_commit
- allow_rebase_merge

Also the team_id parameter should be an id not a Team object.


santiagocardenas:
+1 on this. I would appreciate a prompt merge and release!

Issue #572, opened 2017-05-05
Is there a more direct way to get PullRequest from Issue?


davidxia:
Thanks for the awesome repo and work. It's made my life a lot easier.

I'm writing a script to comment on all open PRs created by me in order to remind people to review them. It turns out the most direct way to list all my open PRs is to use the issue search API endpoint. [See discussion here](http://stackoverflow.com/a/43807803/553994).

This works great and is better than my original [roundabout way](http://stackoverflow.com/questions/17412809/how-to-get-my-pull-requests-from-github-api/43807803#comment74624202_17438627).

Unfortunately, `Github.search_issues()` returns `Issue` and there's no easy way to get the associated `PullRequest` object. There's `Issue.pull_request`, but this returns an [`IssuePullRequest`](http://pygithub.readthedocs.io/en/latest/github_objects/IssuePullRequest.html#github.IssuePullRequest.IssuePullRequest) object which isn't helpful.

My current workaround is `Issue` -> `Repository` -> `PullRequest`.

```python
issue.repository.get_pull(issue.number)
```

Is there a better way to get the corresponding `PullRequest` from an `Issue`?

mikeage:
I use pygithub for parsing webhooks, and as most of the notifications about a pull request come as an `issue_comment` notification, I have the same challenge you do; I haven't found anything better than `issue.repository.get_pull(issue.number)`. 

Not sure if that's the answer you're looking for, but at least I can confirm that you're not alone

tmshn:
I added the methods for this!
:point_right: https://github.com/PyGithub/PyGithub/pull/630

sfdye:
Yes, there is now. You can use the `type` parameter to specify only search against PR:
https://developer.github.com/v3/search/#parameters-3

Something like:

`g.search_issues(query="query", qualifiers={'type':'pr'})`

sfdye:
Feel free to re-open if needed.

tmshn:
Thanks @sfdye, but I'm afraid you are seeing wrong point (though I can't re-open this cz I'm not the author).

Even if you specify `{'type':'pr'}` on `search_issues`--well that's true it returns only pull requests, but it's represented in the `github.Issue.Issue` class. This is that kind of interface.

```python
from github import Github

g = Github()

for pr in g.search_issues(query='License', **{'repo':'PyGithub/PyGithub', 'type':'pr'}):
[...]

PR #630, opened 2017-09-18
Added Issue.as_pull_request() and PullReqest.as_issue()

tmshn:
As reported in #572, there's no straightforward way to convert from `github.Issue.Issue` to `github.PullRequest.PullRequest` (when it is actually a pull request), and vice versa.

A workaround is introduced there, but that is ugly and may be confusing for readers.

```python
>>> issue.repository.get_pull(issue.number)
```

To fix this, I added `Issue.as_pull_request()` and `PullReqest.as_issue()` methods to do this.
Internally it calls same api as the workaround does, but way clearer.

```python
>>> issue.as_pull_request()
```


github/Issue.py
```diff
@@ -230,6 +230,17 @@ def user(self):
...
         return self._user.value
 
+    def as_pull_request(self):
+        """
+        :calls: `GET /repos/:owner/:repo/pulls/:number <http://developer.github.com/v3/pulls>`_
+        :rtype: :class:`github.PullRequest.PullRequest`
+        """
+        headers, data = self._requester.requestJsonAndCheck(
+            "GET",
+            "/pulls/".join(self.url.rsplit("/issues/", 1))
```
tmshn: This is the hack to replace only last occurrence of the pattern; just `self.url.replace("/issues/", "/pulls/")` will replace all, which might be user name or repo name.
https://stackoverflow.com/questions/2556108/rreplace-how-to-replace-the-last-occurrence-of-an-expression-in-a-string

sfdye:
@tmshn Thanks very much for your PR. However now as #572 is close, I don't feel the necessity for this PR anymore. I am afraid I will close it for now.

github/Issue.py
sfdye: Could there be more than one occurrence? If not, why not just use `str.replace()`

tmshn: I don't know much about the Github specification, but I cannot guarantee that following situation never happen:

- User/organization name or repository name is `"issues"` or `"pulls"`
- Hostname of Github Enterprise is configured to something like `"mycompany.com/issues/ghe"`

Of course it's ridiculous idea to do that, but if it's POSSIBLE, we should care that.

If we can say this never happens, I will use `str.replace()`.

sfdye: You are absolutely right. Thanks for being so prudent.

Issue #627, opened 2017-09-12
pyjwt causing issues with GithubIntegration


edouardpoitras:
I encountered this issue when using the GithubIntegration class:

    import github
    app_id = aaaa
    private_key = open('key.pem').read()
    installation_id = bbbbb
    integration = github.GithubIntegration(app_id, private_key)
    token = integration.get_access_token(installation_id)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/eddie/.local/lib/python3.6/site-packages/github/MainClass.py", line 692, in get_access_token
        data=response_text
        github.GithubException.GithubException: 401 {"message":"Invalid credentials","documentation_url":"https://developer.github.com/v3"}

In the GithubIntegration.create_jwt() function line 636:
Replaced:

    return jwt.encode(
                payload,
                key=self.private_key,
                algorithm="RS256"
            )
With:

    from jose import jwt as JWT                                                                 
    return JWT.encode(payload, self.private_key, 'RS256')

Trying again...

    import github
    app_id = aaaa
    private_key = open('Downloads/test.pem').read()
    installation_id = bbbbb
    integration = github.GithubIntegration(app_id, private_key)
    token = integration.get_access_token(installation_id)
    token
    InstallationAuthorization(expires_at=2017-09-12 17:11:32)

Not sure what the difference is between the pyjwt and python-jose implementations are, but pyjwt doesn't seem to work for me.

Looks like python-jose was the original library used for jwt tokens and that was switched to pyjwt in issue #542 on March 27th.

jasonwhite:
Duplicate of #588. Should be fixed in the next release.

Issue #588, opened 2017-07-05
Integration support does not work in python 3


impala2:
The token (?) created by the jwt library is a `bytes` object in python 3 so instead of the header being `Authorization: Bearer kajshdkshkshkhd` it's `Authorization: Bearer b'kajshdkshkshkhd'`.

Need to update the function like so:

    def create_jwt(self):
        """
        Creates a signed JWT, valid for 60 seconds.
        :return:
        """
        now = int(time.time())
        payload = {
            "iat": now,
            "exp": now + 60,
            "iss": self.integration_id
        }
        encrypted = jwt.encode(
            payload,
            key=self.private_key,
            algorithm="RS256"
        )
        
        if atLeastPython3:
            encrypted = encrypted.decode('utf-8')
            
        return encrypted

I'd make a pull request, but I'm rather confused by how the tests are supposed to be written. To have deterministic results you'd need to mock `time.time()` and you don't seem to have it as part of your test dependencies (not in stdlib in python 2).

jasonwhite:
Should be fixed in the next release.

PR #633, opened 2017-09-20
Fix GithubIntegration in Python3

colinhoglund:
In Python2, `jwt.encode()` returns a `str` object, and in python3 it returns a `bytes` object.

This fix simply performs a `.decode('utf-8')` on the bytes object in Python3 (as described in https://github.com/PyGithub/PyGithub/issues/588). This should also address https://github.com/PyGithub/PyGithub/issues/627.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@
 ################################################################################
 
 *.pyc
+.eggs/
+.python-version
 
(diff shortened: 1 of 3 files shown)


jasonwhite:
Thanks!

PR #651, opened 2017-10-25
Update api calls to loki-preview

babineaum:
Loki-preview is merged into master branch of API v3.

Clean up :)


.github/config.yml
```diff
@@ -0,0 +1,21 @@
+# Configuration for welcome - https://github.com/behaviorbot/welcome
```
sfdye: This file not related to main purpose of the PR, would you mind removing it?

If you would like to add such probot plugins, feel free to open another PR.

sfdye:
I am afraid I have to close this one, as the branches API are now migrated to https://developer.github.com/v3/repos/branches/. So the `get_protected_branch` method is no longer applicable. Instead there is new endpoint for branch protection which needs to be implemented. (https://developer.github.com/v3/repos/branches/#get-branch-protection) Feel free to open a new PR to address this.

PR #655, opened 2017-10-31
Add support for access level for organization repo

NikolaiMagnussen:


NikolaiMagnussen:
This was already added in #699, though there don't seem to be any tests for it.

PR #664, opened 2017-11-26
Use 'requests' instead of 'httplib'

mikeage:
This is an updated version of #383 , with the following changes:
* Resolves conflict in setup.py
* Supports HTTP and HTTPS
* Supports base_url other than github.com
* Uses requests's sessions to further optimize timing
* Proxies are now handled using requests' default handling. This means that the environment variables `$HTTP_PROXY`/`$HTTPS_PROXY`/`$http_proxy`/`$https_proxy` are now used (request seems to prefer the lower case version if both are defined) to specify proxies. A further PR could be added to pass this as a parameter to the calls, if required


mikeage:
I know that this PR does not address the questions about testing and development from the original PR, and so I do not expect it to be merged as-is. 

However, on a local setup, it reduces a flow of 18 calls from 30 seconds to 8 seconds (of which 7 are spent in read(); i.e., network traffic). Without sessions, we went from 30 to 15, which is also a significant improvement.

I'm offering this for anyone who wants, and perhaps as a springboard for someone with more experience to address the testing regression from both this PR and the original.

mikeage:
The failure, btw, is related to `requests` not supporting python 3.2 any more.

jasonwhite:
That sounds like a great performance improvement!

We could probably drop support for Python 3.2.  That release is almost 7 years old after all. Doing that should be a separate PR and might warrant some discussion, however.

I'm in favor of merging this, but it some people should kick the tires a bit more to make sure there are no regressions.

mikeage:
FWIW, I found a reasonable way to continue to support Python 3.2, even though I would argue that it's old enough that there's no reason to maintain support.

jasonwhite:
FYI, there's no need to support those old versions of Python anymore. See #674.

mikeage:
Great. Removing the hack, and syncing with master.

mikeage:
Any thoughts on what additional testing would be necessary or desirable? I can tell you that I've been using this patch for... *checks logs*... several million requests against our internal Github Enterprise server with no regressions

jasonwhite:
Alright, you convinced me!

First, would you mind merging `master` into your branch again? Another PR just got merged (https://github.com/PyGithub/PyGithub/pull/691) and I'm wondering if they will play well together.

mikeage:
Sure, merging and testing now

mikeage:
Works locally, and travis seems happy.

I didn't actually realize I was that persuasive ;)

Credit for this really goes to @nricklin; I just tweaked it a bit, so I'll take any blame.

github/Requester.py
```diff
@@ -51,19 +51,82 @@
...
+        self.text = r.text
+
+    def getheaders(self):
+        return self.headers.iteritems()
+
+    def read(self):
+        return self.text
+
+class HTTPSRequestsConnectionClass(object):
+    # mimic the httplib connection object
```
jasonwhite: Isn't there a cleaner way to do this than by mimicking the httplib classes?

github/Requester.py
```diff
@@ -51,19 +51,82 @@
...
+
+    def request(self, verb, url, input, headers):
+        self.verb = verb
+        self.url = url
+        self.input = input
+        self.headers = headers
+
+    def getresponse(self):
+        r = getattr(self.session,self.verb.lower())("%s://%s:%s%s" % (self.protocol, self.host, self.port, self.url), headers=self.headers, data=self.input, timeout=self.timeout)
+        return RequestsResponse(r)
```
jasonwhite: To make this more readable, I think it would better to use some temporary variables, like so:

```python
def getresponse(self):
    verb = getattr(self.session,self.verb.lower())
    url = "%s://%s:%s%s" % (self.protocol, self.host, self.port, self.url)
    r = verb(url, headers=self.headers, data=self.input, timeout=self.timeout)
    return RequestsResponse(r)
```

Same with `HTTPRequestsConnectionClass.getresponse` below.

mikeage: sure

mikeage:
@jasonwhite , anything else?

jasonwhite:
Sorry, I got a little busy recently. I finally read through the comments on the other PR (#383) and I was wondering if this addresses any of the concerns raised about the test framework?

mikeage:
no problem.

I don't think it does, but to be honest, I don't really understand the test framework here. I would agree that a more standard framework would be better in the abstract, but that's definitely a non trivial task, and one that I don't think I can tackle

The tests pass and still detect failures, but I don't know what would be involved in writing new tests / automatically updating the JSON files if needed.

If you're worried about the original concerns, I agree that this shouldn't be merged, but I'm just hoping we don't wind up stuck with an inferior implementation because of a legacy decision (I realize that sounds a bit passive aggressive; that's not my intention at all)

calebgroom:
`setup.py` doesn't depend on it but since `requirements.txt` exists, shouldn't this PR update it as well?

mikeage:
Can't hurt. Done

nricklin:
ship it!     :)

sfdye:
I really like to see this happen too. 

Just installed your branch:
`pip install git+https://github.com/mikeage/PyGithub.git@requests`

it reduces a complex query from 9.x seconds to 3.x seconds for me 👍

mikeage:
[updated to the last few weeks worth of commits, since I see some good stuff has been added]

jakirkham:
Plus `requests` is just a fantastic library. It is really a joy to work with.

mikeage:
@sfdye , you're an owner now, I see. What do we need to do to get this in?

sfdye:
@mikeage Just so you know, I really want this too. But I am little worried about its impact on the testing framework, so give me some time to review pls.

mikeage:
Sure, no problem. Didn't mean to pressure you too much!

mikeage:
The requests library should support proxies using the standard `$HTTP_PROXY` environment variable. I don't have a suitable proxy to test right now (I will on Monday), but are you saying that this doesn't work?

Or are you asking for support for a custom proxy, like in #568 ?

sfdye:
@mikeage Yes, that's what I meant. If could initialize a Github instance with a custom proxy settings. Also, other env vars like `no_proxy` should be supported as well. If you could test it out on Monday, that would be great!

mikeage:
Ok. Do you think it'd be better to do this in a separate PR, since this functionality doesn't yet exist and therefore it's not a regression, or do you want to see it here to make sure that we don't lose the ability to do it at all?

sfdye:
@mikeage Yeah I agree this could be a separate PR, my intention here is to be thorough when considering migrating from `httplib` to `requests`, so the minimum requirement is that we don't lose the existing support for using this library behind a proxy by setting the `HTTP_PROXY` environment variable.

mikeage:
ok, found a proxy to test. I've updated the first comment to explain how proxies are handled.

github/Requester.py
```diff
@@ -98,6 +98,8 @@ def request(self, verb, url, input, headers):
     def getresponse(self):
         verb = getattr(self.session, self.verb.lower())
         url = "%s://%s:%s%s" % (self.protocol, self.host, self.port, self.url)
+        import pdb
```
sfdye: Did you accidentally commit this?

github/Requester.py
```diff
@@ -123,6 +125,8 @@ def request(self, verb, url, input, headers):
...
```
sfdye: And this?

github/Requester.py
mikeage: Yes, just fixed that.

github/Requester.py
```diff
@@ -365,20 +432,13 @@ def __createConnection(self):
         ## set.
         ## http_proxy: http://user:password@proxy_host:proxy_port
```
sfdye: Probably we can remove these few lines of comments

mikeage:
I think the @TravisBuddy comment is from the commit where I accidentally left in `import pdb / pdb_set_trace()`. The latest build passed.

github/Requester.py
mikeage: Good point. I shouldn't be working at this hour ;)

sfdye:
@mikeage Awesome! I think almost ready to be merged. Let me do some more regression next week and get back to you.

mikeage:
Great, thanks. Let me know if there's anything else. (also note that I pushed one more commit after your last comment)

github/Requester.py
```diff
@@ -67,19 +67,86 @@
...
 
+class RequestsResponse:
+    # mimic the httplib response object
+    def __init__(self, r):
+        self.status = r.status_code
+        self.headers = r.headers
+        self.text = r.text
+
+    def getheaders(self):
+        return self.headers.iteritems()
```
sfdye: `iteritems()` is not available in Python3.

I am afraid we have to do something like:
```
if atLeastPython3:
	return self.headers.items()  # items() in Python3 returns an iterator
else:
	return self.headers.iteritems()
```

https://docs.python.org/3/library/stdtypes.html#dict.items

setup.py
```diff
@@ -104,6 +104,7 @@
         use_2to3=True,
         python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
         install_requires=[
+            "requests>=2.9.1",
```
sfdye: What is the reason to pin at this specific version please?

mikeage: It's not a pin, just a greater than. It's what was in the earlier PR :-)

We could probably leave it empty; I doubt anyone is trying to use a requests library > 3 years old (and if they are, they're probably on python 2.6 or 3.3 or something, and we'll have other problems anyway!)

requirements.txt
```diff
@@ -1 +1,2 @@
+requests>=2.9.1
```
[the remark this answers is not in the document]
mikeage: Yes (although I'm wondering if `python setup.py develop` isn't a better idea than `pip install -r requirements.txt` for a package, in general)

sfdye:
I think I am gonna merge this and cut a alpha release.

Any other concern? @jasonwhite

github/Requester.py
```diff
@@ -67,19 +67,89 @@
...
+    def __init__(self, r):
+        self.status = r.status_code
+        self.headers = r.headers
+        self.text = r.text
+
+    def getheaders(self):
+        if atLeastPython3:
+            return self.headers.items()
+        else:
+            return self.headers.iteritems()
```
jakirkham: So `viewitems` is a bit more accurate/similar to the Python 3 behavior on Python 2.

mikeage: True, although this works on python2 ;-) . I think the only difference is performance, no?

jakirkham: They would both return a `dict_items` object. That would be the main motivator in my mind at least.

sfdye: Since the only place this will be used is:
```
responseHeaders = dict((k.lower(), v) for k, v in response.getheaders())
```
I think the type doesn't really matter.

sfdye:
Would like to keep the ball rolling here, so I will make an alpha release first. If anything breaks or missing, we can make more PRs to fix in subsequent alpha.

sfdye:
Thanks @mikeage for the amazing work!

Please have a try:

`pip install --pre PyGithub==1.40a1`

issues/feedback are welcome

jayfk:
Thanks to everyone working on this 👌🏻

I think the impact on the testing framework is neglible. No one is able to fully understand it anyway. It’s arcane and a thing that’s holding this library back.

I’d love to spend 2-3 days and replace this behemoth with something that’s understandable and easy to use but this won’t happen anytime soon...

glennmatthews:
Noticeably faster performance with 1.40a1. Thank you!

sfdye:
Released `1.40a2`, please have a try before I make the stable release.

kyogi14:
Test recording no longer works as the requests are getting the "fixAuthorizationHeader()" treatment prior to the request actually being made.

PR #668, opened 2017-12-08
Adding unit test for tree attribute of GitCommit

sfdye:
Adding a simple unit test for tree attribute for GitCommit.

### DIFF
--- a/github/Commit.py
+++ b/github/Commit.py
@@ -126,14 +126,6 @@ class Commit(github.GithubObject.CompletableGithubObject):
         self._completeIfNotSet(self._url)
         return self._url.value
 
-    @property
-    def tree(self):
-        """
-        :type: :class:`github.GitTree.GitTree`
-        """
-        self._completeIfNotSet(self._tree)
-        return self._tree.value
-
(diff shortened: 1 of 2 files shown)


github/Commit.py
```diff
@@ -256,3 +265,5 @@ def _useAttributes(self, attributes):
             self._stats = self._makeClassAttribute(github.CommitStats.CommitStats, attributes["stats"])
         if "url" in attributes:  # pragma no branch
             self._url = self._makeStringAttribute(attributes["url"])
+        if "tree" in attributes:  # pragma no branch
+            self._tree = self._makeClassAttribute(Commit, attributes["tree"])
```
jasonwhite: This doesn't really make sense. The tree is not a commit. The JSON also doesn't quite provide enough information to construct a `GitTree` object either (it's missing the subtree elements).

It might make more sense to expose a property called `tree_sha` instead. Then, it could be used in conjunction with `Repository.get_git_tree()`.

sfdye: Seems like I was confusing the Commit with GitCommit, actually the tree attribute is already. So I will just add the missing unit test for this PR.

sfdye:
@jasonwhite how about now?

sfdye:
Thanks for pointing out my mistake earlier 😃

PR #31, opened 2012-05-27
Title edited by PyGithub

jacquev6:
Body edited by PyGithub

PR #670, opened 2017-12-14
PullRequest: add support for getting a single review's comments.

byohay:
Add support for the URL:
/repos/:owner/:repo/pulls/:number/review/:id/comments.

### DIFF
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -448,6 +448,20 @@ class PullRequest(github.GithubObject.CompletableGithubObject):
             None
         )
 
+    def get_single_review_comments(self, id):
+        """
+        :calls: `GET /repos/:owner/:repo/pulls/:number/review/:id/comments <https://developer.github.com/v3/pulls/reviews/>`_
+        :param id: integer
+        :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.PullRequestComment.PullRequestComment`
+        """
+        assert isinstance(id, (int, long)), id
+        return github.PaginatedList.PaginatedList(
+            github.PullRequestComment.PullRequestComment,
+            self._requester,
+            self.url + "/reviews/" + str(id) + "/comments",
+            None
+        )
+
     def get_commits(self):
         """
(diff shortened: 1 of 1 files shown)


github/PullRequest.py
```diff
@@ -448,6 +448,18 @@ def get_review_comments(self):
             None
         )
 
+    def get_single_review_comments(self, id):
+        """
+        :calls: `GET /repos/:owner/:repo/pulls/:number/review/:id/comments <https://developer.github.com/v3/pulls/reviews/>`_
+        :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.PullRequestComment.PullRequestComment`
+        """
+        return github.PaginatedList.PaginatedList(
```
jasonwhite: There should be an `assert` to ensure that `id` is an integer:

```python
assert isinstance(id, (int, long)), id
```

jasonwhite: The `id` param should also be in the doc string, same as with `get_review_comment` further up in the file.

jasonwhite: A test for this would also be great, too.

byohay: 1. Fixed the first two comments.
2. I wanted to add a test, but saw that the PR being used in the `PullRequest` tests (#31) doesn't have any reviews. What do you suggest to do?

jasonwhite: > Fixed the first two comments.

Did you forget to push those changes?

byohay: @jasonwhite now I pushed the changes :)

sfdye:
@byohay I still don't see any tests

sfdye:
I could add a review for that PR:
https://api.github.com/repos/PyGithub/PyGithub/pulls/31/reviews/105701257/

but I couldn't add a comment for the review, maybe because the PR is already closed?  Given all other PullRequest tests are tied to the replayData for this old PR, I don't see the point to make additional effort just to pass this test. I have cloned and tested your branch and it's working. So approving this PR.

PR #505, opened 2016-12-17
Add source import 

usjeong:
Plz we need this for migration

PR #673, opened 2017-12-18
Source Import API

brix4dayz:
- `PUT` and `GET` for `/repos/:owner/:repo/import` according to the [preview docs](https://developer.github.com/v3/migration/source_imports/)
- Recycled from #505


github/SourceImport.py
```diff
@@ -0,0 +1,163 @@
+import github.GithubObject
```
jasonwhite: This file is missing the Python source encoding comment (`# -*- coding: utf-8 -*-`) and copyright header.

github/tests/SourceImport.py
```diff
@@ -0,0 +1,28 @@
+
```
jasonwhite: Same with this file.

github/SourceImport.py
```diff
@@ -0,0 +1,190 @@
+
+# -*- coding: utf-8 -*-
```
jasonwhite: I believe this needs to be the first line in the file.

sfdye:
Thank you very much for your contribution @brix4dayz

PR #678, opened 2017-12-28
Add committer/author to Repository.delete_file

ikonst:
As per spec:
https://developer.github.com/v3/repos/contents/#optional-parameters-2


ikonst:
@jasonwhite can you have a look?

Issue #687, opened 2018-01-25
PullRequestReviewerRequest should be updated


gshefer:
The payload of GET /repos/:owner/:repo/pulls/:number/requested_reviewers has changed:
https://developer.github.com/changes/2018-01-24-end-thor-preview/
From Jan.24 2018 the module raises as a result:
AtributeError: PullRequestReviewerRequest has no attribute '_url'

We should either pass 'users' as list_item arg to the PullRequestReviewerRequest module or changed the entire module design.

PR #690, opened 2018-01-25
Updated github.PullRequestReviewerRequest

gshefer:
Updated github.PullRequestReviewerRequest according to the API changes: https://developer.github.com/changes/2018-01-24-end-thor-preview/

Addresses issue: https://github.com/PyGithub/PyGithub/issues/687

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -175,7 +175,7 @@ class PaginatedList(PaginatedListBase):
         self.__nextParams = None
 
         if self.__list_item in data:
-            self.__totalCount = data['total_count']
+            self.__totalCount = data.get('total_count')
             data = data[self.__list_item]
 
         content = [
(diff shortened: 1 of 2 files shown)


github/PaginatedList.py
```diff
@@ -175,7 +175,7 @@ def _fetchNextPage(self):
         self.__nextParams = None
 
         if self.__list_item in data:
-            self.__totalCount = data['total_count']
+            self.__totalCount = data.get('total_count')
```
jasonwhite: How does this change relate to the other one?

gshefer: I'm not sure why you decided that this key exists if the if self.__list_item in data. in this case it doesn't, so I'm using .get to not fail, and it works fine (anyway it's None by default).

jasonwhite: I see. Then, to be consistent, the same change should also be made to the `get_page` method.

PR #691, opened 2018-01-27
Remove the default "null" input send during GET request

R1kk3r:
**Background**
I used the library for automating a process on a GHE instance. Unfortunately, simple GET requests were always rejected. After analysing the raw request, I figured out that when you use the Github API that don't need any input (/pulls, /user, /repos, etc) the string "null" was appended to the request, making it malformed and invalid.

A quick test, the following example failed systematically

```python
import github
github.enable_console_debug_logging()
github.Github("<token>", base_url="https://<ghe_instance>/api/v3").get_user().get_repos()[0]
```
And the logging showed:

> GET https://<ghe_instance>/api/v3/user/repos {'Authorization': 'token (oauth token removed)', 'User-Agent': 'PyGithub/Python'} null ==> 400 {'content-type': 'text/html; ...

We can see the "null" value appended to the request, which is wrong ! I guess the reaction depends on the HTTP server software. At least in my company, it doesn't like it at all.

**Changes**
- Removed the default value for input "null" by None. We really want to send no input when we have nothing to send.

**Test**
Rerun the example above, everything is fine.


jasonwhite:
Thanks for the PR, @R1kk3r. It looks like the Travis CI build failed for only Python 2.7. Mind taking a look?

R1kk3r:
@jasonwhite Done ! Of course, I forgot to update ReplayDatas to reflect the change from 'null' to 'None'.

jasonwhite:
I'm wondering why it was set to `"null"` instead of `None` [in the first place](https://github.com/R1kk3r/PyGithub/commit/681128d41a43c1d54ea44a1a1630db59fb4d63dd).

The only difference I see is that `Content-Length` is `0` versus `4`:
 * https://docs.python.org/2/library/httplib.html#httplib.HTTPConnection.request
 * https://docs.python.org/3/library/http.client.html#http.client.HTTPConnection.request

Hope this change doesn't break something else...

PR #692, opened 2018-01-30
Add a member to an organization given a NamedUser

MattPhillips1:
Problem: A missed endpoint from the API. `PUT /orgs/:org/memberships/:username`

Solution: Add an easy method that takes a NamedUser and sends to this endpoint. An email invitation is then sent to the NamedUser account

See https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership for the api docs

### DIFF
--- a/github/Organization.py
+++ b/github/Organization.py
@@ -291,16 +291,22 @@ class Organization(github.GithubObject.CompletableGithubObject):
             self.url + "/public_members/" + public_member._identity
         )
 
-    def add_membership(self, member):
+    def add_membership(self, member, role=github.GithubObject.NotSet):
         """
         :calls: `PUT /org/:org/memberships/:username <https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership>`_
         :param member: :class:`github.NamedUser.NamedUser`
+        :param role: string
         :rtype: None
         """
         assert isinstance(member, github.NamedUser.NamedUser), member
+        assert role is github.GithubObject.NotSet or isinstance(role, (str, unicode)), role
+        put_parameters = {}
+        if role is not github.GithubObject.NotSet:
+            put_parameters["role"] = role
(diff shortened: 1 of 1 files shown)


jasonwhite:
Thanks for the PR @MattPhillips1. Would you mind adding support for the `role` parameter and adding a test for this?

github/Organization.py
```diff
@@ -291,6 +291,18 @@ def add_to_public_members(self, public_member):
             self.url + "/public_members/" + public_member._identity
         )
 
+    def add_member(self, member):
```
jasonwhite: I think the name should be `add_membership` to be consistent with the [`Team` API](https://github.com/PyGithub/PyGithub/blob/cbfe8d0f623ca29d984ec09d2b566e9ab10ae024/github/Team.py#L128).

Constantin07:
@MattPhillips1 could you please add the below test case (patch) to PR?

```
---
 github/tests/Organization.py | 7 +++++++
 1 file changed, 7 insertions(+)

diff --git a/github/tests/Organization.py b/github/tests/Organization.py
index e7948c69..9a2a71db 100644
--- a/github/tests/Organization.py
+++ b/github/tests/Organization.py
@@ -97,6 +97,13 @@ class Organization(Framework.TestCase):
     def testGetPublicMembers(self):
         self.assertListKeyEqual(self.org.get_public_members(), lambda u: u.login, ["jacquev6"])

+    def testAddMembership(self):
+        lyloa = self.g.get_user("Lyloa")
+        self.org.add_membership(lyloa)
+        self.assertTrue(self.org.has_in_members(lyloa))
+        self.org.remove_from_members(lyloa)
+        self.assertFalse(self.org.has_in_members(lyloa))
+
     def testGetIssues(self):
         self.assertListKeyEqual(self.org.get_issues(), lambda i: i.id, [])

--
2.16.2
```

PR #698, opened 2018-02-07
Method create_milestone now uses datetime object

btorresgil:
This fits better with v3 API which uses ISO8601 date/time format with timezone.

Date object still allowed for backward compatibility

The problem with current implementation is it always passes UTC to the API.  So for example, if you're in the United States and pass a `date` object like `date(2018, 2, 4)`, then GitHub will interpret that as 2018-02-04T00:00:00Z which is actually 2018-02-03 in the United States.  So the due date in the milestone will be the day before the one you passed in.  This is incorrect behavior.

Since GitHub API needs expects a date/time in ISO 8601 format, it would be best to pass in a `datetime` object and call `.isoformat()` on the object, than to bother with `date` objects.  This will also be more compatible with enhanced datetime libraries like [pendulum](https://pendulum.eustace.io/).

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -968,13 +968,13 @@ class Repository(github.GithubObject.CompletableGithubObject):
         :param title: string
         :param state: string
         :param description: string
-        :param due_on: date
+        :param due_on: datetime
         :rtype: :class:`github.Milestone.Milestone`
         """
         assert isinstance(title, (str, unicode)), title
         assert state is github.GithubObject.NotSet or isinstance(state, (str, unicode)), state
         assert description is github.GithubObject.NotSet or isinstance(description, (str, unicode)), description
-        assert due_on is github.GithubObject.NotSet or isinstance(due_on, datetime.date), due_on
+        assert due_on is github.GithubObject.NotSet or isinstance(due_on, datetime.date) or isinstance(due_on, datetime.datetime), due_on
         post_parameters = {
             "title": title,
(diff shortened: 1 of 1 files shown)


github/Repository.py
```diff
@@ -968,13 +968,13 @@ def create_milestone(self, title, state=github.GithubObject.NotSet, description=
...
         :param description: string
-        :param due_on: date
+        :param due_on: datetime
         :rtype: :class:`github.Milestone.Milestone`
         """
         assert isinstance(title, (str, unicode)), title
         assert state is github.GithubObject.NotSet or isinstance(state, (str, unicode)), state
         assert description is github.GithubObject.NotSet or isinstance(description, (str, unicode)), description
-        assert due_on is github.GithubObject.NotSet or isinstance(due_on, datetime.date), due_on
+        assert due_on is github.GithubObject.NotSet or isinstance(due_on, datetime.date) or isinstance(due_on, datetime.datetime), due_on
```
jasonwhite: It would be cleaner to condense this:
```python
isinstance(due_on, datetime.date) or isinstance(due_on, datetime.datetime)
```
into this:
```python
isinstance(due_on, (datetime.date, datetime.datetime))
```

btorresgil: Thanks for the review.  Changed syntax per your recommendation in commit 7c05c04ef6d903bf2f248a01f85333349b7d0486.

Issue #611, opened 2017-08-07
Add ability to set visibility of organization team


rpocase:
When creating or managing a team, the only privacy level is "secret". The `Organization.create_team` API should be extended to accept `privacy` of value either `secret` (default) or `closed`.

Relevant API spec: https://developer.github.com/v3/orgs/teams/#create-team

sfdye:
Closed in #702

PR #702, opened 2018-02-13
Add team privacy parameter to create team

res0nance:
Implements #611 

Replay data is still not correct as I was unable to to run with the record parameter leading to issues such as attempting to write a str object to a binary file. Input being None for RecordingConnection.request. Fixing the above 2 led to a 404 error. Could someone direct me on how to properly obtain the recording data?


calebgroom:
@res0nance Can you pull my commit into your branch? It should sort out your testing failures.

res0nance:
@calebgroom Thanks it seems to have fixed it.

It seems like the testing framework does not support python 3 and some tests seem to be broken when trying to record. I'm not certain if there is an issue for that yet.

grahamlyons:
Very glad to see someone has tackled this - thank you.

sfdye:
Thanks @res0nance and @calebgroom !

PR #704, opened 2018-02-15
make datetime objects timezone aware using UTC timezone (+tests)

### LINKED ISSUE #512: Datetime returned is naive, should be UTC
If I query an issue and access ```issue.created_at.tzinfo```, it is ```None```, indicating a naive datetime object

The values always appear to be UTC values so the ```tzinfo``` should be set to UTC (offset 0)

If the values are naive, all kinds of unusual bugs can occur in programs that try to insert the values into databases or other systems that assume a naive datetime is in local time.

euxneks:
Fixes #512 

Currently, datetime objects returned by PyGithub are not timezone aware, even though times returned through the APIv3 denote UTC (e.g. `2018-02-08T18:46:46Z`) - I have modified PyGithub so that it adds the necessary UTC timezone information wherever datetime is used (that I can see). As far as I can tell, all timezone information returned from the Github API using PyGithub have no timezone information in the headers, and therefore all JSON responses are in UTC, again, as far as I can tell.

I have also modified the datetime comparisons in all tests so they are aware of UTC timezone as well - with these changes all tests pass with python2 and python3.

I have also run `pep8 --ignore=E501 github` (with python2) as recommended in CONTRIBUTING.md (there were a couple left over from the master branch which I fixed). It should pass completely now.

I have written this with the assumption that we don't want to add any other libs or modules like `pytz`, mainly because I also assumed all datetimes sent to us from github are UTC, so we don't need to rely upon a huge lib like that.

Feel free to reject this if you want, I saw a #512 and figured I could help solve it :)


euxneks:
I just noticed you guys use the gitflow naming scheme for branches. dangit thought I had all bases covered! :stuck_out_tongue:

sfdye:
I think this makes sense. We should be timezone aware. What do you think? @jasonwhite

euxneks:
I just want to be clear that this adds UTC as a timezone in all datetime objects. If there somehow is a timezone from github, this will **overwrite** the timezone with UTC - not convert it. (note  datetime.replace on the datetime objects [around here](https://github.com/PyGithub/PyGithub/pull/704/files#diff-1300f4cc7926b75e0ac22c373c5c7814R183))

Though, I think this is pretty safe - I _think_ according to the github api documentation we have to send a special header to get the datetime with a timezone other than UTC, though the documentation seems to be a bit unclear on this. I refer to this github blog post from 2014: 
https://developer.github.com/changes/2014-03-04-timezone-handling-changes/
and also api v3 documentation:
https://developer.github.com/v3/#timezones

allevin:
@euxneks I want to make sure that this change will be backwards compatible with existing code.  I am getting the time and am using pytz to convert it to my local timezone in my code.

docktermj:
Has this been fixed?

PR #706, opened 2018-02-20
Add __eq__ and __hash__ to NamedUser

### LINKED ISSUE #608: Implement __hash__() and __eq__() on NamedUser
I have a case where I'm processing the list of reviewers on a Pull Request. When the same person does two reviews the `NamedUser` return from the `PullRequest.get_reviews()`, while they have the same `login`, `id`, and `name`, don't have the same hash value. 

If you implement `__hash__()` the python docs strongly recommend implementing `__eq__()` based upon the same values. 

It would be nice to have the python build in functions be able to construct the unique list of reviewers.

namratachaudhary:
Resolves #608

### DIFF
--- a/github/NamedUser.py
+++ b/github/NamedUser.py
@@ -51,8 +51,8 @@ class NamedUser(github.GithubObject.CompletableGithubObject):
 
     def __eq__(self, other):
         return (isinstance(other, type(self))
-                and (self.login == other.login)
-                and (self.id == other.id))
+                and self.login == other.login
+                and self.id == other.id)
 
     @property
     def avatar_url(self):
--- a/github/NamedUser.py
+++ b/github/NamedUser.py
@@ -46,6 +46,14 @@ class NamedUser(github.GithubObject.CompletableGithubObject):
     def __repr__(self):
         return self.get__repr__({"login": self._login.value})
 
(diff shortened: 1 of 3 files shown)


github/NamedUser.py
```diff
@@ -46,6 +46,14 @@ class NamedUser(github.GithubObject.CompletableGithubObject):
     def __repr__(self):
         return self.get__repr__({"login": self._login.value})
 
+    def __hash__(self):
+        return hash((self.id, self.login))
+
+    def __eq__(self, other):
+        return (isinstance(other, type(self))
```
sfdye: Thanks for the change. Even the most outer pair of brackets are not need.
```python
return isinstance(other, type(self)) and self.login == other.login and self.id == other.id)
```

sfdye:
@namratachaudhary LGTM, I will merge once you remove the brackets.

github/NamedUser.py
namratachaudhary: Ah! The linter was throwing an error, so I hesitated. Will make the change nevertheless

sfdye: What did the linter say?

sfdye:
@namratachaudhary Merged. Thank you for your contribution!

PR #709, opened 2018-02-24
Enforce the Host Header When Using Proxies

jeff-minard-ck:
I found that, when using a proxy, it was sending this value:

```
send: 'CONNECT api.github.com:443 HTTP/1.0\r\n'
send: '\r\n'
send: 'GET /repos/group/repo/pulls/7 HTTP/1.1\r\nHost: proxy.host.name.com:3128\r\nAccept-Encoding: identity\r\nAuthorization: token REDACTED\r\nUser-Agent: PyGithub/Python\r\n\r\n'
reply: 'HTTP/1.0 400 Bad Request\n'
```

Github did not like the `Host` header _not_ being api.github.com and would return bad things.

This patch forces the header to the correct value.


jeff-minard-ck:
aaaaand I've just realized why this fails Travis so badly; every file which has recorded behavior is checking for headers and I've just extended the list, so none of these pass anymore.

I'll have a crack at updating them.

sfdye:
@jeff-minard-ck Thanks for your PR. I am closing this in favor of #732 and #568

jeff-minard-ck:
I'm not quite clear on how those other two PR's fix the problem. The main thrust of this PR was (d8a1adc7fe425c8b461ae85476afe0261fc75c50) which adds one line

```
requestHeaders["Host"] = self.__hostname
```

that, without it, causes the proxy go get the wrong host header.

I'm, however, happy to wait for those other PRs to merge and re-run the code in my proxied env.

_(#664 would also handily fix this problem too)_

sfdye:
Sure, I understand your problem now. #664 is the priority now, I will release an alpha version soon so you can help me test it. Cool?

jeff-minard-ck:
Works for me!

sfdye:
Please have a try:
`pip install --pre PyGithub==1.40a1`

Issue #717, opened 2018-03-02
Comments do not include reply info


JessMorgan:
assert github.Github().get_user().get_repo("PyGithub").get_pull(664).get_comment(166456140).in_reply_to_id == "166453895"

Currently, in_reply_to_id is undefined.  This makes it impossible to understand the comment threading.

PR #718, opened 2018-03-02
Add PullRequestComment.in_reply_to_id

JessMorgan:
Fix for issue #717


sfdye:
@JessMorgan 
Thank you very much for your PR. It will be so helpful if you can add a unit test.

PR #726, opened 2018-03-08
Add organization add_membership, remove_membersip and get_invitations…

hile:
The organization API was missing a couple of calls to manage invitations to organization.


sfdye:
Hi @hile Thank you very much for your PR. It will be so helpful if you can add a unit test.

PR #662, opened 2017-11-21
Added option to create review for Pull request

gshefer:
Added option to create review for Pull request.
With this code you have the ability to create a review object and also specify the event (APPROVE, REQUIRE_CHANGES, ...) which you don't have in reviewer request.

PR #731, opened 2018-03-16
Add url attribute to PullRequestReview object

electrofelix:
To compare PullRequestReview objects, they need to have the 'url'
attribute because the CompletableGithubObject expects to be able to
inspect a '_url' attribute on objects when testing for equality.


electrofelix:
This is needed by #662

sfdye:
Thanks @electrofelix

PR #732, opened 2018-03-16
Support recording test data from behind proxies

electrofelix:
When recording test data from behind a proxy, make sure to support
set_tunnel on the original connection and write out the correct
settings from the httplib connection rather than the proxy settings.


sfdye:
@electrofelix Very useful. If you could provide some test evidence, I will merge!

sfdye:
This should be closed by #664

electrofelix:
That's a much better solution, many thanks, as i didn't have time to further improve this

PR #741, opened 2018-04-05
Add add_to_members() and remove_from_membership()

### LINKED ISSUE #399: Add support for adding/updating organization membership

tjboring:
This PR provides the code and unittests to support adding a new member to an organization. Some things to note:

1. The code and unittests are not original. They were copied and slightly modified from the following PR/branches:
https://github.com/vmagro/PyGithub/commit/dc21acfcb48a8b71cfd39a2b6ea8ff4b149f5462
https://github.com/jwplayer/PyGithub/commit/3fb0e82d71bb112f3d4632e6b5883fd98d680866
https://github.com/jbweston/PyGithub/commit/c1ce18a24632e8d4b0752b19ab3fec9806e54593

- Closes #424 
- Closes #726 
- Closes #399

2. The `BeaverSoftware` Github organization no longer exists. To work around this issue, I used another organization that I belong to for the purpose of testing. I am not, however, submitting the replay data that was generated locally as a result of that. The two included tests passed when run locally using the other org I belong to.

3. When a new member is added to an organization, the state of their membership is set to `pending`. As a result, they are not returned in the list of members returned by the call the `/orgs/:org/members/:user` endpoint. Instead, their status can be checked by calling `/orgs/:org/memberships/:user`.


tjboring:
FYI...the travis failure is due to the second point in the PR description above.

tjboring:
@sfdye i figured out how to scrub the sensitive stuff from the replay data that my local testing generated. please take a look.

github/Organization.py
```diff
@@ -726,6 +760,19 @@ def has_in_public_members(self, public_member):
         )
         return status == 204
 
+    def remove_from_membership(self, member):
+        """
+        :calls: `DELETE /orgs/:org/memberships/:user <http://developer.github.com/v3/orgs/memberships>`_
```
sfdye: This url seems to point to nowhere

github/Organization.py
```diff
@@ -726,6 +760,19 @@ def has_in_public_members(self, public_member):
...
+        :param member: :class:`github.NamedUser.NamedUser`
+        :rtype: None
+        """
+        assert isinstance(member, github.NamedUser.NamedUser), member
```
sfdye: What exactly is the difference between `remove_from_members` and `remove_from_membership`?

If I understand correctly, the only difference is that `remove_from_membership`  is for non-active members, in which case this method will cancel the invitation?

github/Organization.py
```diff
@@ -713,6 +731,22 @@ def has_in_members(self, member):
             )
         return status == 204
 
+    def has_in_pending_members(self, member):
```
sfdye: Since you are referencing this endpoint:
https://developer.github.com/v3/orgs/members/#get-organization-membership

Why not implement a more generic `get_membership` method instead of this?

sfdye:
@tjboring Looks good! Please see my comments.

github/Organization.py
tjboring: I was following the pattern used in the tests, and discovered that the `has_in_members` method doesn't include members with a state of `pending`. So implementing a similar method for pending members seemed like the better way to go rather than to try to modify `has_in_members` to also handle pending members.

Also, we could implement a more generic `get_membership` method, but the returned Json from the call is different from that returned from the call to `/Users/:user`, meaning we'd either have to return it as just a regular python dict or write a new class to encapsulate the data. The latter is certainly a legitimate option, but I think that's beyond the scope of what this PR is attempting to achieve; we could always open an issue to have that method implemented in a subsequent PR.

github/Organization.py
tjboring: hmm, likely copy/paste error on my part. fixed.

github/Organization.py
tjboring: Correct. Per the [docs](https://developer.github.com/v3/orgs/members/#remove-organization-membership):

```
If the specified user is an active member of the organization, this will remove them from
the organization. If the specified user has been invited to the organization, this will cancel 
their invitation. The specified user will receive an email notification in both cases.
```

tjboring:
@sfdye responded. let me know what you think.

sfdye:
@tjboring Thank you very much for the effort!

Aricg:
Correct me if I'm wrong, but this does not implement the api. 
The API:
DELETE /teams/:team_id/memberships/:username
Implementation:
Parameters: | member – github.NamedUser.NamedUser

I cannot pass team_id so I am forced to remove user from all teams.

PR #743, opened 2018-04-07
documentation: clarify semantics of get_comments

### LINKED ISSUE #707: get_comments does not return the comments created with create_issue_comments
I notice that calling `get_comments` on a PR object does not return the comments created with `create_issue_comment`, it only returns the user-based comments made with the web interface. Similarly, `get_issue_comments` does not return the API-created comments.

This is quite counter-intuitive.

Is that on purpose?

monperrus:
fix #707

### DIFF
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -480,6 +480,7 @@ class PullRequest(github.GithubObject.CompletableGithubObject):
 
     def get_comments(self):
         """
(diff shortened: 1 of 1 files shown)


sfdye:
Thanks! @monperrus

adbridge:
I've just hit this issue. I don't think the proposed solution is actually correct. Using get_issues_comment() how do you know what issue a particular comment returned as associated with ? So in my case where I want all the comments for a particular issue how does this help?

PR #758, opened 2018-04-17
Add verify option for to requests wrapper

### LINKED ISSUE #746: add ignore SSL option for github enterprise accessibility
Please provide a 'verify=False' kinda flag for accessing enterprise servers with self signed certificates.

mikeage:
Closes #746

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -81,7 +81,7 @@ class Github(object):
     This is the main class you instantiate to access the Github API v3. Optional parameters allow different authentication methods.
     """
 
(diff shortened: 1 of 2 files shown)


sfdye:
Looks good. Were you able to test this?

mikeage:
Yes, but only with giving a `base_url` to a non-Github site. Our GHE server has a proper certificate, so instead, I tried connecting to https://self-signed.badssl.com and confirmed that it gave an SSL error by default, but a 404 when I passed `verify=False`.

mikeage:
@pinakc, can you test this on your environment?

sfdye:
I've tested this in my GHE instance, works well. Also playing nice with `github.com`.

PR #768, opened 2018-04-24
Fix github API requests after asset upload

mfonville:
In the old code the `self.__hostname` would be overwritten with `uploads.github.com`
but it could not be correctly re-set to `api.github.com` after completing the upload
By handling the situation of connections to `https://uploads.github.com` the same way
we handle `status.github.com` we can resolve this problem.

The old code results if first an `upload_asset()` and then a regular api method in the following trace:
```
  File "github/Requester.py", line 185, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
  File "github/Requester.py", line 234, in requestJson
    return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
  File "github/Requester.py", line 279, in __requestEncode
    url = self.__makeAbsoluteUrl(url)
  File "github/Requester.py", line 351, in __makeAbsoluteUrl
    assert o.hostname in [self.__hostname, "uploads.github.com"], o.hostname
AssertionError: api.github.com
```

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -188,8 +188,6 @@ class Requester:
         return self.__check(*self.requestMultipart(verb, url, parameters, headers, input))
 
     def requestBlobAndCheck(self, verb, url, parameters=None, headers=None, input=None):
-        o = urlparse.urlparse(url)
-        self.__hostname = o.hostname
         return self.__check(*self.requestBlob(verb, url, parameters, headers, input))
 
     def __check(self, status, responseHeaders, output):
@@ -272,6 +270,8 @@ class Requester:
         if self.__apiPreview:
             requestHeaders["Accept"] = "application/vnd.github.moondragon+json"
 
+        if url.startswith("https://uploads.github.com"):
+            cnx = "uploads"
         url = self.__makeAbsoluteUrl(url)
         url = self.__addParametersToUrl(url, parameters)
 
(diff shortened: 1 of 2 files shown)


github/Requester.py
```diff
@@ -188,8 +188,6 @@ def requestMultipartAndCheck(self, verb, url, parameters=None, headers=None, inp
         return self.__check(*self.requestMultipart(verb, url, parameters, headers, input))
 
     def requestBlobAndCheck(self, verb, url, parameters=None, headers=None, input=None):
```
sfdye: I am more inclined to add a keyword argument `cnx` to `requestBlobAndCheck` here

github/Requester.py
```diff
@@ -272,6 +270,8 @@ def __requestEncode(self, cnx, verb, url, parameters, requestHeaders, input, enc
         if self.__apiPreview:
             requestHeaders["Accept"] = "application/vnd.github.moondragon+json"
 
+        if url.startswith("https://uploads.github.com"):
```
sfdye: So we don't need this check here

mfonville:
Thanks for your hints, I made an updated PR at https://github.com/PyGithub/PyGithub/pull/771

This update will then also replace the current method the connections to `status.github.com` are handled

mfonville:
Superseded by #771

Issue #80, opened 2012-09-12
Assertion Failure ??


mnsanghvi:
``` python
>>> from github import Github
>>> gh = Github( "login", "password", "base_url")
>>> for repo in gh.get_user().get_repos():
...         print repo.name
...
repo1
repo2
my-person-linux-kernel-repo
the-secret-macosx-repo

>>> for repo in gh.get_organization( org_name ).get_repos():
...         print repo.name
...

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "build/bdist.macosx-10.7-intel/egg/github/Organization.py", line 311, in get_repos
  File "build/bdist.macosx-10.7-intel/egg/github/Requester.py", line 60, in requestAndCheck
  File "build/bdist.macosx-10.7-intel/egg/github/Requester.py", line 76, in requestRaw
AssertionError

```

Is this because of my lack of knowledge about python, or is there something else going on here ? 

What I would like to do is be able to get a list of the hooks in all the repos in an organization and also go through that list of repos in an organization and be able to create hooks.

PR #771, opened 2018-04-25
Fix github API requests after asset upload

mfonville:
In the old code the self.__hostname would be overwritten with uploads.github.com
but it could not be correctly re-set to api.github.com after completing the upload
Create a separate connection if hostname or port differ in requestBlobAndCheck

in the end this became quite a large overhaul, to also make this change generic
for e.g. connecting to status.github.com and similar methods

~~not sure if tests need (more) updating, if so I will update the PR accordingly~~

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -72,6 +72,7 @@ import Invitation
 atLeastPython3 = sys.hexversion >= 0x03000000
 
 DEFAULT_BASE_URL = "https://api.github.com"
+DEFAULT_STATUS_URL = "https://status.github.com"
 # As of 2018-05-17, Github imposes a 10s limit for completion of API requests.
 # Thus, the timeout should be slightly > 10s to account for network/front-end
 # latency.
@@ -625,8 +626,7 @@ class Github(object):
         """
         headers, attributes = self.__requester.requestJsonAndCheck(
             "GET",
-            "/api/status.json",
-            cnx="status"
+            DEFAULT_STATUS_URL + "/api/status.json"
         )
         return Status.Status(self.__requester, headers, attributes, completed=True)
 
@@ -639,8 +639,7 @@ class Github(object):
         """
         headers, attributes = self.__requester.requestJsonAndCheck(
             "GET",
-            "/api/last-message.json",
-            cnx="status"
+            DEFAULT_STATUS_URL + "/api/last-message.json"
         )
         return StatusMessage.StatusMessage(self.__requester, headers, attributes, completed=True)
 
@@ -653,8 +652,7 @@ class Github(object):
         """
         headers, data = self.__requester.requestJsonAndCheck(
             "GET",
-            "/api/messages.json",
-            cnx="status"
+            DEFAULT_STATUS_URL + "/api/messages.json"
         )
         return [StatusMessage.StatusMessage(self.__requester, headers, attributes, completed=True) for attributes in data]
 
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -254,22 +254,28 @@ class Requester:
         self.__apiPreview = api_preview
         self.__verify = verify
 
-    def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
-        return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
+    def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None):
+        return self.__check(*self.requestJson(verb, url, parameters, headers, input, self.__customConnection(url)))
 
     def requestMultipartAndCheck(self, verb, url, parameters=None, headers=None, input=None):
-        return self.__check(*self.requestMultipart(verb, url, parameters, headers, input))
+        return self.__check(*self.requestMultipart(verb, url, parameters, headers, input, self.__customConnection(url)))
 
     def requestBlobAndCheck(self, verb, url, parameters=None, headers=None, input=None):
-        o = urlparse.urlparse(url)
-        self.__hostname = o.hostname
-        return self.__check(*self.requestBlob(verb, url, parameters, headers, input))
+        return self.__check(*self.requestBlob(verb, url, parameters, headers, input, self.__customConnection(url)))
 
     def __check(self, status, responseHeaders, output):
         output = self.__structuredFromJson(output)
         if status >= 400:
             raise self.__createException(status, responseHeaders, output)
         return responseHeaders, output
+
+    def __customConnection(self, url):
+        cnx = None
+        if not url.startswith("/"):
+            o = urlparse.urlparse(url)
+            if o.hostname != self.__hostname or o.port != self.__port:
+                cnx = self.__httpsConnectionClass(o.hostname, o.port)
+        return cnx
 
     def __createException(self, status, headers, output):
         if status == 401 and output.get("message") == "Bad credentials":
@@ -303,7 +309,7 @@ class Requester:
 
         return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
 
-    def requestMultipart(self, verb, url, parameters=None, headers=None, input=None):
+    def requestMultipart(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
         def encode(input):
             boundary = "----------------------------3c3ba8b523b2"
             eol = "\r\n"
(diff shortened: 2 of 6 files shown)


mfonville:
@sfdye it's now done, including the tests and should be ready to merge :-)

mfonville:
My last commit, a small subsequent improvement, made it fail for the workaround that was introduced for issue #80, I am now trying to include a fix for that too.

mfonville:
Should now be good to go :-)

github/Requester.py
```diff
@@ -251,23 +251,34 @@ def __init__(self, login_or_token, password, base_url, timeout, client_id, clien
         self.__userAgent = user_agent
         self.__apiPreview = api_preview
 
-    def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
-        return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
+    def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None):
```
sfdye: There are still some usage of `requestJsonAndCheck()` that uses the `cnx` argument, you might wanna refactor those as well.

mfonville: I checked the source code but I can't find them anywhere else that I didn't refactor yet (and that was the `status.github.com` stuff)

sfdye:
This PR is rather large, will it affect Github enterprise?

mfonville:
Afaik there should be no impact on Github Enterprise, though I don't have any experience or testing with GH enterprise myself.

The only major change in functionality is the actual important fix (that `self.__hostname` is not overwritten with `uploads.github.com'). For the rest is the code overhaul should not change any functional behavior.

mfonville:
(and to be honest, if Github enterprise would start to offer a status-api and an uploads-api, this refactor will be a great first step to be able to implement that too)

mfonville:
Could this PR be included in any upcoming release, so that we can use the stable branch without being confronted by the upload asset bug?

sfdye:
@mfonville I am cool with that. Once it's merged I will cut a new release, together with the requests change. But please give me some time to review it, got busy with work recently.

mfonville:
fyi, I am now using this code ( including #787 ) in production to upload more than 200 assets per day to Github

mfonville:
I rebased the PR on the latest master (to accommodate for the Content-Length patch)

mfonville:
Pity to see that it didn't make it into today's alpha release; what could I do to help to get this important fix included? Because asset uploading is broken with all current stable and alpha releases... (and I will rebase again on master to fix the merge conflict)

sfdye:
Because this PR is touching some important underlying files (e.g. `Requester.py`), I am little concerned about the impact. What you can help is to convince me, that this change is backward compatible and won't break anything.

mfonville:
Ok, I can guarantee you that it is backward compatible, and I am using it successfully in production. The change is indeed a bit invasive, but that was based on the feedback for #768 .

I will write some comments within the PR to highlight why I made the subsequent change, and why it won't break anything compared to the old situation.

github/MainClass.py
```diff
@@ -639,8 +639,7 @@ def get_last_api_status_message(self):
         """
         headers, attributes = self.__requester.requestJsonAndCheck(
             "GET",
-            "/api/last-message.json",
-            cnx="status"
+            DEFAULT_STATUS_URL + "/api/last-message.json"
```
mfonville: Passing here the full URL, see comment above

github/MainClass.py
```diff
@@ -653,8 +652,7 @@ def get_api_status_messages(self):
         """
         headers, data = self.__requester.requestJsonAndCheck(
             "GET",
-            "/api/messages.json",
-            cnx="status"
+            DEFAULT_STATUS_URL + "/api/messages.json"
```
mfonville: Passing here the full URL, see comment above

github/Requester.py
```diff
@@ -254,23 +254,34 @@ def __init__(self, login_or_token, password, base_url, timeout, client_id, clien
         self.__apiPreview = api_preview
         self.__verify = verify
 
-    def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
-        return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
+    def requestJsonAndCheck(self, verb, url, parameters=None, headers=None, input=None):
+        return self.__check(*self.requestJson(verb, url, parameters, headers, input, self.__customConnection(url)))
```
mfonville: we remove support for passing a magic-string as `cnx` variable for the outer method.
But for the inner method, all requests can now pass a custom connection as a parameter;

the method `__customConnection()` will parse the value of `url` and check if the request is meant for `https://api.github.com` or respective enterprise URL, which would mean it can re-use the existing API-connection and return `None`;
but if the URL is to another hostname or port than the existing connection, it will initiate new custom connection to the host and pass it as an argument for further processing.

github/Requester.py
```diff
@@ -254,23 +254,34 @@ def __init__(self, login_or_token, password, base_url, timeout, client_id, clien
...
 
     def __check(self, status, responseHeaders, output):
         output = self.__structuredFromJson(output)
         if status >= 400:
             raise self.__createException(status, responseHeaders, output)
         return responseHeaders, output
 
+    def __customConnection(self, url):
+        cnx = None
+        if not url.startswith("/"):
```
mfonville: if it is an absolute URI

github/Requester.py
```diff
@@ -254,23 +254,34 @@ def __init__(self, login_or_token, password, base_url, timeout, client_id, clien
...
+            o = urlparse.urlparse(url)
+            if o.hostname != self.__hostname or \
+               (o.port and o.port != self.__port) or \
+               (o.scheme != self.__scheme and not (o.scheme == "https" and self.__scheme == "http")):  # issue80
```
mfonville: if either the port, the hostname or the http-scheme is different; but ignoring silent upgrades from http to https (that is issue 80)

github/Requester.py
```diff
@@ -303,7 +314,7 @@ def encode(input):
 
         return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
 
-    def requestMultipart(self, verb, url, parameters=None, headers=None, input=None):
+    def requestMultipart(self, verb, url, parameters=None, headers=None, input=None, cnx=None):
```
mfonville: allow passing a custom connection

github/Requester.py
```diff
@@ -372,9 +383,6 @@ def __requestRaw(self, cnx, verb, url, requestHeaders, input):
         original_cnx = cnx
         if cnx is None:
             cnx = self.__createConnection()
-        else:
```
mfonville: We don't have to check for the magic string value anymore, because we already get the proper custom connection to `status.github.com` passed if we are connecting to status API

github/Requester.py
```diff
@@ -413,8 +421,8 @@ def __makeAbsoluteUrl(self, url):
             url = self.__prefix + url
         else:
             o = urlparse.urlparse(url)
-            assert o.hostname in [self.__hostname, "uploads.github.com"], o.hostname
-            assert o.path.startswith((self.__prefix, "/api/uploads"))
+            assert o.hostname in [self.__hostname, "uploads.github.com", "status.github.com"], o.hostname
```
mfonville: assert that the hostname can either be the existing API hostname, or to uploads or status API.
I'd prefer to not have this as hardcoded as it is here, but I don't want to overhaul even more; and the assertion at least helps in spotting coding errors.

github/Requester.py
```diff
@@ -413,8 +421,8 @@ def __makeAbsoluteUrl(self, url):
...
+            assert o.path.startswith((self.__prefix, "/api/"))
```
mfonville: We change this to `/api/` so that the path can match the patterns of both the upload and status API

sfdye:
Thanks for all the comments. Preparing PyCon workshop this weekend, I will try to get this reviewed the hopefully merged soon by next week.

nurupo:
>fyi, I am now using this code ( including #787 ) in production to upload more than 200 assets per day to Github

Uh, sorry for off-topic, but I just saw your project pushing automatic daily releases (daily Travis Cronjob?) that [total around 10-20GiB each](https://github.com/opengapps/arm64/releases/tag/20180602), moreover you have around 800 of these releases across 4 repositories, that's like 8TiB of build artifacts GitHub is storing, that's crazy! I'm surprised GitHub allows you to do this. Have you considered keeping only N latest releases (where N is <10), instead of keeping 200 latest daily releases dating back to September 2017?

mfonville:
@nurupo pssst don't tell them ;-) 
But I am indeed GitHub very thankful for hosting all those releases; if it ain't necessary I prefer to keep the old release(assets) around, for those that have older devices on which even the newest Google Play Services doesn't fit on their `/system` partition.

I do the builds and uploads daily on a dedicated machine. The uploading used to be via a shellscript + GotHub(aka Github-Release) but I switched to PyGithub recently, because of issues with GitHub's servers and anti-DDoS measures. PyGithub proved to be a much more suitable and reliable solution :-)

sfdye:
@mfonville Merged, finally. Thanks for the awesome work!

nurupo:
Great, good work everyone!

mfonville:
Thanks!

PR #776, opened 2018-04-30
Refactor Repository.get_protected_branch

s-t-e-v-e-n-k:
The URL mentioned in the docstring for Repository.get_protected_branch
no longer seems to mention useful content, so to avoid an API break,
just return Repository.get_branch.

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -48,6 +48,7 @@ import time
 import sys
 from httplib import HTTPSConnection
 import jwt
+import warnings
 
 from Requester import Requester, json
 import AuthenticatedUser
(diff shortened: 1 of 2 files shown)


mfonville:
Maybe in the API docs it could be mentioned that using `get_branch()` is recommended and that this method is deprecated?

s-t-e-v-e-n-k:
Actually, I think I removed the custom header from get_protected_branch in a prior PR, which means the methods were not identical, and now they are -- but they would have returned identical data since September last year.

@sfdye I'm happy adding a deprecation notice (and raising a warning if you wish); or we can just deal with the API break and remove the method.

sfdye:
I would issue an depreciation warning first.

PR #782, opened 2018-05-05
Use TestCase.assertRaises()

s-t-e-v-e-n-k:
Now that we no longer support Python 2.6, we can use the assertRaises()
method of TestCase as a context manager, tiding up a large amount of
technical debt.


s-t-e-v-e-n-k:
@sfdye Want to get this non-release note test framework clean up in before a release too? :-)

PR #788, opened 2018-05-10
Include target_commitish in GitRelease

mfonville:


sfdye:
Just need one simple unit test before I merge 😄

mfonville:
It already has its unit test included (!)

PR #791, opened 2018-05-14
Handle HTTP 202 HEAD&GET with a retry

### LINKED ISSUE #564: No HTTP status checks in statistics
For statistics such as `contributors` or `commit_activity`, there are no checks whether the API responded with 200. In a lot of cases, GitHub API responds with 202 as some statistics are computed asynchronously on demand [1]. Even if PyGithub does not perform waits, I would expect notification that I should retry after some time.

[1] https://developer.github.com/v3/repos/statistics/#a-word-about-caching

mfonville:
Fixes #564
After some testing, it seems GitHub seems to be able to process the chrunching of statistics
for every repository I tested in around 2 seconds. So a delay of around 5 seconds should be
plenty to be very certain that the next try should succeed.
Another option would be to put this value lower and possibly waste an extra retry on it.


mfonville:
I will need some help to get new ReplayData for the tests, that should now handle `HTTP 202` responses and re-request the url.

mfonville:
fyi, I limited this now to `HEAD`&`GET` only, to avoid e.g. the situation where GitHub makes new forks using a `POST` and also returns a `202` while it is busy with the process of forking. So only safe operations, that we can rely on that they will return a 200 with a proper result set, will be affected by this change.

mfonville:
Ok, I pushed now a value of 2 seconds

Issue #693, opened 2018-01-31
Large asset upload timed out.


tokejepsen:
Trying to upload an asset of ~600 MB to a GitHub release, but getting the below error.

```python
Traceback (most recent call last):
  File "run_deploy.py", line 47, in <module>
    label=release_title
  File "C:\Python27-x64\lib\site-packages\github\GitRelease.py", line 161, in upload_asset
    input=path
  File "C:\Python27-x64\lib\site-packages\github\Requester.py", line 177, in requestBlobAndCheck
    return self.__check(*self.requestBlob(verb, url, parameters, headers, input))
  File "C:\Python27-x64\lib\site-packages\github\Requester.py", line 245, in requestBlob
    return self.__requestEncode(None, verb, url, parameters, headers, input, encode)
  File "C:\Python27-x64\lib\site-packages\github\Requester.py", line 268, in __requestEncode
    status, responseHeaders, output = self.__requestRaw(cnx, verb, url, requestHeaders, encoded_input)
  File "C:\Python27-x64\lib\site-packages\github\Requester.py", line 295, in __requestRaw
    response = cnx.getresponse()
  File "C:\Python27-x64\lib\httplib.py", line 1121, in getresponse
    response.begin()
  File "C:\Python27-x64\lib\httplib.py", line 438, in begin
    version, status, reason = self._read_status()
  File "C:\Python27-x64\lib\httplib.py", line 394, in _read_status
    line = self.fp.readline(_MAXLINE + 1)
  File "C:\Python27-x64\lib\socket.py", line 480, in readline
    data = self._sock.recv(self._rbufsize)
  File "C:\Python27-x64\lib\ssl.py", line 766, in recv
    return self.read(buflen)
  File "C:\Python27-x64\lib\ssl.py", line 653, in read
    v = self._sslobj.read(len)
ssl.SSLError: ('The read operation timed out',)
```

All I'm doing is ```release.upload_asset("/path/to/deployment.zip")```.

Am I using the library wrong somehow?

ncapalbo:
Within the last week, I've been encountering the same SSL error (i.e. 'The read operation timed out') with frequency I've never experienced before.  The API call that seems to be causing it is get_public_events() of the NamedUser object.  I've tried a bunch of things to reproduce the error but I can't pinpoint the root cause.

sfdye:
Is this still an issue? 

@ncapalbo can you provide some details of your case?

ncapalbo:
[...]

PR #793, opened 2018-05-17
increase default timeout to be > github request timeout

jhoblitt:
Github imposes a 10s limit for completion of API calls and returns an HTTP 502
if when the max time is exceeded.  However, the current default 10s request
timeout in pygithub will cause the a socket timeout before an HTTP status is
received, probably due to network latency and/or latency within the github
front-end.

#693

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -72,7 +72,7 @@ import Invitation
 atLeastPython3 = sys.hexversion >= 0x03000000
 
 DEFAULT_BASE_URL = "https://api.github.com"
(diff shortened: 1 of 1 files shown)


sfdye:
Thanks for the PR. Is this in any of the official docs or only from your communication with Github support?

jhoblitt:
@sfdye AFAIK, only the email I have from support.  I did write back with a suggestion that this be added to the official docs.

sfdye:
Okay merged. Let's see how it's doing in the next alpha release.

PR #794, opened 2018-05-17
add Github.runtime_rack attribute

jhoblitt:


jhoblitt:
I'm having trouble understanding how the test fixture data is handled and need some help... (as an aside, the scripts in the contrib instructions seem to be broken or simply don't work with py3)

I've added `Github.testRuntimeRack.txt` as a copy of `Github.testGetOrganizations.txt`, and `get_organizations()` is returning data from inside the test case, but it doesn't look like` __requester` is seeing the header?  This does work "live".

sfdye:
See #795

sfdye:
Give it another shot please, it should be fixed in master.

Issue #256, opened 2014-07-10
PullRequest.mergeable spuriously returns None on first run


bilderbuchi:
I have a script which fetches pull requests, iterates through them and prints the value of `p.mergeable`. 
Curiously, when I run this script for the first time, the value is often `None`:

```
nr 3078, mergeable:True
nr 3076, mergeable:True
nr 3058, mergeable:True
nr 3046, mergeable:None
nr 3045, mergeable:None
nr 3042, mergeable:None
nr 3004, mergeable:None
nr 3001, mergeable:None
nr 2990, mergeable:True
nr 2974, mergeable:None
```

but if I run this script a second time, the correct values appear:

```
nr 3078, mergeable:True
nr 3076, mergeable:True
nr 3058, mergeable:True
nr 3046, mergeable:True
nr 3045, mergeable:True
nr 3042, mergeable:False
nr 3004, mergeable:False
nr 3001, mergeable:False
nr 2990, mergeable:True
nr 2974, mergeable:True
```

What could be the reason behind this, and how can I work around it in the meantime?

This is using PyGithub 1.25.0 under Python 3 (but I have also previously observed this under Python2)

PR #801, opened 2018-05-22
Add tests for mergeable state

MarcoFalke:
As outlined in https://developer.github.com/v3/pulls/#response-1 and https://github.com/PyGithub/PyGithub/issues/256#issuecomment-51213771 `mergeable` can be `"null"`. I couldn't find tests that cover that behavior, so I added them in this patch.


MarcoFalke:
@TravisBuddy They are green for me

sfdye:
Thanks! @MarcoFalke

PR #803, opened 2018-05-24
Retry50x

mfonville:
The Github API (without any proxy or such) returns in some occasions a `502` error, even on a trivial operations. But it would be ridiculous to try to catch those for literally every Github API operation in programs using PyGithub. So this feature allows to specify a number of retries.

It based on the already existing PR #771 because it touches the same code and I don't want to have any regressions.

### DIFF
--- a/github/Consts.py
+++ b/github/Consts.py
@@ -31,6 +31,7 @@
 
 REQ_IF_NONE_MATCH = "If-None-Match"
 REQ_IF_MODIFIED_SINCE = "If-Modified-Since"
+HTTPSTATUS_50X_WAIT_TIME = "10"
 
 # ##############################################################################
 # Response Header                                                              #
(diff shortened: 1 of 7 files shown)


mfonville:
It still needs tests, and also not sure whether we might want to reduce it to only covering the `502`. Because with `502` I am 100% certain that the server has not processed the request at all, while I do not know this for sure for the `504`

mfonville:
I have done some more testing with this code in production, and apparently with `502` errors we can still not be sure that there was some kind of (partial) processing. This is no problem for `HEAD` and `GET` statements (those are not allowed to change anything server side) but can be problematic for `POST` and `PUT`, e.g. when uploading release assets, this can mean that there are partial uploads.

When refactoring this, I will have to check that automatic retries are only allowed for these just like in #791 ; and we'd probably need a suitable exception to be thrown for the unrecoverable errors, when the state of the processing on Github's server is unknown.

mfonville:
I will close this PR, I think I will take another shot at a later time.

After looking how other projects this did, the bes solution would be to allow the passing of Requests-parameters or object when making the PyGitHub instance.

PR #820, opened 2018-06-14
Calcuate totalCount without iterating PaginatedList

koglinjg:
This PR #596 seems stale, so I have opened up this one with the requested changes from @sfdye.

This is the code from @Tommos0's [comment][1] with some modification.

There are 2 tests failing. I think this is due to the change in the query string and the cached test data just needs to be refreshed for those tests.

[1]: https://github.com/PyGithub/PyGithub/pull/596#issuecomment-315013949


sfdye:
It looks like this change is affecting the testing framework.

There are three places in the tests where `totalCount` is used. All failed due to this line:

```
self.__testCase.assertEqual(self.__splitUrl(url), self.__splitUrl(readLine(self.__file)))
```

github/PaginatedList.py
```diff
@@ -134,8 +134,20 @@ def __init__(self, contentClass, requester, firstUrl, firstParams, headers=None,
...
+                self.__firstUrl,
+                parameters=params,
+                headers=self.__headers
+            )
+            if 'link' not in headers:
+                self.__totalCount = len(data)
+            else:
+                links = self.__parseLinkHeader(headers)
+                lastUrl = links.get("last")
+                self.__totalCount = int(re.match(r'.*[&\?]page=(\d*).*?', lastUrl).group(1))
```
mfonville: Isn't it better to use `urlparse` to find the `query` part of the lastURL and then using `parse_qs(l)` to find the value of `page`, than using your own regex?

sfdye: I agree, this is better

sfdye:
@koglinjg If you don't mind, I have helped you to fix the tests and also changed the implementation to use `urllib` to get the page number.

@mfonville what do you think?

sfdye:
Also what about implementing the `__len__` magic method for `PaginatedList` to just return `self.totalCount`, because I feel some people might not be aware of the `totalCount` property but try to call `len()` on the list, which now will throw a AttributeError: PaginatedList instance has no attribute '__len__'.

mfonville:
@sfdye lgtm
Also the len() is I think a good idea.

sfdye:
Thank you @koglinjg for this PR and @Tommos0 for the original idea in this [comment](https://github.com/PyGithub/PyGithub/issues/415#issuecomment-381050047). Kudos! We now have a much more efficient way to determine the list length, without actually going through all the items.

cmackenzie1:
Any update on adding `__len__`?

> Also what about implementing the `__len__` magic method for `PaginatedList` to just return `self.totalCount`, because I feel some people might not be aware of the `totalCount` property but try to call `len()` on the list, which now will throw a AttributeError: PaginatedList instance has no attribute '**len**'.

Issue #819, opened 2018-06-13
Delete release operation missing


jglapa:
Could this be added?

Side note:
It might seem odd to remove a release but I'm backfiling releases in an old project and realized that I did it incorrectly and now have to cleanup.

mfonville:
I think this should work, did you check if you can call `delete_release()` on a release?
https://github.com/PyGithub/PyGithub/blob/master/github/GitRelease.py#L168

jglapa:
true, I couldn't find that method in the reference : https://pygithub.readthedocs.io/en/latest/github_objects/GitRelease.html#github.GitRelease.GitRelease

mfonville:
I think the reference misses a lot of about the functions of GitRelease; if you or anyone would have time to update those docs, that'd be great

sfdye:
Indeed, a lot of documentation are missing.

hemanthsavasere:
Can anyone provide me the basic usage example of this library as I am a beginner?

sfdye:
@hemanthsavasere 
https://pygithub.readthedocs.io/en/latest/examples.html

will get you started.

hemanthsavasere:
Hi @sfdye,
I looked into examples, I want to delete the release in Github after executing the command g.create_git_release(). I don't know how to proceed further.

sfdye:
First get that release and then https://pygithub.readthedocs.io/en/latest/github_objects/GitRelease.html#github.GitRelease.GitRelease.delete_release

hemanthsavasere:
Thanks, Can I add this usage example to the documentation part, as it was confusing for a beginner on how to use it.

sfdye:
Sure, feel free to make a PR. Any improvements to the docs is welcomed :)

PR #821, opened 2018-06-19
GitRelease methods documentation

jglapa:
Hi, here's my small contribution to the documentation in relation to #819 

Few things:
* the `delete_release` method always returns `True` so not sure if the return type is truly valid - potentially a bug
* I've tried generating the documentation locally to check if formatting is ok but failed - is there a guide to it?   I followed the read the docs starter guide but that didn't seem to reflect the way it's done here

### DIFF
--- a/github/GitRelease.py
+++ b/github/GitRelease.py
@@ -170,7 +170,7 @@ class GitRelease(github.GithubObject.CompletableGithubObject):
         :calls: `DELETE /repos/:owner/:repo/releases/:release_id <https://developer.github.com/v3/repos/releases/#delete-a-release>`_
         :rtype: None
         """
-        self._requester.requestJsonAndCheck(
+        headers, data = self._requester.requestJsonAndCheck(
             "DELETE",
             self.url
         )
--- a/github/GitRelease.py
+++ b/github/GitRelease.py
@@ -166,6 +166,10 @@ class GitRelease(github.GithubObject.CompletableGithubObject):
         return self._zipball_url.value
 
     def delete_release(self):
+        """
+        :calls: `DELETE /repos/:owner/:repo/releases/:release_id <https://developer.github.com/v3/repos/releases/#delete-a-release>`_
+        :rtype: bool
+        """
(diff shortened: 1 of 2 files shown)


github/GitRelease.py
```diff
@@ -166,13 +166,21 @@ def zipball_url(self):
         return self._zipball_url.value
 
     def delete_release(self):
+        """
+        :calls: `DELETE /repos/:owner/:repo/releases/:release_id <https://developer.github.com/v3/repos/releases/#delete-a-release>`_
+        :rtype: bool
```
sfdye: Actually we don't need to return anything here. You can remove the return and add a `:rtype: None`

jglapa: done here: 0b2a6c30cd34356b3

jglapa: so there is a test expecting this method to return `True`. Shall I remove it?

sfdye: Of course, otherwise the tests will fail.

jglapa: ok, it was more a question if the assertion or the whole test case should be removed. I did the former now: 438b10c5935550f3

github/GitRelease.py
```diff
@@ -166,13 +166,20 @@ def zipball_url(self):
         return self._zipball_url.value
 
     def delete_release(self):
-        headers, data = self._requester.requestJsonAndCheck(
+        """
+        :calls: `DELETE /repos/:owner/:repo/releases/:release_id <https://developer.github.com/v3/repos/releases/#delete-a-release>`_
+        :rtype: None
+        """
+        self._requester.requestJsonAndCheck(
```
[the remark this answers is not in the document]
jglapa: they are unused, what's the point?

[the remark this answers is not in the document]
sfdye: Simply for consistency. If we are gonna remove them, we will do it at one go.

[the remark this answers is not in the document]
jglapa: revert here 24654976

sfdye:
@jglapa Thank you very much!

PR #809, opened 2018-05-31
add Dismiss PR Reviews functionality

cjkoral:
This PR will allow end users to dismiss pull request reviews. This is helpful in conjunction with Github webhooks to dismiss stale reviews when new commits are pushed.

This was accomplished by a 'PUT' request very similar to the functionality of the `pull_request.get_review(<id>)` that contains a body in the form of `{"message": "<your message>}` and returns the dismissed review.

PR #827, opened 2018-06-25
Unit test 809

eamanu:
Add testcase for PR #809


eamanu:
Hello, 

I am working to add test case for the PR #809. But if I run the python -m _testcase_ --record this command makes that I login to my repo (eamanu/PyGithub) and I don't have any PR (for example the 31 that is used on the tests) so fail.

What can I do to create testcases?

Thanks!

github/PullRequest.py
```diff
@@ -525,7 +525,7 @@ def get_review_comment(self, id):
             "GET",
             self._parentUrl(self.url) + "/comments/" + str(id)
         )
-        return github.PullRequestComment.PullRequestComment(self._requester, headers, data, completed=True)
+        return github.PullRequestComment.PullRequestComment(self._requester, headers, data, completed=True)
```
wyardley: Looks like you're adding some trailing whitespace here, FWIW

wyardley:
@eamanu I think you'll need to add fake data to https://github.com/PyGithub/PyGithub/tree/master/github/tests/ReplayData

github/PullRequest.py
eamanu: Ok, I will fix in the next PR. Thanks

eamanu:
@wyardley  Sorry, I don't understand you. How I add fake data?

PR #635, opened 2017-09-21
Add support for getting and replacing topics on a repository

peterkline:
Fixes #634

I added the committed code and tested it by setting up a new repository and using an existing one
The get was tested by creating a github login and setting up a repository object.  I then used the get_topics function to return a list of existing topics I had created manually.  I did get back the topics that were already added.  I also used get_topics on a repository that had no topics created and it does return an empty list.

Testing the replace_topics I simply created a list with a few entries and also an empty list.  Both performed as expected.

Issue #634, opened 2017-09-21
Add support for getting and replacing topics on a repository


peterkline:
Topics are a fairly new feature within Github.  Currently this functionality is provided in the API in a beta mode.  We have a need to use these to add attributes and then consume them for automation.  Hopefully this helps others.

API Documentation:
[List topics in a repo](https://developer.github.com/v3/repos/#list-all-topics-for-a-repository)
[Replace topics in a repo](https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository)

Examples of use:
```python
# Get list of topics from a repository    
topics = repo_obj.get_topics()

# Replace current (if there are any) topics with list
repo_obj.replace_topics(['base', 'personal', 'playground'])
```

TheVetter:
Has topics support been added?

sfdye:
@TheVetter Should be soon, I have reviewed the PR and still waiting response for the original author @peterkline

migueleliasweb:
While this issue is not fixed. we should be able to access the topics with the `_rawData`, right ?

sfdye:
@migueleliasweb I am afraid not, as it requires calling another endpoint. (`/topics`) However, I will try to get this merged soon.

sfdye:
@migueleliasweb You are right, the `topics` attributes is currently in preview, but to edit the attributes we still need to go back to `/topics` endpoint.

migueleliasweb:
@sfdye I just wanted to list them =/

Anyway... I seem unable to get them even accessing the `_rawData` directly. See:

```python
ipdb> ll
    643     def get_repo(self, name):
    644         """
    645         :calls: `GET /repos/:owner/:repo <http://developer.github.com/v3/repos>`_
    646         :param name: string
    647         :rtype: :class:`github.Repository.Repository`
    648         """
    649         assert isinstance(name, str), name
    650         headers, data = self._requester.requestJsonAndCheck(
    651             "GET",
    652             "/repos/" + self.login + "/" + name
    653         )
--> 654         return github.Repository.Repository(self._requester, headers, data, completed=True)
    655 

ipdb> data["topics"]
*** KeyError: 'topics'
```

sfdye:
@migueleliasweb You're missing the custom header `application/vnd.github.mercy-preview+json`

migueleliasweb:
@sfdye Pardon me, I totally missed that !

edanhauon:
Hi @sfdye 
How soon is soon?

[...]

PR #832, opened 2018-07-05
Add support for getting / replacing topics (#634)

wyardley:
This is mostly @peterkline's work, just trying to replace #635 (which hasn't been updated) and adjust for some of the changes requested / discussed there.

This seems to resolves the existing test failure, and I think resolves most of the other feedback in the other PR. It also switches the input to `replace_topics()` to the expected array as discussed. Would appreciate feedback, esp. on the test case I added. Because there's no expected return from the method, I think that test is acceptable as-is? I'm not that familiar with this module's test setup, though, so let me know if I should update anything.

I didn't have much luck rebasing, so I just patched, but I did retain original author on the first commit.
cc @sfdye

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -673,7 +673,7 @@ class Repository(github.GithubObject.CompletableGithubObject):
     @property
     def topics(self):
         """
-        :type: list 
+        :type: list of strings
         """
         self._completeIfNotSet(self._topics)
         return self._topics.value
@@ -2314,7 +2314,7 @@ class Repository(github.GithubObject.CompletableGithubObject):
 
     def get_topics(self):
         """
-        :calls: `GET /repos/:owner/:repo/topics <http://developer.github.com/v3/repos>`_
+        :calls: `GET /repos/:owner/:repo/topics <https://developer.github.com/v3/repos/#list-all-topics-for-a-repository>`_
         :rtype: list of strings
         """
         headers, data = self._requester.requestJsonAndCheck(
@@ -2452,11 +2452,14 @@ class Repository(github.GithubObject.CompletableGithubObject):
         :param topics: list of strings
         :rtype: None
         """
+        post_parameters = {
+            'names': topics
+        }
         headers, data = self._requester.requestJsonAndCheck(
             "PUT",
             self.url + "/topics",
             headers={'Accept': 'application/vnd.github.mercy-preview+json'},
-            input=topics
+            input=post_parameters
         )
 
     def remove_from_collaborators(self, collaborator):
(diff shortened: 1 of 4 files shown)


github/tests/Repository.py
```diff
@@ -711,6 +712,13 @@ def testStatistics(self):
     def testGetLicense(self):
         self.assertEqual(len(self.repo.get_license().content), 47646)
 
+    def testGetTopics(self):
+        topic_list = self.repo.get_topics()
+        topic = u'github'
+        self.assertEquals(topic in topic_list, True)
+
+    def testReplaceTopics(self):
+        self.repo.replace_topics(['github', 'testing'])
```
wyardley: Looking at some of the other tests, I think no assertion is needed here? Let me know if this needs to be improved.

github/Repository.py
```diff
@@ -2426,6 +2446,22 @@ def protect_branch(self, branch, enabled, enforcement_level=github.GithubObject.
             input=post_parameters
         )
 
+    def replace_topics(self, topics):
+        """
+        :calls: `PUT /repos/:owner/:repo/topics <http://developer.github.com/v3/repos>`_
```
sfdye: https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository

github/Repository.py
```diff
@@ -2304,6 +2312,18 @@ def get_teams(self):
             None
         )
 
+    def get_topics(self):
+        """
+        :calls: `GET /repos/:owner/:repo/topics <http://developer.github.com/v3/repos>`_
```
sfdye: Let's add the exact link:
https://developer.github.com/v3/repos/#list-all-topics-for-a-repository

github/tests/Repository.py
```diff
@@ -711,6 +712,13 @@ def testStatistics(self):
     def testGetLicense(self):
         self.assertEqual(len(self.repo.get_license().content), 47646)
 
+    def testGetTopics(self):
+        topic_list = self.repo.get_topics()
+        topic = u'github'
+        self.assertEquals(topic in topic_list, True)
```
sfdye: `self.assertIn`?

github/tests/Repository.py
sfdye: It's fine

github/Repository.py
```diff
@@ -670,6 +670,14 @@ def teams_url(self):
         self._completeIfNotSet(self._teams_url)
         return self._teams_url.value
 
+    @property
+    def topics(self):
+        """
+        :type: list
```
sfdye: list of string

wyardley:
@sfdye Thanks for the review feedback, I think I addressed your points -- let me know how that looks.

sfdye:
@wyardley Thank you very much for putting this together. Merged!

PR #834, opened 2018-07-07
Add tag_name and target_commitish arguments to GitRelease.update_release

### LINKED ISSUE #833: Allow updating tag_name and target_commitish of GitRelease
GitHub API [allows updating](https://developer.github.com/v3/repos/releases/#edit-a-release) `tag_name` and `target_commitish` of a release, among `name`, `body`, `draft` and `prerelease`, but PyGithub API wrapper allows updating all of them but  `tag_name` and `target_commitish`:

https://github.com/PyGithub/PyGithub/blob/e389396405146bb1b0c4587de1f6f757a9514a43/github/GitRelease.py#L178-L199

Could we add `tag_name` and `target_commitish` parameters to that function?

nurupo:
Fixes #833

### DIFF
--- a/github/GitRelease.py
+++ b/github/GitRelease.py
@@ -175,22 +175,30 @@ class GitRelease(github.GithubObject.CompletableGithubObject):
             self.url
         )
 
-    def update_release(self, name, message, draft=False, prerelease=False):
+    def update_release(self, name, message, draft=False, prerelease=False, tag_name=None, target_commitish=None):
         """
         :calls: `PATCH /repos/:owner/:repo/releases/:release_id <https://developer.github.com/v3/repos/releases/#edit-a-release>`_
         :rtype: :class:`github.GitRelease.GitRelease`
         """
+        if not tag_name:
+            tag_name = self.tag_name
         assert isinstance(name, (str, unicode)), name
         assert isinstance(message, (str, unicode)), message
         assert isinstance(draft, bool), draft
         assert isinstance(prerelease, bool), prerelease
+        assert isinstance(tag_name, (str, unicode)), tag_name
         post_parameters = {
-            "tag_name": self.tag_name,
+            "tag_name": tag_name,
             "name": name,
             "body": message,
             "draft": draft,
             "prerelease": prerelease,
         }
+        # Do not set target_commitish to self.target_commitish when ommited, just don't send it
+        # alltogether in that case, in order to match the Github API behaviour. Only send it when set.
+        if target_commitish:
+            assert isinstance(target_commitish, (str, unicode)), target_commitish
+            post_parameters['target_commitish'] = target_commitish
         headers, data = self._requester.requestJsonAndCheck(
             "PATCH",
             self.url,
--- a/github/GitRelease.py
+++ b/github/GitRelease.py
@@ -175,18 +175,25 @@ class GitRelease(github.GithubObject.CompletableGithubObject):
             self.url
         )
 
-    def update_release(self, name, message, draft=False, prerelease=False, tag_name=None, target_commitish=None):
+    def update_release(self, name, message, draft=False, prerelease=False,
+                       tag_name=github.GithubObject.NotSet,
+                       target_commitish=github.GithubObject.NotSet):
         """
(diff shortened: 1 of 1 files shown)


github/GitRelease.py
```diff
@@ -175,17 +175,24 @@ def delete_release(self):
...
-    def update_release(self, name, message, draft=False, prerelease=False):
+    def update_release(self, name, message, draft=False, prerelease=False, tag_name=None, target_commitish=None):
         """
         :calls: `PATCH /repos/:owner/:repo/releases/:release_id <https://developer.github.com/v3/repos/releases/#edit-a-release>`_
         :rtype: :class:`github.GitRelease.GitRelease`
         """
+        if not tag_name:
+            tag_name = self.tag_name
+        if not target_commitish:
+            target_commitish = self.target_commitish
```
mfonville: I think we should not set `target_commitish` value from `self`, because of this possible scenario:

- You already have an existing release, with an existing `tag_name` and a `target_commitish` at a specific SHA-hash.
- If then this release is edited, and a new `tag_name` that does not exist yet is set and no `target_commitish` is supplied, the behaviour of Github (API) would be to create this new tag with `target_commitish = master`.
- But if we use the value of `self.target_commitish` the result would be that this new tag would be set with `target_commitish = SHA-value`

The right approach would be that if `target_commitish` is not supplied, to not include it in the `post_parameters` at all.

mfonville:
Great that you made a PR for this! Since I added `target_commitish` to the Release-API, I am happy to see improvements :-)
I have one comment, and maybe the Edit-Release needs in general a bit of review, because the API documentation is not clear if/which input parameters are obligatory at all.

nurupo:
@mfonville does this look better? If so, I will squash the two commits.

github/GitRelease.py
mfonville: The inconsistent naming was also something I noticed, but I assume the maintainer knows why, and whether a refactor would be appropriate.

About editing a release without inputting any `target_commitish`, If you specified its value earlier in the draft-phase, and as long as you don't change the tag_name, the target_commitish will stay the same. I use this in production and it always works like a charm :-)

mfonville:
LGTM now, it is up to @sfdye to review for merging

nurupo:
Alright, squashed.

github/GitRelease.py
nurupo: @mfonville If you don't specify commitish when updating a release, and PyGithub doesn't specify it, it sounds like GitHub API will reset commitish to the default branch, at least going off their API documentation, I have not tested it myself. Not sure why it works for you.

mfonville: I use it in production. when creating the draft release I specify the tag_name and the sha-commit to attach it to with target_commitish. When later editing the release I am not changing the tag, so in effect only changing the draft-status. That results in that it retains the target_commitish

sfdye:
@nurupo This looks good to me. Just one minor thing though, looking at all the other edit/update methods, it seems they are using `github.GithubObject.NotSet` as the the default value. I am not sure either what the intention of the original author is but just trying to be consistent..

nurupo:
Hm, the code base is inconsistent then, as some functions don't use `github.GithubObject.NotSet` for their default values, e.g.:

```bash
$ grep 'def .*update' * --colour
GitReleaseAsset.py:    def update_asset(self, name, label=""):
GitRelease.py:    def update_release(self, name, message, draft=False, prerelease=False):
```

nurupo:
I assume you want me to follow this example then https://github.com/PyGithub/PyGithub/blob/c6802b515ffb64634a99ceb8e493d74a83deed87/github/Repository.py#L1457-L1460?

sfdye:
@nurupo Yes you are right, the code base is not consistent unfortunately. Appreciate if we can at least make this one right.

nurupo:
How does this look?

sfdye:
@nurupo Perfect, thanks!

nurupo:
Uh... why did you merge it?

nurupo:
I didn't squash the commits together yet.

sfdye:
No worries.

________________________________

PR #837, opened 2018-07-09
github/stale: use a stale label

jzelinskie:
Previously it was defaulting to `wontfix` which may not be the case if an issue just goes idle.


sfdye:
@jzelinskie Thanks for this, the default label is kinda misleading. I have fixed the tests in master so if you could rebase I will merge!

jzelinskie:
All good! Thanks again for turning on this integration. It's awesome.

PR #753, opened 2018-04-16
Team description added to Team.py 

manuhortet:
Adressing #450

PR #839, opened 2018-07-10
Allow editing of Team descriptions

jacquerie:
PR https://github.com/PyGithub/PyGithub/pull/753 added this property but forgot to make it editable. The current PR fixes this omission.


jacquerie:
Great, thanks again!

PR #842, opened 2018-07-11
Add missing properties to Repository

jacquerie:
Closes #784
Closes #785

First commit fixes a small bug that I observed while trying to record some new test fixtures: the stringification of the headers is not correctly printed as a list of tuples, but as an `iteritems` object, so I forced the list evaluation.

Second commit adds the missing properties `allow_merge_commit`, `allow_rebase_merge`, `allow_squash_merge`, and `has_projects` to the `Repository` class.

Note however that I ran into some trouble while modifying the `edit` method of the `Repository` class to make use of the new properties, specifically while writing tests, so I will fix the `edit` method in a future PR.

PR #844, opened 2018-07-14
Add missing arguments to Repository.edit

jacquerie:
As promised in https://github.com/PyGithub/PyGithub/pull/842, here's a PR to update the `edit` method so that it allows editing the newly added properties.


github/tests/ReplayData/Repository.testEditWithAllArguments.txt
```diff
@@ -4,10 +4,10 @@ api.github.com
 None
 /repos/jacquev6/PyGithub
 {'Content-Type': 'application/json', 'Authorization': 'Basic login_and_password_removed', 'User-Agent': 'PyGithub/Python'}
-{"has_wiki": false, "name": "PyGithub", "has_downloads": true, "has_issues": true, "homepage": "http://vincent-jacques.net/PyGithub", "private": true, "description": "Description edited by PyGithub"}
+{"has_wiki": false, "name": "PyGithub", "has_downloads": true, "has_issues": true, "homepage": "http://vincent-jacques.net/PyGithub", "private": true, "description": "Description edited by PyGithub", "has_projects": false, "allow_squash_merge": true, "allow_merge_commit": true, "allow_rebase_merge": true}
```
jacquerie: Manually editing the request/response pair is error prone and ugly, but it is necessitated by the fact that `self.repo` in the test points to a repository nobody else has the correct rights to, so whenever one needs to modify one of the existing tests there's no way to just re-record the exchange.

PR #849, opened 2018-07-20
Remove bulky test data from distribution

### LINKED ISSUE #672: Don't include tests in distribution?
One task I'm working on is sliming down a docker image and one thing that came up was `pygithub`.

The library itself is tiny, but we take a *huge* hit from the tests

```console
# du -hs lib/python3.6/site-packages/github/
21M	lib/python3.6/site-packages/github/
# du -hs lib/python3.6/site-packages/github/tests
20M	lib/python3.6/site-packages/github/tests
# rm -rf lib/python3.6/site-packages/github/tests && du -hs lib/python3.6/site-packages/github/
1.3M	lib/python3.6/site-packages/github/
```

Would a PR which doesn't include the tests in the distribution by default be acceptable?

asottile:
Resolves #672

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,9 +82,6 @@ if __name__ == "__main__":
             "github",
             "github.tests",
         ],
-        package_data={
-            "github": ["tests/ReplayData/*.txt"]
-        },
         classifiers=[
             "Development Status :: 5 - Production/Stable",
             "Environment :: Web Environment",


asottile:
oh right, `setup.py test` uses the build directory in python3, I'll need to think more about this then

Issue #606, opened 2017-07-31
Working with projects?


isbm:
Hi,
is there any way to list projects, its columns, add/remove stuff to the columns etc? I kind of can't find that...

sfdye:
Project endpoint is not implemented yet. I am open for any PR.

bbi-yggy:
For reference, these are the API's to be wrapped:
https://developer.github.com/v3/projects
https://developer.github.com/v3/projects/cards

Would also be nice to be able to get the project associated with a pull request, unfortunately that is not (currently) supported by the Github API v3.

bbi-yggy:
I've started looking at this. Should have at least a preliminary pull request in the not too distant future.

bbi-yggy:
Here's a pull request on my own fork to demonstrate what I've done so far. Needs more work (see "To Do" list) but is now functional.
https://github.com/bbi-yggy/PyGithub/pull/1

@sfdye Please let me know how this is looking. Does the overall approach and implementation seem ok? Am I sticking with PyGithub and python conventions to your satisfaction? (I'm a bit of a python noob, so bear with me ...)

bbi-yggy:
BTW, can someone tell me how I can test my in-development version of PyGithub with other scripts I'm working on? Normally I "pip install PyGithub" and then python just knows where to find it.

Should I use "pip install --editable" as described [here](https://pip.pypa.io/en/stable/reference/pip_install/#options)?

sfdye:
@bbi-yggy can you open your PR against PyGithub, not your fork?

bbi-yggy:
@sfdye I planned to wrap up the "to do" items before opening up the PR against the main repo - if you prefer then I'm happy to the put the work-in-progress PR against main (just don't merge it yet!)

bbi-yggy:
Pull request opened as #854.

sfdye:
#854 is merged. For those who wanna try: `pip install pygithub==1.43.1`

scjody:
Remaining work:
* [ ] [Create an organization project](https://developer.github.com/v3/projects/#create-an-organization-project)
* [ ] [Create a user project](https://developer.github.com/v3/projects/#create-a-user-project)
* [ ] [List user projects](https://developer.github.com/v3/projects/#list-user-projects)
* [ ] [Update a project](https://developer.github.com/v3/projects/#update-a-project)
* [ ] [Delete a project](https://developer.github.com/v3/projects/#delete-a-project)
[...]

PR #854, opened 2018-08-02
Add support for projects

bbi-yggy:
Initial solution for [PyGithub support for projects](https://github.com/PyGithub/PyGithub/issues/606) (#606). Adds comprehensive project querying API.

Currently does not support for modifying projects, columns or cards. (I only need this interface for reporting purposes. Of course once I'm done others are more than welcome to add additional features!)

API that was integrated:
https://developer.github.com/v3/projects
https://developer.github.com/v3/projects/columns
https://developer.github.com/v3/projects/cards

Note that the Github project API is in preview mode - it requires a special header in order for requests to be processed, and the API is subject to change without notice.

## Summary
- new classes: Project, ProjectColumn, ProjectCard
- add Organization.get_projects method
- add Repository.get_projects method
- add MainClass.get_project method
- add test cases to exercise all new classes and methods and verify attributes are as expected; replay data is included, recorded (mostly) from my fork of PyGithub
- updated add_attribute script to:
  - use makeXXXAttribute API
  - be able to add to the end of the current set of properties
  - handle both Completable and NonCompletable class types correctly

## Checklist
Not to be merged until these are done:

- [x] remaining Project properties
- [x] remaining ProjectColumn properties
- [x] remaining ProjectCard properties
- [x] support for archived_state parameter when getting cards: all,archived, or not_archived
- [x] get issue from card, not just pull request ("get_content" method?)
- [x] centralize header management for "preview" access to projects API
- [x] add test for retrieving organization projects
- [x] add test for getting issue / pull request content from card


CONTRIBUTING.md
```diff
@@ -51,7 +51,7 @@ oauth_token = "my_token"  # Can be left empty if not used
 ```
 
 If you use 2 factor authentication on your Github account, tests that require a login/password authentication will fail.
-You can use `python -m github.tests Issue139.testCompletion --record --auth_with_token` to use the `oauth_token` field specified in `GithubCredentials.py` when recording a unit test interaction. NB that the `password = ""` (empty string is ok) must still be present in `GithubCredentials.py` to run the tests even when the `--auth_with_token` arg is used.
+You can use `python -m github.tests Issue139.testCompletion --record --auth_with_token` to use the `oauth_token` field specified in `GithubCredentials.py` when recording a unit test interaction. Note that the `password = ""` (empty string is ok) must still be present in `GithubCredentials.py` to run the tests even when the `--auth_with_token` arg is used. Also note that you need to provide the `--auth_with_token` command line argument when running the test, not just when recording.
```
bbi-yggy: I'll also add a note about "pip install --editable (workspace)" for testing PyGithub work in progress with external scripts.

github/Organization.py
```diff
@@ -18,6 +18,7 @@
 # Copyright 2018 Anton Nguyen <afnguyen85@gmail.com>                           #
 # Copyright 2018 sfdye <tsfdye@gmail.com>                                      #
 # Copyright 2018 Jacopo Notarstefano <jacopo.notarstefano@gmail.com>           #
+# Copyright 2018 bbi-yggy <yossarian@blackbirdinteractive.com>                 #
```
bbi-yggy: What's the protocol on copyright notices? I don't mind credit, though tbh I'd rather just include my github username and not my email.

github/Project.py
```diff
@@ -0,0 +1,269 @@
...
+#                                                                              #
+################################################################################
+
+import json
+
+import github.GithubObject
+
+
+# TODO: remaining Project properties
+class Project(github.GithubObject.CompletableGithubObject):
```
bbi-yggy: I wasn't entirely clear on the difference between Completable and NonCompletable objects, and in particular how to decide which to use?

github/Project.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @property
+    def url(self):
+        """
+        :type: string
+        """
+        return self._url.value
+
+    # TODO: get issue from card, not just pull request ("get_content" method?)
+    def get_pullrequest(self):
```
bbi-yggy: Thoughts on this? What do you think of a `get_content( )` method that returns either a PullRequest or an Issue? Note that "note" cards have no content_url, so the only content is the note.

github/tests/Issue606.py
```diff
@@ -0,0 +1,45 @@
...
+
+# TODO: test for retrieving organization projects
+# TODO: test for getting issue / pull request content from card
+
+class Issue606(Framework.TestCase):
+    def setUp(self):
+        Framework.TestCase.setUp(self)
+        self.repo = self.g.get_user().get_repo("PyGithub")
+
+    def testProjectList(self):
```
bbi-yggy: Not really a unit test so much as a light weight exercising of the API.

github/tests/ReplayData/Issue606.testProjectList.txt
```diff
@@ -0,0 +1,55 @@
+https
```
bbi-yggy: Does the ReplayData let you run my tests? Or do you need to record your own data?

github/Project.py
```diff
@@ -0,0 +1,269 @@
...
+# along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
+#                                                                              #
+################################################################################
+
+import json
+
+import github.GithubObject
+
+
+# TODO: remaining Project properties
```
bbi-yggy: I saw there's a script to automate the addition of new properties. I'll be using that.

bbi-yggy:
@sfdye I added some comments to guide your review. Feedback welcome.

bbi-yggy:
The failing build looks to me like the issue I noticed while running tests, that I had to provide the --auth-with-token parameter when running a test for which I had recorded ReplayData using --auth-with-token. I suspect the CI system is not using token-based authentication, and therefore my ReplayData isn't compatible.

bbi-yggy:
@sfdye Please see my comment above on why I think the tests are failing. Looks like a flaw in the test runner to me. I need to use token-based authentication because I use two-factor authentication, so password alone doesn't work. Looks like my recorded ReplayData is thus made incompatible with the test runner.

sfdye:
Good work! For recording the test connection, if you're using two-factor like I do, you can create a personal token to be filled in as the password.

github/Project.py
sfdye: You are right

CONTRIBUTING.md
sfdye: Okay this is a good idea

bbi-yggy:
Yes, I am using two-factor and a personal token - but this method requires that I use the `--auth-with-token` not just when I record the ReplayData, but also when I run the test. Is there something I need to change to allow my recorded data to be used on the build system? The test passes for me locally.

github/Project.py
bbi-yggy: GitHub's new v4 GraphQL API has a ProjectCard class with a `content` property, which I will emulate. However I think it is more clear that this is issuing another API call, so I will call the method `get_content`. (Feedback welcome.)

github/tests/Issue606.py
bbi-yggy: Oh, I guess I can assert the expected results from my recorded data. I'll do that.

bbi-yggy:
@sfdye BTW. regarding your addition that this "Closes" #606 - my intention is to complete the read-only functionality needed for listing projects, columns, and cards. I don't plan to add support at this time for creating, moving, or modifying cards. So I expect to be able to "list projects, its columns" but not yet "add/remove stuff to the columns". But at least it will be a step in the right direction. :-)

bbi-yggy:
Woop woop! Build checks passed! It's a little odd that tests recorded with `--auth_with_token` need to have their code modified to use the replay data, but now I don't need to modify the testing Framework. :-)

Still not ready for merge, but it's a step closer.

CONTRIBUTING.md
```diff
@@ -51,7 +51,13 @@ oauth_token = "my_token"  # Can be left empty if not used
 ```
 
 If you use 2 factor authentication on your Github account, tests that require a login/password authentication will fail.
-You can use `python -m github.tests Issue139.testCompletion --record --auth_with_token` to use the `oauth_token` field specified in `GithubCredentials.py` when recording a unit test interaction. NB that the `password = ""` (empty string is ok) must still be present in `GithubCredentials.py` to run the tests even when the `--auth_with_token` arg is used.
+You can use `python -m github.tests Issue139.testCompletion --record --auth_with_token` to use the `oauth_token` field specified in `GithubCredentials.py` when recording a unit test interaction. Note that the `password = ""` (empty string is ok) must still be present in `GithubCredentials.py` to run the tests even when the `--auth_with_token` arg is used. Note that if you record your test data with `--auth_with_token` then you may need to modify the setup code for your test(s) to authenticate with the token, `self.g = github.Github(self.oauth_token)`.
+
```
bbi-yggy: Update previous line to indicate how a test can authenticate with token (even when `--auth_with_token` not passed on command line). I still wonder if the Framework should be modified to make this unnecessary?

CONTRIBUTING.md
```diff
@@ -51,7 +51,13 @@ oauth_token = "my_token"  # Can be left empty if not used
...
+To run manual tests with external scripts that use the PyGithub package, you can install your development version with:
```
bbi-yggy: New note added here.

github/tests/Issue606.py
```diff
@@ -0,0 +1,51 @@
...
+import Framework
+
+import github
+
+# TODO: test for retrieving organization projects
+# TODO: test for getting issue / pull request content from card
+
+# https://github.com/PyGithub/PyGithub/issues/606
+# https://github.com/PyGithub/PyGithub/pull/854
+class Issue606(Framework.BasicTestCase):
```
bbi-yggy: I changed this from a FrameworkTestCase to a BasicTestCase as I was following another example that uses token-based authentication. What's the difference?

github/tests/Issue606.py
```diff
@@ -0,0 +1,51 @@
...
+    def setUp(self):
+        Framework.BasicTestCase.setUp(self)
+        self.g = github.Github(self.oauth_token)
```
bbi-yggy: This line was the key to being able to record with `--auth_with_token` and replay without.

github/tests/ReplayData/Issue606.testProjectList.txt
bbi-yggy: Understood now - anyone can run my recorded test.

sfdye:
Ping me if this is ready for review please.

bbi-yggy:
@sfdye I will let you know when it's ready. I'm picking away at it after work hours, so progress is a little slow. Probably one more evening should do it, hopefully this weekend.

sfdye:
No worries, really appreciate your contribution.

bbi-yggy:
@sfdye Ready for review!

Couple questions for you ...
- I wasn't sure about the difference between Completable and NonCompletable objects, so I don't know if I used them correctly, please advise.
- I added my name to the list of copyright notices on each file I modified, is that the correct protocol?

github/GithubObject.py
```diff
@@ -151,6 +151,10 @@ def _makeStringAttribute(value):
     def _makeIntAttribute(value):
         return GithubObject.__makeSimpleAttribute(value, (int, long))
 
+    @staticmethod
```
bbi-yggy: I added this because the add_attribute script wants it to exist - but I'm not sure any API's have floating point attributes, so might be better to remove this method and remove support for `float` from add_attribute.

github/MainClass.py
```diff
@@ -300,6 +300,19 @@ def get_repos(self, since=github.GithubObject.NotSet):
             url_parameters
         )
 
+    def get_project(self, id):
```
bbi-yggy: You can also get project columns and cards by their global ID, but I didn't add support for those.

github/Project.py
```diff
@@ -0,0 +1,215 @@
...
+
+# Original request for project support:
+#   https://github.com/PyGithub/PyGithub/issues/606
+# and pull request with initial functionality:
+#   https://github.com/PyGithub/PyGithub/pull/854
+
+# 'Accept' header required while Github Projects API still in preview mode.
+PROJECT_PREVIEW_HEADERS = {"Accept": "application/vnd.github.inertia-preview+json"}
+
+class Project(github.GithubObject.CompletableGithubObject):
```
bbi-yggy: I made this guy "completable" but to be honest I don't know what that means.

github/ProjectCard.py
```diff
@@ -0,0 +1,175 @@
...
+        return self._updated_at.value
+
+    @property
+    def url(self):
+        """
+        :type: string
+        """
+        return self._url.value
+
+    # Note that the content_url for any card will be an "issue" URL, from
```
bbi-yggy: This is a bit unfortunate, but seems like the nature of github issues and pull requests. Note that in a browser you can use a "<repo>/pull/1" style URL to navigate to an issue, and "<repo>/issues/1" to get to a pull request.

github/tests/Project.py
```diff
@@ -0,0 +1,140 @@
...
+        issue_card = cards[1]
+        issue = issue_card.get_content()
+        self.assertIsInstance(issue, github.Issue.Issue);
+        self.assertEqual(issue.title, "Test issue")
+
+        note_card = cards[2]
+        note_content = note_card.get_content()
+        self.assertEqual(note_content, None)
+        
+    def testGetAllProjectCards(self):
```
bbi-yggy: Is there a way to generate output from a test when just that test is running, but suppress it when multiple tests are being run? Printing intermediate results would be helpful at time while developing tests.

bbi-yggy:
Oops, forgot to re-record test data after that last fix, which changed a URL.
Will do so as soon as I can (currently waiting for Github to reset my rate limit, sigh ...)

sfdye:
About "completable" or "non-completable", I am also not very sure. Have already asked the original author and waiting for his reply.

github/Organization.py
sfdye: There is actually a script (inside this repo) to generate all the credits, from git data. The maintainer is supposed to run the script from time to time to update. So I guess your email information will still be public anyway.

github/GithubObject.py
sfdye: Hmm, don't think Github will have any `float` attributes, so remove it from the `add_attribute` script altogether?

github/MainClass.py
sfdye: Okay, don't seem to have good use cases here. Future PR by someone else maybe.

github/Project.py
```diff
@@ -0,0 +1,215 @@
...
+import github.GithubObject
+import github.ProjectColumn
+
+# Original request for project support:
+#   https://github.com/PyGithub/PyGithub/issues/606
+# and pull request with initial functionality:
+#   https://github.com/PyGithub/PyGithub/pull/854
+
+# 'Accept' header required while Github Projects API still in preview mode.
+PROJECT_PREVIEW_HEADERS = {"Accept": "application/vnd.github.inertia-preview+json"}
```
sfdye: Good idea, will probably refactor all the custom headers into a constant file.

github/Project.py
sfdye: Your guess is correct.

sfdye:
@bbi-yggy Sharing what the original author of library has replied to me via email (with permission):

> The `CompletableGithubObject` class implements "lazy completion": if for some reason an attribute accessed by the user has never been retrieved from the API, this class does an implicit request to retrieve all missing fields. This happens for example when retrieving a NamedUser from an Issue's user field: some fields or the user field are not returned in https://developer.github.com/v3/issues/#response-2, but the Issue class exposes a NamedUser anyway, and this object completes itself if needed.

> More generally, every time an API returns a partial object as an attribute, and this object has an `url` attribute to get more information, PyGithub should expose a class deriving from `CompletableGithubObject`, and each field of this class should call `_completeIfNotSet` before returning.

I've also learned something 😃

github/MainClass.py
bbi-yggy: `MainClass.get_project` is a direct implementation of [this API endpoint](https://developer.github.com/v3/projects/#get-a-project).

The use case I see is if someone wants to record a project ID for future retrieval, which seems likely and reasonable. There is no other way to get to a project unless you list all projects for a repo or organization and then find the one you want - you cannot retrieve a single project from either a repo or organization, so I think this method is required. Let me know if you agree.

bbi-yggy: Oops, sorry, just realized you were commenting on my comment, not my change. Never mind!

sfdye: Yup

github/ProjectCard.py
```diff
@@ -0,0 +1,178 @@
...
+import json
+
+import github.GithubObject
+
+# NOTE: There is currently no way to get cards "in triage" for a project.
+# https://platform.github.community/t/moving-github-project-cards-that-are-in-triage/3784
+#
+# See also https://developer.github.com/v4/reference/object/projectcard for the next generation GitHub API,
+# which may point the way to where the API is likely headed and what might come back to v3. E.g. ProjectCard.content member.
+class ProjectCard(github.GithubObject.NonCompletableGithubObject):
```
sfdye: Shouldn't this be `CompletableGithubObject`?

github/ProjectColumn.py
```diff
@@ -0,0 +1,148 @@
...
+#                                                                              #
+################################################################################
+
+import json
+
+import github.GithubObject
+import github.Project
+import github.ProjectCard
+
+class ProjectColumn(github.GithubObject.NonCompletableGithubObject):
```
sfdye: Same here

sfdye:
It's a rather large PR so I don't want to rush. Maybe after you are back. Also, this PR seems to include all the "get" method, but lacking some of the "post/patch/delete" methods? Maybe we can make it more useful by implementing those as well. This was requested in the original issue #606 too.

bbi-yggy:
As I mentioned previously, I only had need for the read-only reporting methods, so I wasn't planning to implement the post/patch/delete API at this time. Of course someone else is welcome to do so. :)

github/ProjectCard.py
bbi-yggy: My understanding was that any class whose attributes could be fully retrieved from the results of the original request should be NonCompletable, since no additional API methods are needed. Which may actually suggest that Project should also be NonCompletable. From the information you cited earlier, "if for some reason an attribute accessed by the user has never been retrieved from the API, this class does an implicit request to retrieve all missing fields" - but for ProjectCard and ProjectColumn all the attributes are available after the first request.

I'm happy to be corrected if my understanding is incorrect.

github/Project.py
```diff
@@ -0,0 +1,215 @@
...
+#                                                                              #
+################################################################################
+
+import datetime
+import json
+
+import github.GithubObject
+import github.ProjectColumn
+
+# Original request for project support:
```
sfdye: Maybe these are not necessary

github/Project.py
```diff
@@ -0,0 +1,215 @@
...
+        self._html_url = github.GithubObject.NotSet
+        self._id = github.GithubObject.NotSet
+        self._name = github.GithubObject.NotSet
+        self._node_id = github.GithubObject.NotSet
+        self._number = github.GithubObject.NotSet
+        self._owner_url = github.GithubObject.NotSet
+        self._state = github.GithubObject.NotSet
+        self._updated_at = github.GithubObject.NotSet
+        self._url = github.GithubObject.NotSet
+        self._zz = github.GithubObject.NotSet
```
sfdye: What is `_zz`?

sfdye:
Sure, read-only endpoints for now then. I will try to finish rest of the review, hopefully over the weekend.

github/Project.py
bbi-yggy: Oops! That was me testing that the add_attribute script could add an attribute after all the existing attributes. I will remove.

mitar:
I do not think this PR closes #606, but it does move us closer to that.

sfdye:
@mitar It's a start. Subsequent PRs are always welcome 😉

sfdye:
@bbi-yggy If you are back from your holiday, I would like to resume this.

mitar:
I agree. I am just saying that in the opening description of this PR it says that it closes #606 so when you will merge this PR, GitHub will automatically close #606.

sfdye:
@mitar Okay, removed the closing clause. Now it's just linked.

bbi-yggy:
Just got back yesterday, ready to resume on this. @mitar is correct, it doesn't close #606. To be fair, that feature request is a little light on details, but if you read it as "Support all the capabilities of the GitHub API for working with projects" then no, I haven't addressed everything. What I've done is a good start though I hope, and I think is ready to merge in its current form.

sfdye:
Awesome, let's resume this.

github/Project.py
```diff
@@ -0,0 +1,209 @@
...
+################################################################################
+
+import datetime
+import json
+
+import github.GithubObject
+import github.ProjectColumn
+
+# 'Accept' header required while Github Projects API still in preview mode.
+PROJECT_PREVIEW_HEADERS = {"Accept": "application/vnd.github.inertia-preview+json"}
```
sfdye: Huh, some refactoring was going on while you were away. Now we have a central place to put all the preview headers. So would you mind add this in as well?

Ref: https://github.com/PyGithub/PyGithub/blob/master/github/Consts.py

github/ProjectCard.py
sfdye: My understanding is that as long as we can lazy retrieve a resource (the response has the `url` attribute, kinda like a permanent link), then the corresponding GithubObject type should be `CompletableGithubObject`. You can make comparisons with Issue/PullRequests, they have a similar relationship with org/repos. Also an observation is that, most of the types in this library are `CompletableGithubObject`. Again, I did not author the original code so I may be wrong.

CONTRIBUTING.md
sfdye: > Also note that if you record your test data with `--auth_with_token` then you should modify the setup code for your Framework.TestCase to set `self.TokenAuthMode = True` before calling `Framework.TestCase.setUp(self)`.

I am not sure if this is necessary. If I wanna authenciate with token, shouldn't I always pass the `--auth_with_token` command line parameter?

Also, I am thinking to disable the password feature, since we don't want to accidentally commit 
 password to git and some users might be concerned about leaving their password in plaintext. Always use a personal access token, with a set of permissions defined by the user. What's your thoughts on this?

CONTRIBUTING.md
```diff
@@ -51,7 +51,13 @@ oauth_token = "my_token"  # Can be left empty if not used
...
-You can use `python -m github.tests Issue139.testCompletion --record --auth_with_token` to use the `oauth_token` field specified in `GithubCredentials.py` when recording a unit test interaction. NB that the `password = ""` (empty string is ok) must still be present in `GithubCredentials.py` to run the tests even when the `--auth_with_token` arg is used.
+You can use `python -m github.tests Issue139.testCompletion --record --auth_with_token` to use the `oauth_token` field specified in `GithubCredentials.py` when recording a unit test interaction. Note that the `password = ""` (empty string is ok) must still be present in `GithubCredentials.py` to run the tests even when the `--auth_with_token` arg is used. Also note that if you record your test data with `--auth_with_token` then you should modify the setup code for your Framework.TestCase to set `self.TokenAuthMode = True` before calling `Framework.TestCase.setUp(self)`.
+
+To run manual tests with external scripts that use the PyGithub package, you can install your development version with:
+
+```
+pip install --editable (workspace)
```
sfdye: Maybe be a bit more specific?

`pip install --editable path/to/project`

github/Project.py
```diff
@@ -0,0 +1,209 @@
...
+        self._owner_url = github.GithubObject.NotSet
+        self._state = github.GithubObject.NotSet
+        self._updated_at = github.GithubObject.NotSet
+        self._url = github.GithubObject.NotSet
+
+    def _useAttributes(self, attributes):
+        if "body" in attributes:  # pragma no branch
+            self._body = self._makeStringAttribute(attributes["body"])
+        if "columns_url" in attributes:  # pragma no branch
+            assert attributes["columns_url"] is None or isinstance(attributes["columns_url"], (str, unicode)), attributes["columns_url"]
```
sfdye: I believe the underlying `_makeXXXAttribute` method have None and type checks, so the additional assert is not needed

scripts/add_attribute.py
```diff
@@ -111,7 +124,8 @@
                 attrName = line[12:-36]
             if not line or attrName > attributeName:
                 newLines.append("        if \"" + attributeName + "\" in attributes:  # pragma no branch")
-                newLines.append("            assert attributes[\"" + attributeName + "\"] is None or isinstance(attributes[\"" + attributeName + "\"], " + attributeAssertType + "), attributes[\"" + attributeName + "\"]")
+                if attributeAssertType:
+                    newLines.append("            assert attributes[\"" + attributeName + "\"] is None or isinstance(attributes[\"" + attributeName + "\"], " + attributeAssertType + "), attributes[\"" + attributeName + "\"]")
```
sfdye: Same here

github/tests/Project.py
```diff
@@ -0,0 +1,133 @@
...
+#                                                                              #
+# ##############################################################################
+
+import Framework
+import github
+
+class Project(Framework.TestCase):
+    def setUp(self):
+        # Force token authenticaton mode, since that's how the replay data was recorded.
+        self.tokenAuthMode = True
```
sfdye: I see what's you are doing here. Normally I just manually replace all the `token private_token_removed` with `Basic login_and_password_removed`. It's a bug obviously. I will think about how to improve it.

github/tests/Project.py
sfdye: I don't follow. Can't you just run a single test at a time?

CONTRIBUTING.md
bbi-yggy: Sure, that makes more sense.

github/Project.py
bbi-yggy: Sure, will do.

github/Project.py
bbi-yggy: Ah, ok - I'll double-check the behaviour of makeXXXAttribute and change if that's the case.

github/ProjectCard.py
```diff
@@ -0,0 +1,178 @@
...
+################################################################################
+
+import json
+
+import github.GithubObject
+
+# NOTE: There is currently no way to get cards "in triage" for a project.
+# https://platform.github.community/t/moving-github-project-cards-that-are-in-triage/3784
+#
+# See also https://developer.github.com/v4/reference/object/projectcard for the next generation GitHub API,
```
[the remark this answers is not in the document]
bbi-yggy: Oops, looks like the docs moved around. Removing the "reference" portion of the path fixes it. Will fix.

github/ProjectCard.py
bbi-yggy: Hmm, ok, I'll revisit this. Where in the API does "completability" get used?

github/tests/Project.py
bbi-yggy: Oh, does that work? Could the test data recorder just take care of that?

github/tests/Project.py
bbi-yggy: I meant that while I'm testing my work I would like to know details of what the test is doing - by printing additional information to the output - but when running a test suite I just want to know pass or failure status.

scripts/add_attribute.py
bbi-yggy: Right - this is the code that generates the other code you were commenting on.

github/tests/Project.py
sfdye: It works for me. Well the test framework is quite old already, I'd rather not touch it in this PR. If needed, we can discuss it in a new thread dedicated to improving the tests framework. I've been exploring the possibility to use something vcr.py or betamax, but that seems a lot of work too.

bbi-yggy: Aha, yes, search and replace for authentication mode works. I will update CONTRIBUTING.md

scripts/add_attribute.py
bbi-yggy: Note that `attributeAssertType` may have its purposes, so I am just going to remove that field from the simple attribute types.

bbi-yggy:
@sfdye All changes made, tests still pass. Ready to merge I think?

sfdye:
@bbi-yggy Yes! Great effort. Thank you!

bbi-yggy:
Happy to help. I needed the functionality and if it helps others so much the better. :-)

bbi-yggy:
(lol, had to look that up, thought it was Let's Go To Master ;)

sfdye:
Bumped version to 1.43.1. You can upgrade from PyPi now :)

Issue #757, opened 2018-04-17
For the new "requests" communications engine, implement a developer controlled retry feature.


allevin:
Reference comments in #539, conversation moved to this issue for discussion.

Request a feature to the new **_requests_** implementation for low level communications to implement a developer controlled retry feature when communication POST/GET fail.

The feature should allow the developer via a high level interface to control retry operations (number of retrys, time between retrys, etc.).  

Our network is behind an aggressive proxy server/filter and we occasionally lose responses from Github. This causes comm exceptions that could be avoided if the low level code would just perform an (optional - user configured) retry.

mfonville:
Indeed a controlled retry would be very welcome. Today I hit on the Github API (without any proxy or such) a 502 error on a trivial operation, but it would be ridiculous to try to catch those for literally API command I execute in my program:

```
  File "/home/maarten/Development/PyGithub/build/lib/github/Repository.py", line 1317, in get_commit
    self.url + "/commits/" + sha
  File "/home/maarten/Development/PyGithub/build/lib/github/Requester.py", line 258, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, self.__customConnection(url)))
  File "/home/maarten/Development/PyGithub/build/lib/github/Requester.py", line 269, in __check
    raise self.__createException(status, responseHeaders, output)
github.GithubException.GithubException: 502 {'message': 'Server Error'}
```

mfonville:
See https://github.com/PyGithub/PyGithub/pull/803 to at least cover the retries on commands that failed on Github's side

PR #762, opened 2018-04-19
Establish basis for examples

BarbourSmith:
The pyGithub documentation is excellent, but interacting with the Github api for the first time can be confusing without any examples.

There have been a number of requests for more examples of how to use PyGithub.

This pull request establishes an examples.rst file which serves as a place for the community to contribute examples.

The examples file starts off with some from @pconrad showing off the basic use of PyGithub and some from me showing how to interact with pull requests as issues.

This pull request is an attempt to begin to address issue #321

PR #860, opened 2018-08-09
Adding communications Retry functionality into requests via urllib3 retry object.

allevin:
This PR is related to #757 and #378

This is an attempt to add a communications retry feature into PyGithub that can be user defined at the top level.

This works directly with requests and uses the low level urllib3 Retry object (already used by requests) to implement the functionality.  It is backwards compatible with the current codeset.

The user can set the retry parameter in the top level `github(<login>, retry = X)` call.  Where X can be a simple `int` to define the number of retrys or it can be an `urllib3.util.retry.Retry(...)` object.  This allows the user very fine grained control over the retry system as defined by urllib3.

The structure for the Retry object can be setup like this example:
```
num_retries=3
backoff_factor=5
status_forcelist=(500, 502, 504)
retry_data = urllib3.util.retry.Retry(
                                      total=num_retries,
                                      read=num_retries,
                                      connect=num_retries,
                                      backoff_factor=backoff_factor,
                                      status_forcelist=status_forcelist
                                     )
gserver = github(<login or token>, retry = retry_data)
```
The urllib3 Retry object has many parameters, that allows the user much flexibility to configure retry's to their heart's content.

The only issue I have at the moment, is that I am completely baffled on how to add tests for this feature given the current testing infrastructure.  Any suggestions would be appreciated.

CC: @mfonville @jrouquie @sfdye 

EDIT: The code for this PR is now stale, and is being moved forward via PR #1002.  However, the discussion is still relevant.

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -85,7 +85,7 @@ class Github(object):
     This is the main class you instantiate to access the Github API v3. Optional parameters allow different authentication methods.
     """
 
-    def __init__(self, login_or_token=None, password=None, base_url=DEFAULT_BASE_URL, timeout=DEFAULT_TIMEOUT, client_id=None, client_secret=None, user_agent='PyGithub/Python', per_page=DEFAULT_PER_PAGE, api_preview=False, verify=True):
+    def __init__(self, login_or_token=None, password=None, base_url=DEFAULT_BASE_URL, timeout=DEFAULT_TIMEOUT, client_id=None, client_secret=None, user_agent='PyGithub/Python', per_page=DEFAULT_PER_PAGE, api_preview=False, verify=True, retry=None):
         """
         :param login_or_token: string
         :param password: string
@@ -96,6 +96,7 @@ class Github(object):
         :param user_agent: string
         :param per_page: int
         :param verify: boolean or string
+        :param retry: int or urllib3 Retry Object
         """
 
         assert login_or_token is None or isinstance(login_or_token, (str, unicode)), login_or_token
@@ -106,7 +107,7 @@ class Github(object):
         assert client_secret is None or isinstance(client_secret, (str, unicode)), client_secret
         assert user_agent is None or isinstance(user_agent, (str, unicode)), user_agent
         assert isinstance(api_preview, (bool))
-        self.__requester = Requester(login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify)
+        self.__requester = Requester(login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify, retry)
 
     def __get_FIX_REPO_GET_GIT_REF(self):
         """
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -48,6 +48,7 @@ import time
 import sys
 from httplib import HTTPSConnection
 import jwt
+import urllib3
 
 from Requester import Requester, json
 import AuthenticatedUser
@@ -96,7 +97,7 @@ class Github(object):
         :param user_agent: string
         :param per_page: int
         :param verify: boolean or string
-        :param retry: int or urllib3 Retry Object
+        :param retry: int or urllib3.util.retry.Retry object
         """
 
         assert login_or_token is None or isinstance(login_or_token, (str, unicode)), login_or_token
@@ -107,6 +108,7 @@ class Github(object):
         assert client_secret is None or isinstance(client_secret, (str, unicode)), client_secret
         assert user_agent is None or isinstance(user_agent, (str, unicode)), user_agent
         assert isinstance(api_preview, (bool))
+        assert retry is None or isinstance(retry, (int)) or isinstance(retry, (urllib3.util.Retry))
         self.__requester = Requester(login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify, retry)
 
     def __get_FIX_REPO_GET_GIT_REF(self):
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -87,13 +87,18 @@ class RequestsResponse:
 
 class HTTPSRequestsConnectionClass(object):
     # mimic the httplib connection object
-    def __init__(self, host, port=None, strict=False, timeout=None, **kwargs):
+    def __init__(self, host, port=None, strict=False, timeout=None, retry=None, **kwargs):
         self.port = port if port else 443
         self.host = host
         self.protocol = "https"
         self.timeout = timeout
         self.verify = kwargs.get("verify", True)
         self.session = requests.Session()
+        # Code to support retries
+        if retry != None:
+            self.retry = retry
+            self.adapter = requests.adapters.HTTPAdapter(max_retries=self.retry)
+            self.session.mount('https://', self.adapter)
 
     def request(self, verb, url, input, headers):
         self.verb = verb
@@ -113,13 +118,18 @@ class HTTPSRequestsConnectionClass(object):
 
 class HTTPRequestsConnectionClass(object):
     # mimic the httplib connection object
-    def __init__(self, host, port=None, strict=False, timeout=None, **kwargs):
+    def __init__(self, host, port=None, strict=False, timeout=None, retry=None, **kwargs):
         self.port = port if port else 80
         self.host = host
         self.protocol = "http"
         self.timeout = timeout
         self.verify = kwargs.get("verify", True)
         self.session = requests.Session()
+        # Code to support retries
+        if retry != None:
+            self.retry = retry
+            self.adapter = requests.adapters.HTTPAdapter(max_retries=self.retry)
+            self.session.mount('http://', self.adapter)
 
     def request(self, verb, url, input, headers):
         self.verb = verb
@@ -212,7 +222,7 @@ class Requester:
 
     #############################################################
 
-    def __init__(self, login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify):
+    def __init__(self, login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify, retry):
         self._initializeDebugFeature()
 
         if password is not None:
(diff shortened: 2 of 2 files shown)


mfonville:
Thanks for making this PR!
This is the right way to approach and tackle the retry issue imho, I will review and comment on the code next week after my holidays

mfonville:
The code looks good to me. It takes proper advantage of the Request API.

I think it would be a good idea to add an example of the Retry-object (prominent) to the documentation.
Because handling the `500`, `502` and `504` are the main culprit that many developers have to deal with (GitHub returns them relatively often; the frequency of `503` I don't know).

Only improvement I can think of is, whether an (artificial) unit test could be included?

mfonville:
Small note, I think `status_forcelist` should be a list like `[500, 502, 504]`

allevin:
> Small note, I think status_forcelist should be a list like [500, 502, 504]

@mfonville Not sure that is matters...  status_forcelist defined in urllib3 is an iterable object and technically a set.  So defining it either as a list or a tuple should work.  

In regards to the documentation, What file(s) should be changed?  Point me to the general location, and I will take a stab at it.  I am still a bit of a Python newbie, and still scratch my head on how the docs get created.

Lastly, I agree about an artificial unit test.  Let me play around with it a bit and once I get some time this week I will see what I can achieve.

mfonville:
I think some examples documentation was in a PR: https://github.com/PyGithub/PyGithub/pull/762
But it got closed by stale-bot.

@sfdye what is the way forward on where to put examples as documentation?

sfdye:
@mfonville I will probably start with some basic examples on the README soon.

mfonville:
I am now testing this code, and noticed that when hitting a GitHub error, it doesn't raise `github.GithubException` anymore when hitting a bad status, e.g. this doesn't work anymore:
```
except github.GithubException as e:
    if e.status == 422:
         print('422 (unprocessable) error: probably asset {0} already exists'.format(filename))
```

Because if retries are set, instead it raises `requests.exceptions.ConnectionError` with as reason _Max retries exceeded with url_. I think we should add some extra logic to PyGithub that it can still wrap these errors into the regular GithubException (I found on the internet that the orgiginal error should be wrapped in `e.args[0].reason` )

allevin:
@mfonville I have scratched my head about the "artificial test", and don't think we can do it in the current testing infrastructure.  Basically, the --record mode of the testing system records data when a response is received.  The retry system is buried down deep inside of requests and is basically transparent to PyGithub.  

So unless we put some sort of hooks (monkeycode patch of requests/urllib3? - which seems really unmaintainable?) into the system, I don't know how to force a timeout and retry.

allevin:
This is a bit off topic...

@mfonville in terms of forcing requests exceptions to become PyGithub exceptions, here is a quick attempt on my part.  What I am missing is setting the status value (I just picked 0 to force the general exception).  This seems to work somewhat in my test code, but I am not sure if this is what you are looking at implementing?  

If we want to keep discussing this, we should probably open an issue.

```
    def __requestRaw(self, cnx, verb, url, requestHeaders, input):
        original_cnx = cnx
        if cnx is None:
        cnx.request(
            verb,
            url,
            input,
            requestHeaders
        )
#        response = cnx.getresponse()
        # NEW TRY/EXCEPT BLOCK STARTING HERE
        try:
            response = cnx.getresponse()
        except  requests.exceptions.RequestException as e:
            print "DEBUG ERROR e = {0}".format(e)
            print "DEBUG ERROR e.response = {0}".format(e.response)
            print "DEBUG ERROR e.request = {0}".format(e.request)
            print "DEBUG ERROR e.args = {0}".format(e.args)
            print "DEBUG ERROR e.args[0] = {0}".format(e.args[0])
            raise self.__createException(0, None, e.args[0].reason) # Not sure about status code = 0
        # NEW TRY/EXCEPT BLOCK ENDING HERE

        status = response.status
        responseHeaders = dict((k.lower(), v) for k, v in response.getheaders())
        output = response.read()
```

mfonville:
About the tests, if it is indeed replayed at a high level, we can indeed not write a test to test the retry itself. But maybe we can write a test to at least test for the exceptions that can now occur?

I like the idea of catching the exceptions in __requestRaw. We could indeed raise a new exception there or maybe pass the page and it statuscode itself (somehow?) back again from the method, so that the `__check`-method could handle the error?
I do think it should be still part of this PR and not a new issue, because it'd is important to figure out how we will be handling errors with this new code, without breaking existing code in the future.

Today I hit when using this code to upload assets to GitHub an error, where the contents of the `requests.exceptions.ConnectionError` was:
`('Connection aborted.', OSError("(32, 'EPIPE')",))`
This is really new behavior compared to my old code, that was just able to catch bad status codes with `github.GithubException`. In this case it seems related to the AWS redirecting to a new destination, that I guess the retry doesn't handle properly for its 2nd try. And therefore it fails with the broken pipe.

allevin:
I want to bring some others into the conversation, as I really don't have the knowledge to properly discuss nor implement the changes you are looking for...

@sfdye @mikeage (main author of the requests conversion PR #664) can you guys weigh in on the discussion?  Feel free to bring in others as you guys see fit.

Here are the high points as I see them.

1. The original PR was an attempt to introduce a low level retry feature using the urllib3 retry class.  
1. Any ideas on how to test this PR within the existing framework?
1. @mfonville has brought up the issue that the new implementation of requests combined with the changes for this PR (and possibly without) are generating requests exceptions instead of PyGithub exceptions.  I believe that this behavior may have existed in the past when using httplib, but am not sure.
1. Is the prototype code shown [here](https://github.com/PyGithub/PyGithub/pull/860#issuecomment-413616422) a possible and reasonable solution to solving the handling of requests exceptions and converting into PyGithub exceptions?  Is there ways we can improve it?
1. Should the handling of requests exceptions and converting into PyGithub exceptions be part of this PR or should it be part of a separate PR?  @mfonville just about has me convinced to make it part of this one. 

Any help / suggestions / recommended code etc. would be greatly appreciated.  I am not a super strong Python coder this issue about requests and Exceptions are straining my knowledge to the limits.

mikeage:
I have no idea about testing (and indeed, the switch to requests isn't really reflected in the tests), but this code looks good to me.

To be honest, I'm not sure why we need to raise our own exception; obviously there's an advantage to hiding implementation details, but in this case, requests is so popular and standard that I think using a pygithub exception is almost a step backwards in terms of usability. But my opinion is probably worth about 1/100th of @sfdye 's here ;-)

mfonville:
I want to give a heads-up, though off-topic.

Apparently some of the difficulties I have been experiencing have been related to undocumented changes to the Github API at the same moment when trying this branch.
Apparently failed uploaded Github Release assets are not longer marked with `state = new` but with `state = starter`. I wrote to Github support about this sudden undocumented change, but did nog get a reply yet.

I will soon try to do some more testing, including the exception-handling.

allevin:
Ping @sfdye Still looking for your feedback/opinion on the last few comments from a while back.

allevin:
@sfdye, do we want to keep this one open?

mfonville:
I would like to still take a look at this, I am using it successfully in production. But still want to look at @allevin 's patches. But didn't get to it yet... (serious lack of time :-( )

isouza-daitan:
I have some cycles to work on this. As I understand, we need to address the points listed in https://github.com/PyGithub/PyGithub/pull/860#issuecomment-414424831

I'll start working on the test cases if that's ok.

allevin:
@isouza-daitan That would be fantastic.  I have moved to a different organization and no longer involved with PyGithub.  I am keeping watch on my PR's, but cannot spend any cycles working on it anymore. 

For this PR, I notice that there is now a merge conflict with a few files that will need to be corrected.  

If you can move the ball forward on this PR I (and I suspect a few others) would appreciate it.

isouza-daitan:
@allevin I have fixed the merge conflict on my fork. I also have a strategy for writing tests for the retry logic but that would add a test-only dependency (httpretty). 

We currently do not have test dependencies so I'm not sure if that's ok.

Giving a bit more detail on the tests, we can either:
1. Create a very simple test outside of the current framework that uses httpretty to mock the socket layer and simulate server errors
2. Use httpretty in the current framework, using the replay files to configure the mock requests and responses. (no changes to the current tests should be necessary)

I'm currently pursuing 2 and have most tests working correctly (code is still very ugly).

allevin:
@isouza-daitan Its been a while since I submitted the PR, and I have slept since and have reloaded my brain with other projects  I may not be too much assistance on specific details of the code (I have not actively touched PyGIthub in a few months),  However your solution for testing sounds feasible, and keeping it in the existing testing frameworks would be best.

isouza-daitan:
@allevin I have the test cody ready for the retry feature. It includes Framework changes to use httpretty. What's the best way to contribute this code? Open a brand new PR on PyGithub/PyGithub? I don't think I can open a PR for your fork of PyGithub.

allevin:
@isouza-daitan I am not sure.  @sfdye How do we handle this situation?

isouza-daitan:
w.r.t. converting retry exceptions to GithubExceptions, I believe the best approach is to use the `raise_on_status=False` parameter of the `Retry` object. This causes the original error to be returned, instead of the MaxRetryException being thrown.

We could:
1. force this parameter on the retry object
2. have it as the default, allowing the user to override
3. leave it up to the user to set it

Since this is an opt-in feature, I believe #3 is the best option, if properly documented.

seocam:
That would be really useful for me. I'm tired of getting 502 from Github and getting my scripts interrupted for that.

isouza-daitan:
I'll open a separate PR from my fork and mentions this one in the description.

isouza-daitan:
New PR at https://github.com/PyGithub/PyGithub/pull/1002

allevin:
@sfdye, @isouza-daitan has taken over the task and is moving the retry feature forward in PR #1002.  

However, this PR does have relevant info on the discussion that caused #1002 to be generated.   I believe that this PR can be closed in favor of #1002, but we might want to keep it open until #1002 is merged?

allevin:
Keep Alive

PR #857, opened 2018-08-07
Adding new attributes to IssueEvent

allevin:
See Issue #855 

The class [IssueEvent](https://github.com/PyGithub/PyGithub/blob/master/github/IssueEvent.py) is missing a large number of attributes documented in the [API](https://developer.github.com/v3/issues/events/).

This is also commented about in #653 to a degree

27 of the tests 27 known event types have tests.

**Currently Tested using Issue #30**            
- [x] subscribed
- [x] assigned
- [x] referenced               
- [x] closed                   
- [x] labeled                  

**Currently Tested using Issue/PR #538**
- [x] merged
- [x] mentioned
- [x] review_requested

**Currently Tested using Issue/PR #857**
- [x] reopened
- [x] unassigned
- [x] unlabeled
- [x] renamed
- [x] base_ref_changed
- [x] head_ref_deleted 
- [x] head_ref_restored
- [x] milestoned
- [x] demilestoned
- [x] locked
- [x] unlocked
- [x] review_dismissed
- [x] review_request_removed
- [x] marked_as_duplicate
- [x] unmarked_as_duplicate
- [x] added_to_project       
- [x] moved_columns_in_project
- [x] removed_from_project
- [x] converted_note_to_issue - Note: this event is tied into Issue #866 

This PR is now ready to be merged

Issue #516, opened 2017-01-19
Test recorder adds wrong string for token authorization


meneal:
When recording a test and using a `GithubCredentials.py` like:

```
login = "meneal"
password = ""
oauth_token = "REDACTED"
```

the test recorder places the string `token private_token_removed` in the recorded file.  When the test is run after the recording is made, the following stack trace is produced:

```
Traceback (most recent call last):
  File "github/tests/OrganizationHasInMembers.py", line 32, in setUp
    self.user = self.g.get_user("meneal")
  File "github/MainClass.py", line 159, in get_user
    "/users/" + login
  File "github/Requester.py", line 172, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
  File "github/Requester.py", line 213, in requestJson
    return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
  File "github/Requester.py", line 252, in __requestEncode
    status, responseHeaders, output = self.__requestRaw(cnx, verb, url, requestHeaders, encoded_input)
  File "github/Requester.py", line 277, in __requestRaw
    requestHeaders
  File "github/tests/Framework.py", line 151, in request
    self.__testCase.assertEqual(headers, eval(readLine(self.__file)))
AssertionError: {'Authorization': 'Basic login_and_password_removed', 'User-Agent': 'PyGithub/Py [truncated]... != {'Authorization': 'token private_token_removed', 'User-Agent': 'PyGithub/Python' [truncated]...
- {'Authorization': 'Basic login_and_password_removed',
+ {'Authorization': 'token private_token_removed',
   'User-Agent': 'PyGithub/Python'}
```

I manually changed the string in the recorded file to be `Basic login_and_password_removed` and the test passed with no error. I believe there are two options for this:

- Make the test in `Framework.py` in some way accept the `token_private_token_removed`
- Change something(not sure what this is offhand) in the recorder to make it use `Basic login_and_password_removed` regardless of whether a token or basic login is used.

sfdye:
I have the same concern with you. With a lot of users turning on 2-factor authentication, we should support record test data with user token as well. Right now I have to manually replace all `token_private_token_removed` with `login_and_password_removed`.

s-t-e-v-e-n-k:
I keep getting bitten by this, so I may get sufficiently annoyed to fix it.

Diwasdrpm:
yes, I also get it same

treee111:
[...]

PR #861, opened 2018-08-09
Fix token auth recording so playback can work on Travis

allevin:
For Token authentication when the Authorization header was recorded, it was changed in the recorded file to say "token private_token_removed".

This would work properly for a users playback testing on their own system.

However, when the tests are run on Travis (which apparently uses a basic login method) the tests would fail due to Authorization mis-compares. 

See #516 and #857

Since the user recording data most likely does not care what gets recorded (except to ensure that the authentication data remains private), this simple change will resolve the mis-compare.


sfdye:
Nice one! This used to be a pain, I have to manually change the "token private_token_removed" to "Basic login_and_password_removed".

allevin:
@sfdye Can you go over this with a fine tooth, and possibly come up with other solutions?

This solution works, however I really didn't like having to change "token private_token_removed" to "Basic login_and_password_removed" in the recorded file.   However to maintain compatibility with all the older existing tests, we have to keep the string "Basic login_and_password_removed" basically everywhere.    I tried other solutions, but this one seemed to cause the least impact.   I just wish we could do something before before the line 180 `self.__testCase.assertEqual(headers, eval(readLine(self.__file)))` that would allow us to keep "token private_token_removed" but just before the check convert it to "Basic login_and_password_removed" without changing a bizillion test results in the repo.

allevin:
ping @sfdye Is there any interest in moving forward with this PR?

sfdye:
Of course @allevin just that I have limited bandwidth now.

sfdye:
un-stale

PR #876, opened 2018-08-21
Two fixes related to redirect

lost-a-tooth:
Issue 1) 
__log method is immediately called after receiving the response, and the method wipes out Authorization header for security reasons.  Since the same reqeustHeader is used for redirect, the redirect will fail due to bad credentials.  The fix is to hide the Authorization header on the log but not actually changing the requestHeader.

Issue 2)
responseHeaders['location'] returns a fully qualified URL and requester is expecting a path, so users would get InvalidURL exception.

Ensured both issues are fixed with the new code.
Ensured that the tests were passing, with the exceptions of testDecodeJson, which is failing on master too.


sfdye:
un-stable

PR #878, opened 2018-08-22
Add a trailing slash to URL when updating or deleting a file.

edouardtheron:
solve 877


edouardtheron:
ooops... Of course build fails. 

I need to dig further then, I'll try to do it ASAP

PR #883, opened 2018-08-27
examples/Login.rst

Abdur-rahmaanJ:
basic login demo

### DIFF
--- a/doc/examples/Login.rst
+++ b/doc/examples/Login.rst
@@ -1,0 +1,15 @@
+Login
+======
+
+Basic login demo
+--------------------------
+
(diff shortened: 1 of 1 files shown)


sfdye:
Each rst file should correspond to one Github object class, there is no object called `Login`. Besides this example was already in the introduction section. If you want to elaborate further, I would recommend moving it to MainClass.rst.

Abdur-rahmaanJ:
yes but when searching for examples, the one in the introduction is the best. the intro overlaps, closing PR

sfdye:
Sure, Thank you!

Issue #744, opened 2018-04-07
Add support for searching topics


edanhauon:
Searching by topic is a feature offered in the beta version of the API.  We need this feature in order to collect data about repositories from a topic point of view.

[Searching Topics](https://help.github.com/articles/searching-topics/)
[Searching Topics by API](https://developer.github.com/v3/search/#search-topics)

Example of use:
```
g = Github()
# will return all machine-learning related topics
topics = g.search_topics(query=machine-learning) 

for topic in topics:
    print(topic.display_name)
    print(topic.featured)
    print(topic.curated)
```

sfdye:
@edanhauon Thanks for reporting. If you can put in a PR I will be willing to review it.

PR #892, opened 2018-09-02
Feature: Search By Topics

singh811:
Solves #744

### DIFF
--- a/github/BranchProtection.py
+++ b/github/BranchProtection.py
@@ -49,7 +49,7 @@ class BranchProtection(github.GithubObject.CompletableGithubObject):
     @property
     def required_status_checks(self):
         """
-        :type: :class:github.RequiredStatusChecks.RequiredStatusChecks
+        :type: :class:`github.RequiredStatusChecks.RequiredStatusChecks`
         """
         self._completeIfNotSet(self._required_status_checks)
         return self._required_status_checks.value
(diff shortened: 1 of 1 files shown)


github/Requester.py
```diff
@@ -467,4 +467,4 @@ def __log(self, verb, url, requestHeaders, input, status, responseHeaders, outpu
                     requestHeaders["Authorization"] = "token (oauth token removed)"
                 else:  # pragma no cover (Cannot happen, but could if we add an authentication method => be prepared)
                     requestHeaders["Authorization"] = "(unknown auth removed)"  # pragma no cover (Cannot happen, but could if we add an authentication method => be prepared)
-            logger.debug("%s %s://%s%s %s %s ==> %i %s %s", str(verb), self.__scheme, self.__hostname, str(url), str(requestHeaders), str(input), status, str(responseHeaders), str(output))
+            logger.debug("%s %s://%s%s %s %s ==> %i %s %s", str(verb), self.__scheme, self.__hostname, str(url), str(requestHeaders), str(input), status, str(responseHeaders), output.encode('utf-8'))
```
singh811: Since `output` is a unicode string so using `str(output)` throws `UnicodeDecodeError` in logging when str is not able to decode hex characters, so I replaced `str` by `encode`.

github/MainClass.py
```diff
@@ -510,6 +510,36 @@ def search_commits(self, query, sort=github.GithubObject.NotSet, order=github.Gi
...
+        url_parameters["q"] = ' '.join(query_chunks)
+        assert url_parameters["q"], "need at least one qualifier"
+
+        return github.PaginatedList.PaginatedList(
+            github.Repository.Repository,
+            self.__requester,
+            "/search/topics",
+            url_parameters,
+            headers={
+                "Accept": Consts.mediaTypeTopicsPreview
```
singh811: As described [here](https://developer.github.com/v3/search/#search-topics) `/search/topics` endpoint requires a custom `Accept` value in `requestHeaders`

sfdye:
Thank you for the PR. Don't forget to include unit tests. See contributing for more details on how to add it.

PR #898, opened 2018-09-06
Fix unable to trigger RateLimitExceededException.

Strikersb:
I found when authorized, the ratelimit error message is different from unauthorized.

so i add this.

and run `python setup.py tests`  is ok!


Strikersb:
I haven't written tests, but I'll try. 😆  @sfdye

sfdye:
This may help:

PR #899, opened 2018-09-08
Adding migration api wrapper

### LINKED ISSUE #818: [FR] Migration API integration
The lib is missing Migration API. It also requires custom `Accept: ` header.
* https://developer.github.com/changes/2018-05-24-user-migration-api/
* https://developer.github.com/v3/migrations/users/
* https://developer.github.com/v3/migrations/orgs/

singh811:
Closes #818


singh811:
@sfdye, I have no idea how can I test the organisation functionality. Because in order to create a migration one should be logged in as an organisation.

sfdye:
Have you checked out the [contribution guide](https://github.com/PyGithub/PyGithub/blob/master/CONTRIBUTING.md#automated-tests)?

I believe the easiest (also safe) way is to create a new organization and generate test data from it.

github/Consts.py
```diff
@@ -40,6 +40,8 @@
 # ##############################################################################
 RES_ETAG = "etag"
 RES_LAST_MODIFIED = "last-modified"
+RES_MEDIA_TYPE_KEY = "x-github-media-type"
```
sfdye: Why do you need these? Isn't `mediaTypeMigrationPreview` header enough?

github/Requester.py
```diff
@@ -376,8 +376,8 @@ def __requestEncode(self, cnx, verb, url, parameters, requestHeaders, input, enc
         if Consts.headerRateReset in responseHeaders:
             self.rate_limiting_resettime = int(responseHeaders[Consts.headerRateReset])
 
-        if "x-oauth-scopes" in responseHeaders:
-            self.oauth_scopes = responseHeaders["x-oauth-scopes"].split(", ")
+        if Consts.headerOAuthScopes in responseHeaders:
```
sfdye: 👍

sfdye:
👍 for this PR. It can be more useful to implement the get migration status and download archive as well.

singh811:
If I implement get_migration_state then should I remove the state as property?

sfdye:
I think no need. Just stick to the API.

singh811:
But what if user does migration.state and it will not fetch him the latest state

sfdye:
Well, the `state` attribute merely a "state". It's useful when you call the `get_migrations` and you can see the states of all migrations. (maybe some are `failed` and some are `exported`) If the user wants to know the real-time status of a particular migration, he/she should use `get_migration_status` instead. What do you think?

singh811:
I agree with you on this, thats why I implemented `get_latest_state`. Now I waiting for your green signal.

sfdye:
LGTM. Just one minor comment, what about renaming `get_latest_state` to `get_status` because it's returning more than just the state.

github/GithubObject.py
```diff
@@ -170,7 +170,7 @@ def parseDatetime(s):
                 # The Downloads API has been removed. I'm keeping this branch because I have no mean
                 # to check if it's really useless now.
                 return datetime.datetime.strptime(s, "%Y-%m-%dT%H:%M:%S.000Z")  # pragma no cover (This branch was used only when creating a download)
-            elif len(s) == 25:
+            elif len(s) >= 25:
```
sfdye: What's the rationale for this?

sfdye: In fact, I am not familiar with the checks here, is it because the length of the datetime string could vary, for some APIs?

singh811: I  modified this check because the migration object contains a date with timezone `2015-07-06T15:33:38-07:00` but couldn't find out a way to format it. So I now this ignores the timezone information and simple parse the date into `datetime` object.
P.S. - If you have something on (how to parse the date) please let me know.

singh811: Okay, I'll do it

sfdye: I mean can you parse the string without changing "==" to ">="

sfdye: @singh811 I really want to move this forward, can you revert to my last question? Also there seems to be some conflicts from master that needs to be resolved.

singh811: Since the date directly comes from the api there is no way to parse that date without using >= condition.

singh811: [Here](https://developer.github.com/v3/migrations/orgs/#get-the-status-of-an-organization-migration)

sfdye: Which one exactly? The max length I can see is 25.

sfdye:
Let's resume this

sfdye:
@singh811 Sorry for the long review. Thank you for your contribution!

PR #901, opened 2018-09-10
Cache fetch responses

MattPhillips1:
This pull request caches fetch responses using a static cache for the Requester class (so all objects have access to the global cache). The intent is to limit the number of requests that decrease the rate_limit. There are 3 options:

1. No cache (default behaviour) which is exactly the same as the current deployment
2. Aggressive cache which is a dictionary using the url as the key
3. Clock Cache which is a static sized cache that picks a replacement using the 'clock' method

The responses are cached below the object level so objects have to be reconstructed from the cache


MattPhillips1:
Apologies for the number of commits...

sfdye:
100 commits, can you rebase? Also there is a conflict needs to be resolved.

MattPhillips1:
The caching mechanism combines the headers of the "Not modified" response with the initial "Success" response in order to retain all the information from the 200 response that is not overridden by the 304. This means that for some requests, the comparison that `python setup.py` does (namely the `doCheckFrame` function) asserts false. I'm open to any ideas on how to go about this?

PR #910, opened 2018-09-26
Prevent .netrc from overwriting Auth header

tmshn:
# Problem

`Authorization` header is overwritten by `.netrc` even if I set `login_or_token` in `github.Github`.

## Steps to reproduce

1. Add entry for github.com in `~/.netrc`
2. Construct `Github` with different config than `~/.netrc`: `gh = Github(token)`
3. Make any request

# Cause

Since v1.40, PyGithub start using [**requests**](https://github.com/requests/requests) library for http requesting (see #664).

In **requests**, `auth` attribute  of session or requests is the primary way of authenticate. If no `auth` set, `.netrc` file is used and 'Authorization' header is overwritten.

However, PyGithub authenticates requests by simply adding 'Authorization' header in `Requester.__authenticate`, which is not the way **requests** library expected.

# Solution

Defined custom authentication function which uses `.netrc` file if-and-only-if `Authorization` header isn't set already.

### Set `trust_env=False`

This is [the official way to ignore `.netrc`](https://github.com/requests/requests/issues/2773), but it disables `.netrc` at all even when no token specified, and also disables proxy handling.

### Use custom authentication to set `Authorization` header.

This seems the cleaner way when using only **requests** library.

However, PyGithub's `Requester` support injection of http client, and the above method won't work for injected non-**requests** client.


tmshn:
@sfdye Hello, is it possible to prevent stale bot from closing this?

I understand it takes some time to review this in depth, but how long it will be, I want keep it open until you (or other maintainers) decide to merge/close this.

phith0n:
I also met this problem, it's so confusing.

sfdye:
@tmshn Finally got some time to work on this. Did you have a chance to test all the different scenario?

tmshn:
@sfdye Great!
IIRC, not "all". I just tested with the combination of with/without password and with/without `~/.netrc` file. Maybe I should test similarly with `login_or_token`, `jwt` and `client_secret`?

sfdye:
Yeah that would be great :) Don't want to break the other auth methods just by supporting `netrc` file, don't we? Let me know if you are able to test this locally.

tmshn:
@sfdye Test added for password and token authentication, so could you take a look?
(I'm planning to do similar tests for JWT and client secret authentication, but maybe later...)

jupiny:
It seems to be closed automatically even though the issue was not resolved...
Could you open it again?
I'm suffering from the same problem now...

tmshn:
@sfdye 

Finally caught up everything since 1 year ago! So now it's ready to be reviewed.

## Recent updates:

### Tests covers all auth method

I added tests for client_id/client_secret and JWT auth, so now it covers all authentication method.
(BTW, having JWT auth in `MainClass.Github` seems as weird as client_id/client_secret, because I don't think it can be used to access normal APIs directly. It's for obtaining access token, which in turn used to access normal APIs.)

### Covers `MainClass.GithubIntegration`

This is the one I didn't noticed when I work this last year. They uses `requests` library directly instead of using `Requester`.
The hardest parts is: http response record/replay mechanism is unique to `Requester`, so I have to i"re-invent" the same wheel.

tmshn:
Not stale

PR #914, opened 2018-10-02
Gists with since

itsbruce:
The Github gists api has a since parameter for the various gist searches.  This adds that parameter to Github.get_gists(), NamedUser.get_gists() and Authenticator.get_gists().  Tests and replay data added/updated as needed.

I added a Time module in the tests directory, containing just a UTCtzinfo class, which I felt I needed for one of the tests (and would probably be useful in other since/until tests).  Might not be the best name for the module but it seemed the least invasive thing to do.


sfdye:
@itsbruce Thank you for your contribution!

Issue #435, opened 2016-08-09
Get repos missing params


RSully:
`gh.get_user().get_repos()` is missing parameters: visibility, affiliation

https://developer.github.com/v3/repos/#list-your-repositories

elfosardo:
I would like to take care of this

elfosardo:
I went ahead and created PR #915

sfdye:
Fixed #915

PR #915, opened 2018-10-03
get repos missing params for authenticated users

elfosardo:
first commit to fix issue #435


sfdye:
LGTM. One thing though, would you mind sticking to the original style?

elfosardo:
@sfdye done and please let me know if I missed something

sfdye:
@elfosardo Thanks for your help!

Issue #906, opened 2018-09-18
InvalidURL on 301 - Moved Permanently response when getting repo


nv-tc-tp-test:
## Issue

When I want to retrieve a repo that has been moved, the subsequent request is sent to a badly constructed url.

## Reproduction

```
client = Github(user, pass)
client.get_repo('protoncoin/protoncoin')  # at the time of writing
```

## Suspected cause

The location parameter of a 301 response returns the full URL, while the `_requestRaw` method only expects the path part which it then appends to the api host name.

## Logs and trace

```
GET https://api.github.com/repos/protoncoin/protoncoin {'Authorization': 'Basic (login and password removed)', 'User-Agent': 'PyGithub/Python'} None ==> 301 {'server': 'GitHub.com', 'date': 'Tue, 18 Sep 2018 08:19:03 GMT', 'content-type': 'application/json; charset=utf-8', 'content-length': '155', 'status': '301 Moved Permanently', 'x-ratelimit-limit': '5000', 'x-ratelimit-remaining': '4999', 'x-ratelimit-reset': '1537262343', 'location': 'https://api.github.com/repositories/112116543', 'x-github-media-type': 'github.v3; format=json', 'access-control-expose-headers': 'ETag, Link, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval', 'access-control-allow-origin': '*', 'strict-transport-security': 'max-age=31536000; includeSubdomains; preload', 'x-frame-options': 'deny', 'x-content-type-options': 'nosniff', 'x-xss-protection': '1; mode=block', 'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin', 'content-security-policy': "default-src 'none'", 'x-runtime-rack': '0.058003', 'x-github-request-id': 'C726:6007:FD2DB:246691:5BA0B4F6'} {"message":"Moved Permanently","url":"https://api.github.com/repositories/112116543","documentation_url":"https://developer.github.com/v3/#http-redirects"}
[...]

PR #908, opened 2018-09-25
Prepare redirect url

ilsinszkibal:
I've hit a problem around 301 response, providing possible fix.

PR #916, opened 2018-10-04
Fix handling of 301 redirects

s-t-e-v-e-n-k:
Requester.__requestRaw() was handling 301 redirects by passing the full
URL to itself, when it only expects a URL path, so parse out the path,
and pass that in. Add a test for this behaviour.

Closes: #906


s-t-e-v-e-n-k:
Lightning fast! #908 can also be closed since this is merged.

Issue #874, opened 2018-08-21
PyGithub example usage


sfdye:
Motivation:
#862 #456 #321 #584 #762 

PyGithub is a great library, however it may not be intuitive to use at the beginning. Not everyone is a Github API expert, so I am trying to provide some working examples (recipes) that will hopefully help you get started.

The skeleton has been setup [here](https://github.com/PyGithub/PyGithub/tree/master/doc/examples) and the example docs is already live on readthedocs as usual:

https://pygithub.readthedocs.io/en/latest/examples.html

Note that the list will expand and **any PR to add more examples is extremely welcome** 😄 

If you want to see any examples not covered here, just leave a comment.

Abdur-rahmaanJ:
API expect -> API expert

sfdye:
Done thanks @Abdur-rahmaanJ

Abdur-rahmaanJ:
@sfdye readd the login example please ^^_

sfdye:
@Abdur-rahmaanJ What do you mean?

mscheper:
I just started with PyGithub, and [found some examples](https://pygithub.readthedocs.io/en/latest/examples.html), so obviously some work has been done—thanks, people!

But I don't see any examples for how to use events, e.g. [pushevent](https://developer.github.com/v3/activity/events/types/#pushevent) and [pullrequestevent](https://developer.github.com/v3/activity/events/types/#pullrequestevent). Could those be added?

mscheper:
Thanks, @LeviMatus!

sfdye:
this is not stable

goodship1:
Writing Organization examples now

yks0000:
I am new to PyGithub SDK and trying to do below operations. However, having difficulty going through documents to know the usage.

1. get a file content from a specific branch 

command line equivalent: 

`git checkout <<branch>> and then get file content`

**Issue:**

**I am able to get file content from master branch using this sdk but not from specific branch**

2. repo tags and its details 

command line equivalent: 

`git tag -v <<tag_id>>`

3. Update that file

4. Update user details for commit 

```
git config user.email "cicd@example.com"
git config user.name "cicd"
git commit --amend --reset-author --no-edit
```

5. Git pull that branch before commit  as other automation might have committed some other file.
6. git commit in the branch

```
git pull --no-edit
git add $FILE_NAME
git commit -m "updating $FILE_NAME"
git pull --no-edit && git push origin
```

[...]

PR #927, opened 2018-10-11
Example to show How to get Commit Date

LeviMatus:
## What does this change do?
Adds a small example of how to get at the date a commit took place.

#### doc/examples/Commit.rst
Adds a block of code that demonstrates how to get at a commit's date.

## Motivation
This example is in the spirit of #874.

When I began using PyGithub it was not readily apparent to me where the commit date was being abstracted to. Being unfamiliar with Git's API, it took some convincing for me to be satisfied that the `Commit.commit.author/committer.date` path would get me the commit date.

I would hope that this would save somebody a little bit of time in the future.

## How has this been tested?
I've built the docs locally and made sure the HTML is generated as expected.

### DIFF
--- a/doc/examples/Commit.rst
+++ b/doc/examples/Commit.rst
@@ -22,6 +22,8 @@ Get commit date
 .. code-block:: python
 
     # Datetime is stored in the GitAuthor objects in the commit's GitCommit.
-    commit = repo.get_commit(sha=sha)
-    commit.commit.author.date
-    commit.commit.committer.date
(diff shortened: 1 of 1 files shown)


sfdye:
Simply awesome, kudos to building the docs locally for test evidence! One small suggestion, could you add a real example and print out the result, like what the other examples do?

LeviMatus:
@sfdye, good catch. I've updated the example.

PR #928, opened 2018-10-11
Added better error handling when the server returns a 5XX status code

Paradoxis:
My Sentry recently came across an issue where the server raised a 502 server error, as this is an 'unknown' status code there's no way to handle the exception without having to catch all exceptions rather than specifically server-related errors (this caused a periodic celery task to fail, which lead to it not being run at all). 

The following change ensures all status codes from 500 to 599 (`599 in range(500, 599)` results in `False`, hence the 600) are given a custom exception class in `Requester.__createException`

The error in Sentry:

screen shot 2018-10-09 at 21 27 50

The following would be too broad of an exception handler:

screen shot 2018-10-11 at 09 36 46

### DIFF
--- a/github/GithubException.py
+++ b/github/GithubException.py
@@ -86,6 +86,12 @@ class RateLimitExceededException(GithubException):
     """
 
 
+class UnhandledServerException(GithubException):
+    """
+    Exception raised when a server unexpectedly raises a 5XX error
+    """
+
+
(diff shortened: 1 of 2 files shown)


Paradoxis:
Could you guys look into maybe implementing this (or at least give some feedback), thanks

PR #930, opened 2018-10-13
Example: Listen for New Events with PyGithub and Pyramid

LeviMatus:
## What does this change do?
Adds a small demo implementation of how one might use webhooks to listen for
new events in a given repository.

#### doc/examples/Webhook.rst
Adds a new `.rst` file, `Webhook.rst`, which can be used for demonstrating the usage of Webhooks in PyGithub.

The example that I've added shows how to continually listen for events using Pyramid in conjunction with PyGithub.

## Motivation
This example is in the spirit of #874. [It was asked](https://github.com/PyGithub/PyGithub/issues/874#issuecomment-429075695) that someone provide an example of how to listen to events from a repository.

## How has this been tested?
I've built the docs locally and made sure the HTML is generated as expected.

## Screenshots
webhook1
webhook2


sfdye:
I thought this `hub` endpoint is deprecated? The 2018 way is to use webhook?

LeviMatus:
I thought it was the Github Services that was being deprecated. The [PubSubHubHub section](https://developer.github.com/v3/repos/hooks/#pubsubhubbub) mentions that callback strings using `github://` will cease to work due to Services being deprecated. I'm not super well versed on the API though, so I might (probably) have missed something.

Using webhooks to do this would certainly be good. I'll see if I can get that to either replace/go along with this example.

sfdye:
@LeviMatus At least I've only used webhooks in my projects and it's been good so far. But I may be wrong too 😄 

some ref:
https://developer.github.com/v3/guides/replacing-github-services/
https://developer.github.com/v3/guides/building-a-ci-server/

LeviMatus:
I probably just misunderstand the deprecation of Github Services and its relation to `Hub`. However, you're correct in saying that webhooks are the right way of doing this.

I've updated my example accordingly. I've removed the subscription usage and opted to create a webhook that results in the same usage.

sfdye:
I am thinking whether to make it one level higher and more generic as a webhook example itself, because this will be applicable for not just repo event, but all webhook events. Also FYI there is an official example [here](https://github.com/github/platform-samples/blob/master/api/python/building-a-ci-server/server.py) by Github which happens to be authored by me some time ago, this may give you some ideas. Let me know what you think ;)

LeviMatus:
Having this be a more generic webhook example makes sense. 

This example from Github is cool. I particularly like that you can specify which functions to call based on a header value, like `X-Github-Event:push`. I wish Flask supported this type of filtering!

I'll rename and/or move this to a place more suitable for working with Webhooks. I'll also get around to trying this out in a fashion similar to that example.

sfdye:
Don't forget to add the index for the new file in `examples.rst`！

doc/examples/Webhook.rst
```diff
@@ -0,0 +1,112 @@
...
+
+        g = Github(USERNAME, PASSWORD)
+        repo = g.get_repo("{owner}/{repo_name}".format(owner=OWNER, repo_name=REPO_NAME))
+        repo.create_hook("web", config, EVENTS, active=True)
+
+
+    if __name__ == "__main__":
+        config = Configurator()
+
+        create_webhook()
```
sfdye: Maybe you can add a mention this step can also be done from Github itself. (under repo settings)

LeviMatus:
Good catches and ideas. I've added a commit to address them.

doc/examples/Webhook.rst
```diff
@@ -0,0 +1,119 @@
...
+            print("Pinged! Webhook created with id {}!".format(self.payload["hook"]["id"]))
+            return {"status": 200}
+
+
+    def create_webhook():
+        """ Creates a webhook for the specified repository.
+
+        This is a programatic approach to creating webhooks with PyGithub's API. If you wish, this can be done
+        manually at your repository's page on Github in the "Settings" section. There is a option there to work with
+        and confgurd Webhooks.
```
sfdye: small typo here

sfdye:
@LeviMatus Awesome stuff!

PR #931, opened 2018-10-15
Add a trailing slash to URL when updating or deleting a file (tests fixed)

### LINKED ISSUE #846: Creating new file in repo requires a leading /
If user calls the `create_file()` method of the `Repository` object and specifies a path to the file, and the path to the file does _not_ include a leading `/` character, the user gets an unhelpful "404 object not found" error, with no indication of which object can't be found (the file? the repo? the user?...)

[This Stack Overflow answer](https://stackoverflow.com/a/40614406/463213) mentions that you need a preceding `/` in the file path.

This is not mentioned anywhere in the documentation. I tried to update the documentation but documentation is missing? See #845.

This issue didn't take me long to track down, but this has high, high potential to drive someone nuts if they don't have black belt google-fu skills. _Please_ include this crucial piece of information in the documentation (wherever it lives)!

### LINKED ISSUE #877: Error when updating or deleting file from repo
On a public repository, I tried three things using PyGithub:

1. Create a new file on the master branch, but using the low-level API (eg, create a blob, then a tree, then commiting the new tree, then updating the branch's HEAD)
2. Updating this file (with the `Repository.update_file()` method)
3. Deleting this file (with the `Repository.delete_file()` method)

The first step, low level creation, works good. However, I receive a `404 Not found` response when trying to update or delete this file.

My first guess is that the bug comes the way the URL is build in the `Repository.py` module. From [this line](https://github.com/PyGithub/PyGithub/blob/master/github/Repository.py#L1562) I would say that there is a trailing `/` missing after `/contents`. 

I'll investigate further.

JeffreyLMelvin:
Closes #877
Closes #878
Closes #846 

Address failing tests in https://github.com/PyGithub/PyGithub/pull/878

### DIFF
--- a/doc/examples/Repository.rst
+++ b/doc/examples/Repository.rst
@@ -27,7 +27,7 @@ Get list of open issues
     >>> repo = g.get_repo("PyGithub/PyGithub")
     >>> open_issues = repo.get_issues(state='open')
     >>> for issue in open_issues:
-    ...    print(issue)
+    ...     print(issue)
     ...
     Issue(title="How to get public events?", number=913)
     Issue(title="Prevent .netrc from overwriting Auth header", number=910)
@@ -43,22 +43,22 @@ Get all the labels of the repository
     >>> repo = g.get_repo("PyGithub/PyGithub")
     >>> labels = repo.get_labels()
     >>> for label in labels:
-    ...    print(label)
+    ...     print(label)
     ...
     Label(name="Hacktoberfest")
     Label(name="WIP")
     Label(name="bug")
     Label(name="documentation")
 
-Get all the contents of a directory of the repository
------------------------------------------------------
+Get all of the contents of the root directory of the repository
+---------------------------------------------------------------
 
 .. code-block:: python
 
     >>> repo = g.get_repo("PyGithub/PyGithub")
     >>> contents = repo.get_contents("")
     >>> for content_file in contents:
-    ...   print(content_file)
+    ...     print(content_file)
     ...
     ContentFile(path=".github")
     ContentFile(path=".gitignore")
@@ -76,19 +76,19 @@ Get all the contents of a directory of the repository
     ContentFile(path="scripts")
     ContentFile(path="setup.py")
 
-Get all the contents of the repository
---------------------------------------
+Get all of the contents of the repository recursively
+-----------------------------------------------------
 
 .. code-block:: python
 
     >>> repo = g.get_repo("PyGithub/PyGithub")
     >>> contents = repo.get_contents("")
     >>> while len(contents) > 1:
-    ...   file_content = contents.pop(0)
-    ...   if file_content.type == "dir":
-    ...     contents.extend(repo.get_contents(file_content.path))
-    ...   else:
-    ...     print(file_content)
+    ...     file_content = contents.pop(0)
+    ...     if file_content.type == "dir":
+    ...         contents.extend(repo.get_contents(file_content.path))
+    ...     else:
+    ...         print(file_content)
     ...
     ContentFile(path=".gitignore")
     ContentFile(path=".travis.yml")


github/tests/ReplayData/Issue140.testGetFileContents.txt
```diff
@@ -2,7 +2,7 @@ https
 GET
 api.github.com
 None
-/repos/twitter/bootstrap/contents/js/bootstrap-affix.js
+/repos/twitter/bootstrap/contents//js/bootstrap-affix.js
```
sfdye: Why two `//`?

github/tests/ReplayData/Repository.testGetContentDir.txt
```diff
@@ -2,7 +2,7 @@ https
 GET
 api.github.com
 None
-/repos/jacquev6/PyGithub/contents/
+/repos/jacquev6/PyGithub/contents//
```
sfdye: Same here, is it necessary?

sfdye:
@AetherDeity Thank you for taking this forward. Left some comments. Would also appreciate if you could add some example usage of `ContentFile` with correct path format as per #874. That would be really awesome 😄

github/tests/ReplayData/Issue140.testGetFileContents.txt
AetherDeity: Because the requested path is appended to `/contents/`, when a leading slash is provided, it results in two in the requested URL. Similarly, if one were to request `/js//bootstrap-affix.js`, it would result in `/repos/twitter/bootstrap/contents//js//bootstrap-affix.js`. There is no logic to remove duplicate `/` in the URL, because it does not create an issue to have them.

If you would like this addressed, I can 
1. remove the leading slash from the test, which I dont think should be done, because the change would be intended to hide expected behavior
2. update the code to remove leading slash or duplicate slashes, but that seems unnecessary as duplicate slashes do not alter behavior. Additionally, if a path with duplicate slashes were provided, it means the code alters the requested path before passing it along. The response would be unchanged, but it still feels wrong.

github/tests/ReplayData/Repository.testGetContentDir.txt
AetherDeity: Not "necessary", just an artifact of how the issue is being addressed. Please see previous response.

github/tests/ReplayData/Issue140.testGetFileContents.txt
sfdye: Shouldn't we just remove the leading `/` in: https://github.com/PyGithub/PyGithub/blob/9d2129c7b9ff4dc79aa9e0e0ee6f67678f11e546/github/tests/Issue140.py#L51

AetherDeity:
@sfdye, hopefully I was able to address your comments. I have also added the requested example usages of ContentFile

doc/examples/Repository.rst
```diff
@@ -49,3 +49,92 @@ Get all the labels of the repository
...
+    ContentFile(path="MANIFEST.in")
+    ContentFile(path="README.md")
+    ContentFile(path="doc")
+    ContentFile(path="github")
+    ContentFile(path="manage.sh")
+    ContentFile(path="requirements.txt")
+    ContentFile(path="scripts")
+    ContentFile(path="setup.py")
+
+Get all the contents of the repository
```
sfdye: Add a "recursively"?

doc/examples/Repository.rst
```diff
@@ -49,3 +49,92 @@ Get all the labels of the repository
     Label(name="WIP")
     Label(name="bug")
     Label(name="documentation")
+
+Get all the contents of a directory of the repository
```
sfdye: You mean the "root directory"

doc/examples/Repository.rst
```diff
@@ -49,3 +49,92 @@ Get all the labels of the repository
...
+    ContentFile(path="setup.py")
+
+Get all the contents of the repository
+--------------------------------------
+
+.. code-block:: python
+
+    >>> repo = g.get_repo("PyGithub/PyGithub")
+    >>> contents = repo.get_contents("")
+    >>> while len(contents) > 1:
```
sfdye: PEP8 please :)

sfdye:
I think all the changes and tests look good now. Just a few comments on the example.

doc/examples/Repository.rst
AetherDeity: oops, apparently I spend too much time writing yaml files anymore.

doc/examples/Repository.rst
AetherDeity: fixed

doc/examples/Repository.rst
AetherDeity: fixed

edouardtheron:
Thank you very much @AetherDeity for finishing up what I was struggling to finish. Thanks also @sfdye

PR #938, opened 2018-10-21
Ability to filter repository collaborators

### LINKED ISSUE #937: Add the ability to filter repository collaborators
The GitHub API call for retrieving repository collaborators[allows for [filtering by affiliation](https://developer.github.com/v3/repos/collaborators/#list-collaborators).

This is not possible with the current [github.Repository.Repository.get_collaborators()](https://github.com/PyGithub/PyGithub/blob/0a10d7cdf0f69ebac6d99e9699df5696cb1e9bea/github/Repository.py#L1343-L1353) function.

hegde5:
## Changes
- Add `affiliation` parameter to `get_collaborators()` and use it to filter collaborators through the API
- Resolves #937

PS: Would have loved to add a simple test for this but there seems to be a problem on my end using OAuth as I'm just not able to authenticate to the API during tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@
 /gh-pages/
 /doc/doctrees/
 .vscode*
+.venv
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -1340,16 +1340,26 @@ class Repository(github.GithubObject.CompletableGithubObject):
             None
         )
 
-    def get_collaborators(self):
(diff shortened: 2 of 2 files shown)


hegde5:
After taking a second look at it, I was able to resolve it. Thanks a lot @sfdye for pointing me to that :+1: I have added the unit test for it.

PR #943, opened 2018-10-23
Fix unable to trigger RateLimitExceededException.

Strikersb:
I found when authorized, the ratelimit error message is different from unauthorized.

so i add this.

After studying hard with my 2b colleagues @AnYeMoWang , we finally realized how tests wrote.

😸 😸 😸 😸 😸 😸 😸


github/tests/Exceptions.py
```diff
@@ -123,3 +123,12 @@ def exceed():
                 g.get_user("jacquev6")
 
         self.assertRaises(github.RateLimitExceededException, exceed)
+
+    def testAuthenticatedRateLimitExceeded(self):
+
+        def exceed():
+            for i in range(100):
```
sfdye: Interesting, how fast does the test run? I am a little worried about the running time because of the loop here

AnYeMoWang:
```
└─[0] <git:(fixed_ratelimit c43acd5) > python -m github.tests SpecificExceptions.testAuthenticatedRateLimitExceeded
.
----------------------------------------------------------------------
Ran 1 test in 0.052s

OK
```

sfdye:
@Strikersb Thank you for this PR.

PR #946, opened 2018-10-25
Added missing params to constructor when reversed

### LINKED ISSUE #945: PaginatedList reversed property loses http headers
In reversed(), 'headers' parameter is not passed to PaginatedList(). It makes some APIs not reversible. For example, get_stargazers_with_dates() which requires "Accept: application/vnd.github.v3.star+json" header in the API call.

netsgnut:
This PR proposes adding two missing params to constructor in `.reversed`. This preserves the `headers` as well as the `list_item` once cloned.

Fixes #945.


sfdye:
A unit test would be nice, as this change has a rather large impact.

netsgnut:
@sfdye Indeed.

Would adding a `github/test/Issue945.py` with tests that assert a `reversed` `PaginatedList` to share the `headers` and `list_item` as the original be sufficient?

sfdye:
Yes as long as it can demonstrate the original header does not get lost.

netsgnut:
Thanks @sfdye. I have added a new test file for that. Hope that covers what we should.

sfdye:
@netsgnut Thank you for your PR!

kelvinwonghk01:
@sfdye Thank you (and the whole maintainer team) for the project! :)

netsgnut:
@sfdye Pleasure is mine. Thank you all for this project ;)

PR #947, opened 2018-10-25
Add methods to add and remove repos from installation

ardakuyumcu:
Adding methods for the two APIs below:
* https://developer.github.com/v3/apps/installations/#add-repository-to-installation
* https://developer.github.com/v3/apps/installations/#remove-repository-from-installation

Two reasons why the methods are in `AuthenticatedUser.py` and not in `Installation.py`:
1. The API endpoints start with `/user`
1. The APIs require a personal access token, whereas `Installation.py` uses an installation access token.


sfdye:
Sorry let's keep this open. Would you mind resolving the merge conflicts so I can review? Thank you @ardakuyumcu

PR #948, opened 2018-10-25
Add support for JWT authentication

ardakuyumcu:
APIs such as https://developer.github.com/v3/apps/#find-organization-installation use a JWT for authentication. Adding support for JWT auth.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -214,10 +214,12 @@ class Requester:
 
     #############################################################
 
-    def __init__(self, login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify):
+    def __init__(self, login_or_token, password, jwt, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify):
         self._initializeDebugFeature()
 
-        if password is not None:
+        if jwt is not None:
+            self.__authorizationHeader = "Bearer " + jwt
+        elif password is not None:
             login = login_or_token
             if atLeastPython3:
                 self.__authorizationHeader = "Basic " + base64.b64encode((login + ":" + password).encode("utf-8")).decode("utf-8").replace('\n', '')  # pragma no cover (Covered by Authentication.testAuthorizationHeaderWithXxx with Python 3)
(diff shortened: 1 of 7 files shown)


github/tests/Framework.py
```diff
@@ -241,6 +243,7 @@ def setUp(self):
             self.oauth_token = "oauth_token"
             self.client_id = "client_id"
             self.client_secret = "client_secret"
+            self.jwt = "jwt"
```
sfdye: Shouldn't you add a `activateJwtAuthMode` in this file as well? Since this PR adds support for authenticating with JWT, we should support recording tests through it right?

sfdye:
@ardakuyumcu Awesome addition you've got here. Did you test this change with an actual Github app? I will review the other two PRs once we close this one 👍

github/tests/Framework.py
ardakuyumcu: @sfdye thanks for the suggestion, I added the mode and verified that it works.

ardakuyumcu:
@sfdye yes, I tested the JWT mode with the API I added in https://github.com/PyGithub/PyGithub/pull/949 which actually uses a Bearer JWT token for auth.

github/Requester.py
```diff
@@ -214,10 +214,12 @@ def _initializeDebugFeature(self):
 
     #############################################################
 
-    def __init__(self, login_or_token, password, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify):
+    def __init__(self, login_or_token, password, jwt, base_url, timeout, client_id, client_secret, user_agent, per_page, api_preview, verify):
         self._initializeDebugFeature()
 
-        if password is not None:
+        if jwt is not None:
```
sfdye: Would you mind moving this case after the `login_or_token`?

github/tests/__main__.py
```diff
@@ -44,6 +44,10 @@ def main(argv):
         github.tests.Framework.activateTokenAuthMode()
         argv = [arg for arg in argv if arg != "--auth_with_token"]
 
+    if "--auth_with_jwt" in argv:
```
sfdye: One final touch, this could be documented here in the contribution guide:

ardakuyumcu:
@sfdye added the requested changes

sfdye:
@ardakuyumcu Awesome, thank you very much!

blast-hardcheese:
This option breaks positional argument compatibility 😞

PR #949, opened 2018-10-25
Add find organization installations API

ardakuyumcu:
Adding support for https://developer.github.com/v3/apps/#find-organization-installation

Depends on https://github.com/PyGithub/PyGithub/pull/948


github/tests/ReplayData/Github.testFindOrganizationInstallation.txt
```diff
@@ -0,0 +1,10 @@
+https
+GET
+api.github.com
+None
+/orgs/arda/installation
+{'Authorization': 'Basic login_and_password_removed', 'User-Agent': 'PyGithub/Python', 'Accept': 'application/vnd.github.machine-man-preview+json'}
```
ardakuyumcu: This API does not use `Basic` auth, test works because this is mock data. There's no simple way to toggle `self.g` in `Framework.py`, so maybe this is fine.

sfdye:
Sorry this should not be auto-closed, will review again.

tsipa:
How review is going?

PR #950, opened 2018-10-26
raise BadCredentialException when using Github enterprise

slinkymanbyday:
As per Issue 934

github enterprise does not give the same error message. Having this catch all 401 messages.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -287,10 +287,10 @@ class Requester:
         return cnx
 
     def __createException(self, status, headers, output):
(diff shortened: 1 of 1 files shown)


sfdye:
Hi @slinkymanbyday Thank you very much for your PR. It would be helpful if you can add a unit test. You can refer to the contribution [docs](https://github.com/PyGithub/PyGithub/blob/master/CONTRIBUTING.md#adding-missing-attributes-for-a-githubobject) for instructions.

Issue #144, opened 2013-03-07
Store original JSON response for objects and expose it for document-oriented storage


ptwobrussell:
I'd like to be able to serialize a GithubObject and get back the original JSON that was used to initially construct it for purposes of serializing out to a document oriented store like MongoDB where this representation would be a natural (and highly portable) fit for storage. From looking at how objects are constructed, it seems as though it would be pretty trivial to keep around the JSON and expose it as necessary as opposed to just discarding it.

Perhaps `__str__` could return the readable JSON representation and `__repr__` could contain what currently happens from `__str__` (the unambiguous object with its address)?

If there are concerns over memory, perhaps this could be a configurable option?

At any rate, I could help put together the patch but would first like to get your thoughts and whether or not you're open to merging in such a pull request

enricorotundo:
Took me a while to find out that `raw_data` property returns a dict. I see in other libraries they call it something linke `as_dict`, that makes it easier to find via the search bar... shall I PR a few doc lines?

PR #956, opened 2018-11-06
Improved docstring for raw_data and raw_headers properties #144

enricorotundo:
See my comment in #144


sfdye:
I am not sure if this is the right place to put the docs. Maybe to get more attention, you can add the note to:

https://github.com/PyGithub/PyGithub/blob/master/doc/examples/MainClass.rst

or

even at the root `README`?

PR #963, opened 2018-11-22
Use total_count from data in PaginatedList

### LINKED ISSUE #960: PaginatedList.totalCount is broken when there are 0 results
Self-contained test:
```
from github import Github

github = Github()
repos = github.search_repositories('shouldreturn0repos')

assert repos.totalCount == 0
```

The `totalCount` method has this code:
```
    def totalCount(self):
        if not self.__totalCount:
            [...]
            if 'link' not in headers:
                self.__totalCount = len(data) if data else 0
            [...]
```
The response has no `link` header but it has this data:
```
{"total_count":0,"incomplete_results":false,"items":[]}
```
and `totalCount` returns 3 because there are 3 items inside the data dict.

I'm not sure why the `total_count` value in the response is not used directly.

s-t-e-v-e-n-k:
When there is no link header sent, this signifies that this is the only
page of results. If data is empty, or only contains a list of items, we
should count them, but this can be defeated -- if the returned data
contains a total_count key, we should use it. This also uncovered a bug
in the test suite for searching topics.

Closes #960

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -155,7 +155,12 @@ class PaginatedList(PaginatedListBase):
                 headers=self.__headers
             )
             if 'link' not in headers:
-                self.__totalCount = len(data) if data else 0
+                if data and "total_count" in data:
+                    self.__totalCount = data["total_count"]
+                elif data:
+                    self.__totalCount = len(data)
+                else:
+                    self.__totalCount = 0
             else:
                 links = self.__parseLinkHeader(headers)
                 lastUrl = links.get("last")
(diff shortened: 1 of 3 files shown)


github/PaginatedList.py
```diff
@@ -155,7 +155,12 @@ def totalCount(self):
                 headers=self.__headers
             )
             if 'link' not in headers:
-                self.__totalCount = len(data) if data else 0
+                if data and data.has_key("total_count"):
```
sfdye:
```suggestion
                if data and "total_count" in data:
```
`in` operator is better than `has_key`
https://stackoverflow.com/questions/1323410/has-key-or-in

sfdye:
Thanks for the PR! See my suggestion

github/PaginatedList.py
s-t-e-v-e-n-k: Nice catch! I've pushed up that change.

PR #966, opened 2018-11-23
Add transfer support

titouanfreville:
Add support for the new `transfer` endpoint `https://developer.github.com/v3/repos/#transfer-a-repository`.

I do not have idea to correctly test it as it need two users or organisations and repos in it.

I am using it in small script to transfer repository from an user to an organisation and seems to work well.

I hope I correctly followed the contributing guidelines.

### DIFF
--- a/github/Consts.py
+++ b/github/Consts.py
@@ -54,6 +54,9 @@ defaultMediaType = "application/octet-stream"
 
 # Custom media type for preview API
 
+#  https://developer.github.com/v3/repos/#transfer-a-repository
+mediatTypeTransferPreview = 'application/vnd.github.nightshade-preview+json'
+
 # https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/
 mediaTypeStarringPreview = "application/vnd.github.v3.star+json"
 
--- a/github/Consts.py
+++ b/github/Consts.py
@@ -55,7 +55,7 @@ defaultMediaType = "application/octet-stream"
 # Custom media type for preview API
 
 #  https://developer.github.com/v3/repos/#transfer-a-repository
-mediatTypeTransferPreview = 'application/vnd.github.nightshade-preview+json'
+mediaTypeTransferPreview = 'application/vnd.github.nightshade-preview+json'
 
 # https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/
 mediaTypeStarringPreview = "application/vnd.github.v3.star+json"
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@
 /doc/doctrees/
 .vscode*
 .venv
+venv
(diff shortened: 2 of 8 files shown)


sfdye:
Maybe you can create a new account and test it? Would still be better to use some unit tests.

yigitbey:
there was a typo in Consts.py. I fixed it on this PR: https://github.com/titouanfreville/PyGithub/pull/2

github/Consts.py
```diff
@@ -54,6 +54,9 @@
 
 # Custom media type for preview API
 
+#  https://developer.github.com/v3/repos/#transfer-a-repository
+mediatTypeTransferPreview = 'application/vnd.github.nightshade-preview+json'
```
sfdye:
```suggestion
mediaTypeTransferPreview = 'application/vnd.github.nightshade-preview+json'
```

sfdye:
Please see the docs for how to make tests work:

massdosage:
This looks like a really useful feature - we'd love to have something like this to migrate repos between organisations in a GitHub enterprise instances and currently can't see any way to do this using PyGithub. Is there something I can help with to get this re-opened and merged in?

titouanfreville:
@massdosage The feature in itself should be working (it worked when I used it). I currently do not have the time to go through the unit testing process, as I'm not well versed in Python and its unit testing, it is a bit hard. If someone have the time to fix the unit test, it would be nice. :/

massdosage:
@titouanfreville OK. I will probably try out applying your patch to a local version and ensuring it works for org->org transfers (sounds like you tested user->org?) I'll comment back here as to how that goes. I'll hopefully have some time in a few days and will see if I can get the tests working.

titouanfreville:
Yes I used it to move from user to org and the way back but not between orgs.

massdosage:
OK, so I've forked from @titouanfreville's repo to mine at https://github.com/massdosage/PyGithub. I made some small changes to the `Repository.transfer` method (adding unicode string support and renaming and code formatting to match the rest of the class). Then I had a go at the unit tests - see https://github.com/massdosage/PyGithub/blob/master/github/tests/Repository.py#L431. It appears as if the unit tests are based on the config of the user running them so for this test I wanted to test transferring a repo but the only way I could do that was with my own personal account (`massdosage`) and an organisation that I am owner of (`HotelsDotCom`). The recorded replay data (https://github.com/massdosage/PyGithub/blob/master/github/tests/ReplayData/Repository.testTransfer.txt) is thus completely bound to my user and won't work for anyone else who runs that test. Can one of the PyGihub committers please explain what we should do here to ensure the tests are more generic and can be run by anyone?

I've also tested this patch applied locally and successfully transferred repos between a user and org and back again and also between orgs. So I think the transfer code is fine, it's just a question of sorting out the unit test.

PR #978, opened 2018-12-04
Eval is unsafe

labrys:
Replace `eval` calls with `ast.literal_eval`


labrys:
Closing because upstream does not care to fix issues with code, even security issues.

PR #979, opened 2018-12-04
Fix bugs in default args

labrys:
Fix mutable data structure used for argument default
Fix call in argument default


cclauss:
This helps to avoid the problem discussed at https://docs.python-guide.org/writing/gotchas/

labrys:
Closing because upstream does not care to fix issues with code.

PR #980, opened 2018-12-04
Fix variable shadows a python builtin

labrys:


labrys:
Closing because upstream does not care to fix issues with code.

PR #993, opened 2018-12-12
Adding support for pending team invitations

### LINKED ISSUE #823: Add support for pending team invitations
Per the [docs](https://developer.github.com/v3/teams/members/#list-pending-team-invitations), it's just the following call:

```
GET /teams/:team_id/invitations
```

As mentioned in a blog [post](https://developer.github.com/changes/2018-01-25-organization-invitation-api-preview/) earlier this year, there are two new fields added to the response:

- team_count
- invitation_teams_url

These new fields only appear in the response if the following header is added to the request:

```
Accept: application/vnd.github.dazzler-preview+json
```

So I guess the only question would be whether or not to add the header? Given this is an existing endpoint, probably better to NOT include the header; then when Github moves the new "feature" (which includes the new fields) out of preview, it shows up in the response and PyGithub doesn't have to be changed to remove the above header.

shibasisp:
fixes #823 
@sfdye can you please review it? :smiley:


sfdye:
Looks good but can you add some unit tests to validate?
See docs: https://github.com/PyGithub/PyGithub/blob/master/CONTRIBUTING.md#automated-tests

shibasisp:
@sfdye Can you please review it? I have already added the tests.

shibasisp:
@sfdye  Can you please clarify what is the problem in this PR?

sfdye:
@shibasisp Looks good now. Merged.

shibasisp:
Thanks! 😃

PR #994, opened 2018-12-13
fixes issue#992

shibasisp:
The issue was because, `self._login.value` was being printed without fetching from the API.


HebaruSan:
Is there a recommended way to force retrieval of the login for a token-auth'd user? This seems to block creation of forks, which I'd like to do.

PR #998, opened 2018-12-19
Drop httplib for requests to finish work of #664

blatinier:
#664 got rid of most of httplib and replaced it with requests.
This is the last place it was used. It makes a more coherent code base I think.

Tests are ok locally.


sfdye:
Thank you! @blatinier

blatinier:
My pleasure! :)

PR #1002, opened 2018-12-26
Adding communications Retry functionality into requests via urllib3 retry object.

### LINKED ISSUE #757: For the new "requests" communications engine, implement a developer controlled retry feature.

isouza-daitan:
This is a continuation of the work started by @allevin on https://github.com/PyGithub/PyGithub/pull/860.

I have refactored the testing Framework to use the `httpretty` library in order to use the urllib3 retry feature and therefore be able to test it.

Please refer to https://github.com/PyGithub/PyGithub/pull/860 for additional context.

cc: @allevin @mfonville @jrouquie @sfdye

Closes #757 
Closes #860


isouza-daitan:
Hope everyone had a great new year! :tada: 

Any feedback on this PR? How can we move this forward?

mzizzi:
:+1: on this

github/Requester.py
```diff
@@ -89,13 +89,18 @@ def read(self):
...
-    def __init__(self, host, port=None, strict=False, timeout=None, **kwargs):
+    def __init__(self, host, port=None, strict=False, timeout=None, retry=None, **kwargs):
         self.port = port if port else 443
         self.host = host
         self.protocol = "https"
         self.timeout = timeout
         self.verify = kwargs.get("verify", True)
         self.session = requests.Session()
+        # Code to support retries
+        if retry != None:
```
cclauss: Instead of __if retry != None:__, it might be safer to use __if retry:__ to deal with the case where the if the caller passes in __retry=False__.

github/Requester.py
```diff
@@ -115,13 +120,18 @@ def close(self):
...
-    def __init__(self, host, port=None, strict=False, timeout=None, **kwargs):
+    def __init__(self, host, port=None, strict=False, timeout=None, retry=None, **kwargs):
         self.port = port if port else 80
         self.host = host
         self.protocol = "http"
         self.timeout = timeout
         self.verify = kwargs.get("verify", True)
         self.session = requests.Session()
+        # Code to support retries
+        if retry != None:
```
cclauss: Instead of __if retry != None:__, it might be safer to use __if retry:__ to deal with the case where the if the caller passes in __retry=False__.

github/tests/AllTests.py
```diff
@@ -103,6 +103,7 @@
 from BadAttributes import *
 from Equality import *
 from Search import *
+from Retry import *
```
cclauss: This file already has lots of [_wildcard_ imports (from xyz import *)](https://www.python.org/dev/peps/pep-0008/#imports) but do we _need_ to continue that anti-pattern in the case of __Retry__?

github/Requester.py
isouza-daitan: We check if the `retry` parameter is an `int' at https://github.com/PyGithub/PyGithub/pull/1002/files#diff-408d795214c69592c37ec1f499122deaR122

But I guess it wouldn't hurt to change this to `if retry:` since `retry=0` would be the same as no retry.

github/tests/AllTests.py
isouza-daitan: I believe this is a special case where the wildcard makes sense since this is an aggregation of all test cases. It is more readable and more meaningful to have wildcard imports here, IMO.

cclauss: Understood.

isouza-daitan:
@sfdye can we merge this with one approval or do we need another review?

ccaspers:
For anyone interested: I am using this in a little crawler I have written and I had absolutely no issues with it. The crawler fetches information about issues and pull requests. The largest project I have crawled had ~12.000k issues and ~3500 pulls. My crawler became a lot more stable since it started using retries.

sfdc-afraley:
Would love to see this implemented.  Having to wrap every call with 502 error retry handling makes for some pretty ugly code :)

isouza-daitan:
@sfdye can we merge this PR?

gjabouley-invn:
Hello, can this PR be merged and a new version cut?
GitHub API is quite erratic, and having capability to simply setup a retry count on the main class would be soo helpfull (instead of all those while true: try: except: 502 ...)

sfdye:
Merged and bumped to 1.43.6

allevin:
@isouza-daitan, 

Congrats this was a long time coming.  Thanks for taking the ball and driving it across the goal line!

Also, thanks to @mfonville and @sfdye for the assists!

isouza-daitan:
@allevin thanks for starting this process and guiding me through it!

PR #1008, opened 2019-01-04
No longer blindly mask JSON parsing errors

s-t-e-v-e-n-k:
If the JSON returned fails to parse *for whatever reason*, it is wrapped
in a dictionary and blindly passed on. Where, for example, PaginatedList
then returns a mostly unpopulated object and raises an URL exception
when attempting to complete it. This has wasted my time for the last
time. Clean up all of the unparseble JSON in the replay data (which is
only fair, since I added most of it), and only wrap the data if it
starts with a <html> tag to deal with 503 errors and the like.


s-t-e-v-e-n-k:
As an another idea, we could rip that special case out, and only try and to tease out JSON if the response isn't 5xx. I was tempted to say only 200, but GitHub annoying makes use of lots of 4xx responses in its API.

s-t-e-v-e-n-k:
@sfdye Any opinion on this change?

PR #967, opened 2018-11-24
Fix Python 3 compatibility

labrys:
This fixes the following python 3 compatibility issues

1. Fixes relative imports as all imports are absolute in Python 3
2. Fixes unicode not available in Python 3 (All strings are unicode)
3. Fixes long not available in Python 3 (All integers are long)
4. Fixes iterkeys, itervalues, and iteritems not available in Python 3
5. Fixes print is now a function in Python 3
6. Fixes leading zeros in a non-zero decimal number invalid in Python 3
7. Fixes invalid excpetion syntax

Additionally it fixes the following issues:
1. Fixes missing imports
2. Fixes unused imports
3. Fixes duplicate property
4. Fixes incorrect conditional statement

Fixes #965

PR #1024, opened 2019-01-31
Use print() function in both Python 2 and Python 3

cclauss:
@sfdye Can I please get your help in closing these syntax errors?  Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.


labrys:
This was already addressed in #967 which was ignored.

PR #1071, opened 2019-03-22
Add feature for get repo from another user

HerlanAssis:
Hello, sometimes I want to recover the repository in which I collaborate, but if I use the method get_repo (name) will return error because the default request is for the owner of the repository, to get around this problem I created a small function that searches the repository by full name and so I can instantiate the 'Repositoty' object, even as a contributor.

Can this be helpful to you?

### DIFF
--- a/github/AuthenticatedUser.py
+++ b/github/AuthenticatedUser.py
@@ -895,6 +895,19 @@ class AuthenticatedUser(github.GithubObject.CompletableGithubObject):
             None
         )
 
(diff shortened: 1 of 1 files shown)


sfdye:
Why not just use:
```
g = Github()
g.get_repo(fullname)
```

HerlanAssis:
You're right, I did not realize that. I'll read the documentation again. Thank's

PR #1075, opened 2019-03-29
Add url parameter to include anonymous contributors in get_contributors()

alexmusa:
Add the anon parameter missing from get_contributors() so anonymous contributors can be included in results as specified in the [Github REST API Reference](https://developer.github.com/v3/repos/#list-contributors).


WofWca:
Are anon contributors get resolved to `NamedUser`s correctly?

PR #1099, opened 2019-04-22
Fix permissions (no exec bit) for data and python modules (without shebang) files

yarikoptic:


yarikoptic:
That is sad that such an easy PR was not reviewed/merged

sfdye:
Hi @yarikoptic , what is the rationale for this PR by the way? I noticed the exec permission bit is inconsistent across the whole project, so only change for these files?

yarikoptic:
May be there is more. I have fixed the ones I spotted being wrong - more correct is better than more wrong ;-)

PR #1100, opened 2019-04-23
PullRequest: Add draft attribute

MarcoFalke:
This requires a header to be set: https://developer.github.com/v3/pulls/#list-pull-requests


MarcoFalke:
Anything left to do here? I think you'd prefer to wait until the feature is no longer in the beta stage? Though, I think that it should be provided to pygithub users because it is supposed to mirror the GitHub API?

sfdye:
A unit test would be nice. Does this attribute require custom header?

sfdye:

MarcoFalke:
Indeed, this requires a header to be set: https://developer.github.com/v3/pulls/#list-pull-requests

sfdye:
Awesome! Just one minor comment, normally each test file corresponds to one entity, since `draft` is an attribute on `PullRequest`, I think you can just append the test case in the existing test file for `PullRequest`.

MarcoFalke:
I think I'll wait for GitHub to get this out of the beta first, so that the recorded API calls reflect the stable interface.

MarcoFalke:
I think they put it out of beta, so I will take another look here.

sfdye:
Yes please

peter-evans:
Would be great to have this feature. It looks like the tests are passing. Are there outstanding tasks?

MarcoFalke:
The docs still claim you need to set a header https://developer.github.com/v3/pulls/#list-pull-requests

Which seems untrue, because I could generate the test data without setting that header.

peter-evans:
Perhaps they have removed the need for the header but not publicly announced it yet. It looks like they announce when features become part of the API officially [here](https://developer.github.com/changes/). I don't see an announcement for the draft attribute yet.

MarcoFalke:
```
$ curl https://api.github.com/repos/PyGithub/PyGithub/pulls   -H "Accept: application/vnd.github.shadow-cat-preview+json"|grep '"draft": '|wc -l
9
$ curl https://api.github.com/repos/PyGithub/PyGithub/pulls |grep '"draft": '|wc -l
0
```

Still not sure how I could record the trace without setting the flag in the header. Maybe I should try to regenerate the test data.

peter-evans:
Looks like it's still in preview based on the documentation [here](https://developer.github.com/v3/pulls/#list-pull-requests), but it seems unlikely that the interface will change now. Would you consider releasing this using the `application/vnd.github.shadow-cat-preview+json` media type header?

MarcoFalke:
I am pretty sure I use this in production without specifying the header. Also, the unit tests recording was done without setting a header. I don't know how to proceed here.

s-t-e-v-e-n-k:
Since other pull requests implementing this have landed, I'm going to close this. Thanks!

Issue #580, opened 2017-05-20
Delete/restore branch associated with a pull request


wrjohnson:
GitHub Pull Request page provides a [Delete Branch] that invokes ..../pull/:number/cleanup

Need equivalent Pull Request methods to remove and/or restore the branch associated with the pull request

wrjohnson:
With a little help from http://stackoverflow.com/questions/39065544/how-do-i-remove-all-stale-branches-from-github.   

One covoluted recipe to delete the branch for pull request "pr_num" is:
```
repo.get_git_ref( "heads/%s"%(repo.get_pull(pr_num).head.ref) ).delete()
```

sfdye:
@wrjohnson This is very useful! Would you like to make an PR? I will be happy for review.

austinsasko:
@sfdye is this issue still something that can be worked on/contributed to? If so, mind if I submit the work for the tests/code in a PR?

sfdye:
@austinsasko Go for it!

chouetz:
@sfdye this issue seems closed, could it be possible to close it?

austinsasko:
The PR for this was never merged, is this expected? https://github.com/PyGithub/PyGithub/pull/1784

PR #1104, opened 2019-04-29
Delete brach associated with pr

mklauber:
Issue #580 Added a `delete_branch` flag to the `pr.merge(...)` method.  If the flag is true, it checks if the branch is part of other PRs.  If it is not, then it deletes the branch.

### DIFF
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -781,7 +781,7 @@ class PullRequest(github.GithubObject.CompletableGithubObject):
 
             # Check for other pull requests referencing the same ref.
             # If found, abort.
-            remaining_pulls = self.head.repo.get_pulls(self.head.ref)
+            remaining_pulls = self.head.repo.get_pulls(head=self.head.ref)
             if remaining_pulls > 0:
                 raise GithubException.GitRefInUseException(data="PRs referencing this branch remain.  Not deleting the branch")
 
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -55,7 +55,7 @@ import github.File
 import github.IssueComment
 import github.Commit
 import github.PullRequestReview
-
+import github.GithubException
 
 class PullRequest(github.GithubObject.CompletableGithubObject):
     """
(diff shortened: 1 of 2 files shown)


github/PullRequest.py
```diff
@@ -775,6 +775,19 @@ def merge(self, commit_message=github.GithubObject.NotSet, commit_title=github.G
             self.url + "/merge",
             input=post_parameters
         )
+
+
+        if delete_branch == True:
```
sfdye:
```suggestion
        if delete_branch:
```

github/PullRequest.py
```diff
@@ -775,6 +775,19 @@ def merge(self, commit_message=github.GithubObject.NotSet, commit_title=github.G
...
+
+            # Check for other pull requests referencing the same ref.
+            # If found, abort.
+            remaining_pulls = self.head.repo.get_pulls(head=self.head.ref)
+            if remaining_pulls > 0:
```
sfdye:
```suggestion
            if remaining_pulls:
```
I guess this will work already

sfdye:
Looks good! I will be more convinced if a unit test can be added. See:

github/PullRequest.py
```diff
@@ -775,6 +775,19 @@ def merge(self, commit_message=github.GithubObject.NotSet, commit_title=github.G
...
             input=post_parameters
         )
+
+
+        if delete_branch:
+
+            # Check for other pull requests referencing the same ref.
+            # If found, abort.
+            remaining_pulls = self.head.repo.get_pulls(head=self.head.ref)
+            if remaining_pulls:
```
sfdye: What if I want to force delete a branch? Then `delete_branch=True` will not work, if some other PR is still referencing it right? If this the intended behavior, I am not so sure.

mklauber: Not sure.  I don't think merging one PR should close other PRs without explicit confirmation. 
 Perhaps deleting a branch should be done via a separate call?  So I take the code I added and put it in a `delete_branch(force=False)` method? 

So typical usage would be
```
pr = repo.get_pull(...)
pr.merge(...)
pr.delete_branch(force=True)
```

mklauber: Actually that would make a lot of sense, because then the api for restoring a branch is similar.

jwillenbring:
This would be a very useful feature.

PR #1129, opened 2019-05-28
Support list with protected branches only

ericofusco:
Add support to protected flag parameter when calling List branches endpoint.

https://developer.github.com/v3/repos/branches/#list-branches


ericofusco:
Can this be merged?

s-t-e-v-e-n-k:
Can you add a unit test?

PR #1132, opened 2019-06-01
Add two factor attributes on organizations

brownierin:
If you have the role of `org:admin` on an organization, you can view whether or not a second factor is required in order to be a member of an organization. (See https://developer.github.com/v3/orgs/#get-an-organization for more information). This PR adds `two_factor_requirement_enabled` as an attribute to the organization model. 

If a user doesn't have the correct permissions set on their personal access token or doesn't have `org:admin`, this value will be null in the JSON and `None` in pygithub.


dsully:
+1 for this change.

PR #1021, opened 2019-01-25
Implement OAuth for apps

rigaspapas:
This PR adds support for:
* Getting OAuth login URL for apps
* Getting access tokens for apps
* Listing app installations that are accessible to a user

Related to #828

Issue #828, opened 2018-06-26
Support for github apps


AdyaAbhra:
I am trying to perform oauth using my github app and then read authenticated users repository and organization. However for queries related to "list installations for user", "list repositories accessible to the user for installation" and many other endpoints, I am not able to get the pygithub apis for the same.

ref link : http://pygithub.readthedocs.io/en/latest/apis.html

PS. Previously I was using oauth app and was able to perform the flow, now since I also require to listen to events for which I needed to subscribe to webhooks, so I switched to github app.

singh811:
Would like to work on it.

xmnlab:
this feature would be very helpful :+1:

aosman-tibco:
This is covered in #1021 but is awaiting approval/merge, @sfdye can we get eyes on it?

xor2003:
To get auth_token:
`    installations_url = "https://github.hpe.com/api/v3/app/installations"
    installations_response = requests.get(installations_url, headers=headers, auth=CustomJWTAuth())
    installations_response.raise_for_status()

    for item in installations_response.json():
        installation_id = item['id']
        #print (installation_id)
        access_tokens_url = "https://github.hpe.com/api/v3/app/installations/{installation_id}/access_tokens".format(
            installation_id=installation_id)
        access_tokens_response = requests.post(access_tokens_url, headers=headers, auth=CustomJWTAuth())
        access_tokens_response.raise_for_status()
        #print (access_tokens_response.json())


        token = access_tokens_response.json()['token']
        #print (token)
        #return github3.login(token=token)
        return token

    raise Exception("No app installations")
`

NiklasRosenstein:
Check out https://pypi.org/project/github-bot-api/, it uses provides you with a PyGithub client for the GitHub app and its installations.

Issue #725, opened 2018-03-07
GithubIntegration with APP specific APIs


gfreivasc:
It would be nice to have the class `GithubIntegration` coming with calls like `get_installations()`, `get_installation(:installation_id).get_repos()`, etc. Currently it only serves as a credential fetching helper class, while it is still necessary to manually make app installation specific requests.

webknjaz:
M/b do this through CM?
```python
with GithubIntegration(...).get_installation(...) as github_client:
    github_client.get_repos()
```

webknjaz:
FTR
I've made a CLI for GitHub Checks API recently (https://pypi.org/p/check-in) and here's how I handle this:
https://github.com/webknjaz/check-in/blob/master/check_in/github_api.py#L18

PR #1135, opened 2019-06-09
GithubIntegration: enable getting installation

TomasTomecek:
I tried adding this to the Repository class but since the request requires jwt token and the special headers, it was too much of a hassle.

~~I need a bit of help with adding replay tests.~~ Figured it out, ready for review.

Related #725 #828 #1021

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -28,6 +28,7 @@
 # Copyright 2018 Wan Liuyang <tsfdye@gmail.com>                                #
 # Copyright 2018 sfdye <tsfdye@gmail.com>                                      #
 # Copyright 2018 itsbruce <it.is.bruce@gmail.com>                              #
+# Copyright 2019 Tomas Tomecek <tomas@tomecek.net>                             #
 #                                                                              #
 # This file is part of PyGithub.                                               #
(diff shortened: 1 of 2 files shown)


TomasTomecek:
can this be merged?

ericofusco:
@TomasTomecek Installation can be done for organizations too. https://developer.github.com/v3/apps/#get-an-organization-installation
Does it make sense to have a new method to do this or use the org URI if repo is not given?

TomasTomecek:
@ericofusco the code in this PR should be able to handle both

ericofusco:
@TomasTomecek Maybe I missing something, new method only calls `/repos` endpoint. Accordingly to github docs to get an installation for an organization the URI is under `/orgs` path. Method from PR only gets an installation for a single repo, not for an organization (to manage members and teams for example).

TomasTomecek:
@ericofusco oh, I misunderstood your previous comment. You/re right then.

atodorov:
@TomasTomecek I am looking into the original comment in #725 where from the integration I want to get an installation via its specific ID and for that installation get all the repositories it can access. This PR seems to be different so can you explain a bit? 

I can work on adding what I need and send a PR but I need some context of what & how was done previously.

PR #1155, opened 2019-06-27
Do not import directly into Retry tests

s-t-e-v-e-n-k:
The Retry test class was importing Repository directly, which had the
side effect of masking the tests.Repository import in AllTests with
github.Repository. Switch to importing github only, which is safe. As a
drive-by, switch to using self.assertRaises for one of the tests.

### DIFF
--- a/tests/Retry.py
+++ b/tests/Retry.py
@@ -32,8 +32,7 @@ import Framework
 
 import requests
 
-from github import GithubException
-from github.Repository import Repository
+import github
 
 REPO_NAME = 'PyGithub/PyGithub'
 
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
@sfdye This needs to land immediately, it fixes the Retry test class so we are no longer not running 95 test cases from Repository.

s-t-e-v-e-n-k:
Also, this would be a non-issue if we didn't import * from every module in AllTests, it has the side effect of pulling in everything a module imports directly into AllTests namespace.

PR #1167, opened 2019-07-12
Move a portion of github module to use six

s-t-e-v-e-n-k:
The Python 2 deadline is fast approaching, and we still rely very
heavily on Python 2 features. To make a start on moving this code base
to work without modifications for both Python 2 and 3, use the six
module for github/[A-G] for string checking.

### DIFF
--- a/github/AuthenticatedUser.py
+++ b/github/AuthenticatedUser.py
@@ -41,6 +41,7 @@
 ################################################################################
 
 import datetime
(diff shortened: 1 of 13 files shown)


s-t-e-v-e-n-k:
@sfdye I'm happy keep up with this, and at least one further change we'll need for absolute imports, what do you think?

PR #1168, opened 2019-07-12
Jump from notifications to related PRs/issues.

Ferada:
I couldn't find a way to jump directly from a notification to its pull request or its issue. This worked for me, though it's basically just copied from the existing code. If this is roughly okay I'm going to add tests for it too.


Ferada:
Added test cases, one more method since so I can generate output that I can also publish without editing out sensitive data from the notifications.

Ferada:
Same here, not stale.

Ferada:
I don't intend to create more test cases to check the parameters because it's quite tedious to simulate notifications in particular. So, that's about the last change I'm making here. If that's a quite mandatory  request I'm happy to close it instead.

s-t-e-v-e-n-k:
My concern with checking attributes is you are at least asserting the object is vaguely the one you mean, and I think you're doing enough checking. IE, you mess up, and the function returns None, and the test passes because you're not checking the object. Not that I've done that myself. :shifty_eyes:

PR #1169, opened 2019-07-12
Code review bodies are optional in some cases.

Ferada:
The body of a pull request review is optional for the `APPROVE` and `PENDING` events.


s-t-e-v-e-n-k:
Looks okay to me, requires a unit test.

Ferada:
> Looks okay to me, requires a unit test.

Working on it, don't have a Python 2 available at the moment.

Ferada:
@s-t-e-v-e-n-k so, I had to find a pull request that I could approve (can't approve your own). Looks okay like this? The test case _does_ fail if the changed code isn't present, otherwise I'll simply return 200, nothing much more to check there I think.

Ferada:
Not stale at all, a bit dry perhaps. Will fix that conflict in a bit.

Ferada:
Cheers!

PR #1171, opened 2019-07-15
Fix small issues for Python 3 compatibility.

Ferada:
Still can't run tests with Python 3, but it's improving it a bit.


s-t-e-v-e-n-k:
The print changes require from __future__ imports, but well spotted. Our Python 3 story isn't great, but I'm working on it.

s-t-e-v-e-n-k:
I think this will require a rebase when I've merged my enormous six branch. Sorry for not merging this before. :-(

Ferada:
Ah cool, so you've made all these changes now, then I can also close this.

PR #1178, opened 2019-07-22
Fix URL encoding in GithubObject:__complete

the-bumble:
PyGithub currently fails on content retrieval due to '#' characters in URL path. This code path is hit when using GitHub code search and `GithubObject.py:__complete` is retrieving the i-th repo for `decoded_content`.

I fixed it similar to other instances in the codebase using `urllib.parse.quote`.

Example of search result that cannot be retrieved:
https://github.com/nbsullivan/hashtagcluster/blob/71c978ee41dc1ccdbdba8ee93507a2e418798ece/data/results/emmybatchpreds/#Emmys_batch5080usercount.csv

Returned from code search query:
`"user:nbsullivan GIFShare_es"`


s-t-e-v-e-n-k:
I do like the direction, but this causes widespread test failures -- where is self._path defined? I'd also prefer this to contain a direct unit test that is testing the failure case seen. Further, this will require a rebase due to recent code changes landing.

the-bumble:
@s-t-e-v-e-n-k 

Thank you for the feedback. After taking more time to understand the code flow, I realize that I misplaced my fix. The encoding issue was in for the `_url` field in `ContentFile`. When this `_url` is used by the `GithubObject:__complete` method it isn't properly sanitized. It only affects `ContentFile` not all `GithubObject`s.

Here is a test case demonstrating the issue:
```
g = github.Github(ACCESS_TOKEN)
r = g.search_code("user:nbsullivan GIFShare_es")
print(r[0].decoded_content)
```

This produces the following error:
```
  File "test.py", line 7, in <module>
    print(r[0].decoded_content)
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/ContentFile.py", line 62, in decoded_content
    assert self.encoding == "base64", "unsupported encoding: %s" % self.encoding
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/ContentFile.py", line 82, in encoding
    self._completeIfNotSet(self._encoding)
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/GithubObject.py", line 265, in _completeIfNotSet
    self._completeIfNeeded()
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/GithubObject.py", line 269, in _completeIfNeeded
    self.__complete()
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/GithubObject.py", line 276, in __complete
    self._storeAndUseAttributes(headers, data)
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/GithubObject.py", line 100, in _storeAndUseAttributes
    self._useAttributes(attributes)
  File "/home/thebumble/.local/lib/python3.7/site-packages/github/ContentFile.py", line 192, in _useAttributes
    if "content" in attributes:  # pragma no branch
TypeError: argument of type 'NoneType' is not iterable
```

Did you wan to add unit test to PyGithub? If so, I would need to add a file with special characters for the test case (e.g., `#SpecialCharacters.test.txt`). If we do that, where would be the best place to put this test file?

s-t-e-v-e-n-k:
I absolutely do want a unit test added! I would add it to ContentFile's test, but don't encode # into the test name, just make it descriptive enough testEncodeWithSpecialCharacters or something close.

the-bumble:
Added a test case and the associated replay data. I hope it is close to what you had in mind. Thank you!

tests/ContentFile.py
```diff
@@ -37,7 +37,9 @@
 class ContentFile(Framework.TestCase):
     def setUp(self):
         Framework.TestCase.setUp(self)
-        self.file = self.g.get_user().get_repo("PyGithub").get_readme()
+        repo = self.g.get_user().get_repo("PyGithub")
+        self.file = repo.get_readme()
+        self.specialCharsFile = repo.get_contents("#testSpecialCharacters.txt")
```
s-t-e-v-e-n-k: You only use this in the one test case, so I would export repo as self.repo and then say specialCharFile = self.repo.get_contents in the test case.

github/ContentFile.py
```diff
@@ -209,5 +210,6 @@ def _useAttributes(self, attributes):
             self._type = self._makeStringAttribute(attributes["type"])
         if "url" in attributes:  # pragma no branch
             self._url = self._makeStringAttribute(attributes["url"])
+            self._url.value = self._url.value.replace(self._path.value, six.moves.urllib.parse.quote(self._path.value))
```
s-t-e-v-e-n-k: This to me seems like reaching under the covers -- IE, a layering violation. Are we able to quote it when it's still in attributes["url"] ?

PR #1102, opened 2019-04-25
Request API preview when asking for labels

phako:
Pass preview "Access" headers for label descriptions when requesting

Fixes the first part of #1101

Issue #1101, opened 2019-04-25
get_labels/get_label does not return description


phako:
It seems that it is missing the "Accept" : "application/vnd.github.symmetra-preview+json" header which is added for the edit call

Additionally, if you pass it and have "api_preview = True" it will be overwritten

phako:
Yes, sorry, I'm still struggling with the tests.

WilliamKluge:
Any known workarounds for this? I'm facing the same issue currently.

phako:
@WilliamKluge I proposed a fix in https://github.com/PyGithub/PyGithub/pull/1102 but failed to get the tests working afterwards

WilliamKluge:
I was able to get the test cases working with the added header for the label by modifying the replay data to also have the header. However, despite the test cases passing when I installed it into the project I found this issue in it did not resolve the problem.

PR #1188, opened 2019-08-04
Label API: pass custom media type in the Accept header

sbesson:
See #1101 for the corresponding issue and #1102 for the previous attempt at fixing it

As described in https://developer.github.com/changes/2018-02-22-label-description-search-preview/, a custom media type needs to be passed to the `Access` header to access label descriptions.

6a9fdd8 implements the new `Accept` header for getters methods of various classes
e73863a updates the unit tests accordingly


github/Milestone.py
```diff
@@ -201,7 +203,9 @@ def get_labels(self):
             github.Label.Label,
             self._requester,
             self.url + "/labels",
-            None
+            None,
+            headers={'Accept': Consts.mediaTypeLabelDescriptionSearchPreview}
+
```
s-t-e-v-e-n-k: Would you mind removing the blank line?

sbesson: :+1: pushed in 512b72c

PR #1193, opened 2019-08-06
Remove more Python version specific code

s-t-e-v-e-n-k:
Remove a bunch of other call sites that behaved differently between
Python 2 and 3, massively cleaning up a few messy methods.


s-t-e-v-e-n-k:
@sfdye This ended up being a little more wide-reaching then I intended, if you have no objections, I'll merge it in a few days.

PR #1211, opened 2019-09-01
Fix bug in repository create new file example PyGithub#1210

zume2020:
The previous code would throw an 'error path cannot start with a slash'
. Now fixed it by removing a slash in the example. PyGithub#1210


s-t-e-v-e-n-k:
Great spotting, thanks!

zume2020:
@s-t-e-v-e-n-k  you are Wlecome.

PR #1214, opened 2019-09-05
Support custom base_url for get_installation

nydr:


nydr:
Bugfix of `get_installation` which is hardcoded to use `DEFAULT_BASE_URL` which means it doesn't work with Github enterprise app installations

s-t-e-v-e-n-k:
Looks great, could you add a test or two covering the default case as well as one covering the overridden base_url case?

nydr:
If anyone is able to add test feel free to take over this PR. Unfortunately I don't have a lot of time to spare and I don't have a dev environment setup for this project so I'm unable to do so.

s-t-e-v-e-n-k:
I've finally had time to look at this. Closing this PR and replacing it with my own. Thanks for your contribution!

PR #1226, opened 2019-09-20
Fix redundant int checks

jklingen92:


s-t-e-v-e-n-k:
Ah, this was probably caused by my use of modernize, nice catch!

Issue #1246, opened 2019-10-08
Feature Request: Team Discussions


adambaratz:
The GitHub API supports interactions with [Team Discussions](https://developer.github.com/v3/teams/discussions/). It would be helpful to add a `get_discussion` method and a `get_discussions` method to `github.Team.Team` for retrieving this content.

s-t-e-v-e-n-k:
Closed via #1249

PR #1249, opened 2019-10-09
Add support for team discussions (#1246)

adambaratz:
This patch adds support for a feature in preview. This seemed reasonable since a number of other similar features have been added. I tested by writing a unit test with mock data that replicates the format of a real response. I also tested on a GitHub Enterprise instance and was able to access expected data.

I omitted implementing the ability to get a single comment, as well as create, edit, and delete. I'm happy to go back and do this, but I wanted to get feedback on what I've done so far since this is my first PR.

I omitted another feature, a preview feature inside the preview feature: reactions to discussions. The schema is totally different to the existing `github.Reaction.Reaction` class (and, IMO, kind of peculiar in comparison). I'd prefer deferring on this particular feature.

### DIFF
--- a/github/TeamDiscussion.py
+++ b/github/TeamDiscussion.py
@@ -1,0 +1,211 @@
+# -*- coding: utf-8 -*-
+
+############################ Copyrights and license ############################
+#                                                                              #
+# Copyright 2019 Adam Baratz <adam.baratz@gmail.com>                           #
+#                                                                              #
+# This file is part of PyGithub.                                               #
+# http://pygithub.readthedocs.io/                                              #
+#                                                                              #
+# PyGithub is free software: you can redistribute it and/or modify it under    #
+# the terms of the GNU Lesser General Public License as published by the Free  #
+# Software Foundation, either version 3 of the License, or (at your option)    #
+# any later version.                                                           #
+#                                                                              #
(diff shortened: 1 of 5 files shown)


github/TeamDiscussion.py
```diff
@@ -0,0 +1,211 @@
...
+        :type: datetime.datetime
+        """
+        return self._updated_at.value
+
+    @property
+    def url(self):
+        """
+        :type: string
+        """
+        return self._url.value
```
s-t-e-v-e-n-k: These are all missing self._completeIfNotSet(<attribute>)

s-t-e-v-e-n-k: self._completIfNotSet(self._url) for that one, and the same pattern for the others, since GitHub ate my markup with angle brackets. Bad GitHub, no cookie.

adambaratz: I don't think those calls are needed since `TeamDiscussion` is a child of `NonCompletableGithubObject`.

s-t-e-v-e-n-k: Ah, I missed that. However, since TeamDiscussions have a URL, I think they are completable. Can you check?

adambaratz: Good catch! I added the calls to `_completeIfNotSet` and changed the parent class.

PR #1251, opened 2019-10-09
Specify scope of flake8 to reduce runtime

adambaratz:
On my computer, it worked out to ~15 vs. ~30 seconds.


adambaratz:
This isn't a problem with actual PRs because flake8 runs before any build operations. And, I noticed, there are Python files like setup.py that are part of this project and that live outside of those two directories.

Issue #1253, opened 2019-10-09
`RecordingConnection` string handling issues


adambaratz:
I was working on test coverage for topic search (see #1252). When recording an example API response, the test runner got thrown off by a non-breaking space in the content:
```
Traceback (most recent call last):
  File "tests/Topic.py", line 67, in testSearch
    self.assertListKeyEqual(self.g.search_topics("python"), attrgetter("name"), expected_names)
  File "tests/Framework.py", line 312, in assertListKeyEqual
    realKeys = [key(element) for element in elements]
  File "github/PaginatedList.py", line 61, in __iter__
    newElements = self._grow()
  File "github/PaginatedList.py", line 73, in _grow
    newElements = self._fetchNextPage()
  File "github/PaginatedList.py", line 198, in _fetchNextPage
    headers=self.__headers
  File "github/Requester.py", line 268, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, self.__customConnection(url)))
  File "github/Requester.py", line 328, in requestJson
    return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
  File "github/Requester.py", line 381, in __requestEncode
    status, responseHeaders, output = self.__requestRaw(cnx, verb, url, requestHeaders, encoded_input)
  File "github/Requester.py", line 405, in __requestRaw
    response = cnx.getresponse()
  File "tests/Framework.py", line 129, in getresponse
    self.__writeLine(output)
  File "tests/Framework.py", line 140, in __writeLine
    self.__file.write(line + '\n')
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 2437: ordinal not in range(128)
```

I haven't looked too closely at the code, but this looks related to the fact that this class treats the response body as ASCII and UTF-8 in different places.

There's another issue with the same line in Python 3. Files opened in binary mode expect to receive bytes and `'\n'` is a string.

s-t-e-v-e-n-k:
Closed via #1254

PR #1254, opened 2019-10-10
Handle unicode strings when recording responses (#1253)

adambaratz:
The string handled had some muddled assumptions about binary vs. string vs. unicode data. I updated to assume unicode at all times and added test coverage. I also tried running a test with the `--record` function to validate in practice.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,6 +110,8 @@ if __name__ == "__main__":
         },
         tests_require=[
             "cryptography",
-            "httpretty==0.9.6"
+            "httpretty==0.9.6",
+            "mock==3.0.5",
+            "parameterized==0.7.0",
         ]
     )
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -41,11 +41,23 @@
...
 version = "1.43.8"
 
+tests_require = [
+    "cryptography",
+    "httpretty==0.9.6",
+    "parameterized==0.7.0",
+]
+
+if sys.version_info < (3, 3):
+    tests_require.append("mock==3.0.5")
```
s-t-e-v-e-n-k: Part of me does not like this duplication. However I just merged a change changing the httpretty requirement. Could you change it to >=? And I note I didn't fix test-requirements, sigh.

s-t-e-v-e-n-k:
My only remaining concern is whether or not the new tests will actually make HTTP requests to GitHub itself, but that can addressed (if required) in a future patch. Thanks for the amazing work!

PR #1269, opened 2019-10-22
Skip Issue142 test if no connection to GitHub

s-t-e-v-e-n-k:
The Issue142 test requires actual GitHub due to type checking, but some
test runners are forbidden from network access. Check for this and skip
the test if that is the case, so they do not have to carry patches.

### DIFF
--- a/tests/Issue142.py
+++ b/tests/Issue142.py
@@ -26,11 +26,23 @@
 ################################################################################
 
 from __future__ import absolute_import
+
+import socket
 import unittest
 import github
 
 
+def is_github_reachable():
+    try:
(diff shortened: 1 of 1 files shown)


adambaratz:
I don't think the type issue is present anymore. I recorded a response for this test and it ran fine (#1271). I don't know when this was fixed. I didn't go do a whole git bisect run, but it was even working before my recent PR that moved the replay framework from binary to string data (#1254).

s-t-e-v-e-n-k:
Solved better with https://github.com/PyGithub/PyGithub/pull/1271.

PR #1275, opened 2019-10-26
Create repositories

AndersTornkvist:
I wanted to create user repositories with the package. I added a method in `Github` class in MainClass.py that sends a POST request to /user/repos (following the spec http://developer.github.com/v3/repos#create) and returns a new `Repository` instance based on the response from the server. Added two tests for creating a public and a private repository for the user.


github/MainClass.py
```diff
@@ -690,6 +692,62 @@ def get_installation(self, id):
...
+        post_parameters = {
+            'name': name,
+        }
+
+        for key, value in (
+                ('description', description), ('homepage', homepage), ('private', private), ('has_issues', has_issues),
+                ('has_projects', has_projects), ('has_wiki', has_wiki), ('is_template', is_template),
+                ('auto_init', auto_init), ('gitignore_template', gitignore_template), ('license_template', license_template),
+                ('allow_squash_merge', allow_squash_merge), ('allow_merge_commit', allow_merge_commit), ('allow_rebase_merge', allow_rebase_merge),
+        ):
```
s-t-e-v-e-n-k: I'm struggling to think of a cleaner way to do this, but the duplication! I'm almost tempted to suggest:

```
for arg in ('description', 'homepage' ...):
    if eval(arg) is not github.GithubObjectNotSet:
       post_parameters[key] = eval(arg)
```

github/MainClass.py
```diff
@@ -690,6 +692,62 @@ def get_installation(self, id):
         """
         return Installation.Installation(self.__requester, headers={}, attributes={"id": id}, completed=True)
 
+    def create_repo(self, name, description=github.GithubObject.NotSet, homepage=github.GithubObject.NotSet, private=github.GithubObject.NotSet, has_issues=github.GithubObject.NotSet, has_projects=github.GithubObject.NotSet, has_wiki=github.GithubObject.NotSet, is_template=github.GithubObject.NotSet, auto_init=github.GithubObject.NotSet, gitignore_template=github.GithubObject.NotSet, license_template=github.GithubObject.NotSet, allow_squash_merge=github.GithubObject.NotSet, allow_merge_commit=github.GithubObject.NotSet, allow_rebase_merge=github.GithubObject.NotSet):
+        """
+        :calls: `POST /user/repos <http://developer.github.com/v3/repos>`_
+        :param name: string
```
s-t-e-v-e-n-k: Are you able to create repos for any random user, or only yourself? If only for yourself then AuthenticatedUser is the correct place for this method.

AndersTornkvist:
Thank you so much for your time to review the pull request! I agree that `AuthenticatedUser` is a good place to have the method. I just checked the class and to my surprise I found a similar method in `AuthenticatedUser`. I am ashamed to not have found this when I searched for it before creating the PR. My PR is redundant and I'll close it. Thank you for your time!

PR #1282, opened 2019-11-07
add page limit functionality for getting pull requests

tlong123:
This PR will help reduce API calls to github by allowing a user to only retrieve a specific number of pages of pull requests


s-t-e-v-e-n-k:
I like the direction, but it does need at least one test.

tlong123:
@s-t-e-v-e-n-k I've added a test!

s-t-e-v-e-n-k:
I think that's a good test for PaginatedList, but you're changing far more now than what you were previously. Also, you need to check what happens when you pass page_limit=NotSet, and you're not asserting in the methods that page_limit is either NotSet or an int. But, keep up the good work!

Issue #422, opened 2016-07-18
Timeline missing


hryamzik:
There's a [timeline](https://developer.github.com/v3/issues/timeline/) object in github API, but it's not presented in this library. It provides more information than events. At least it seem to be the only way to get `unsubscribed` events.

malex984:
:+1:

mawkee:
Wasn't this closed?

sfdye:
Closed in #1302

PR #1302, opened 2019-11-26
Support github timeline events (GH-422).

ncb000gt:
Adds support for retrieval of Github timeline events (GH-422).

Usage example:

```python
issue = repo.get_issue(28)
events = issue.get_timeline()
for event in events:
    print(f'Event: {event}')
```

The class doesn't match the output exactly as shown on the api site because the output from the call/events doesn't. This can be seen in the text file used for testing. This can be verified using the following.

```bash
curl -i https://[REPLACE WITH GH TOKEN]@api.github.com/repositories/3544490/issues/28/timeline -H"Accept:application/vnd.github.mockingbird-preview"
```

Signed-off-by: Nick Campbell <nicholas.j.campbell@gmail.com>


s-t-e-v-e-n-k:
This is looking great! Would you mind adding a test case for TimelineEvent where you fetch one from the API and assert it's parsed correctly? tests/RequiredStatusChecks.py is a small example.

ncb000gt:
@s-t-e-v-e-n-k I'll take a look and post back. Thanks.

ncb000gt:
@s-t-e-v-e-n-k I took a look and the difference between individual timeline events is that they can't be lazily loaded. You either get the full timeline or you don't. I removed the "load if not found" calls since they can't be individually loaded. I added some additional checks to the test that I did add.

What are your thoughts?

s-t-e-v-e-n-k:
I'm happy with the replay data including the entire timeline even if you only assert the attributes of the first one. You can always selectively edit the replay data to have more brevity.

tests/Issue.py
```diff
@@ -174,3 +175,29 @@ def testCreateReaction(self):
 
         self.assertEqual(reaction.id, 16917472)
         self.assertEqual(reaction.content, "hooray")
+
+    def testGetTimeline(self):
+        expected_events = {'referenced', 'cross-referenced', 'locked', 'unlocked', 'closed', 'assigned', 'commented', 'subscribed', 'labeled'}
+        events = self.issue.get_timeline()
+        for event in events:
+            print('EVENT: ', event)
```
s-t-e-v-e-n-k: Useful for debugging, not so useful for test runs. :-)

ncb000gt: Hah. Whoops. Thanks.

ncb000gt:
@s-t-e-v-e-n-k let me know if you want me to make any additional changes. Thanks!

github/TimelineEventSource.py
```diff
@@ -0,0 +1,63 @@
...
+#                                                                              #
+################################################################################
+
+from __future__ import absolute_import
+import github.GithubObject
+
+import github.Issue
+
+
+class TimelineEventSource(github.GithubObject.CompletableGithubObject):
```
s-t-e-v-e-n-k: This class is not completable, it does not have a URL property.

tests/Issue.py
```diff
@@ -263,3 +264,39 @@ def testCreateReaction(self):
...
+
+            if event.event == "cross-referenced":
+                # cross-referenced events don't include an event id or node_id
+                self.assertIsNotNone(event.source)
+                self.assertEqual(event.source.type, 'issue')
+            else:
+                self.assertIsNotNone(event.id)
+                self.assertIsNotNone(event.node_id)
+
+                if event.event == 'referenced':
```
s-t-e-v-e-n-k: Are there any referenced events in the timeline? Coverage is saying the next two lines don't fire.

ncb000gt: There wasn't. I removed the check here. Thanks.

github/TimelineEventSource.py
ncb000gt: I changed it to `NonCompletableGithubObject`. Thanks.

github/TimelineEvent.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        return self._commit_url.value
+
+    @property
+    def source(self):
+        """
+        :type: :class:`github.TimelineEventSource.TimelineEventSource`
+        """
+        # only available on `cross-referenced` events.
+        if self.event == 'cross-referenced' and hasattr(self, '_source') and self._source is not None:
```
s-t-e-v-e-n-k: Actually, this and the below checks are also wrong, because they'll be NotSet, not None. The _hasattr() feels wrong too, surely they're set in initAttributes ...

github/TimelineEvent.py
```diff
@@ -0,0 +1,159 @@
...
+################################################################################
+
+from __future__ import absolute_import
+import github.GithubObject
+
+import github.NamedUser
+import github.TimelineEventSource
+
+
+class TimelineEvent(github.GithubObject.NonCompletableGithubObject):
```
s-t-e-v-e-n-k: It is completable! It has a URL and everything.

ncb000gt: They don't all have a url. Some do. if you look through the replay data the `cross-referenced` event does not have a url.

github/TimelineEvent.py
ncb000gt: Hmm. At some point I was having issues with that, but it was likely early on in the changes I was making. I think it was that I wasn't setting the properties in `initAttributes`. Thanks.

PR #1306, opened 2019-11-26
Support for create_fork

### LINKED ISSUE #1305: Support for create_fork in http://developer.github.com/v3/repos/forks
According to http://developer.github.com/v3/repos/forks

You can create a fork via a POST to the repositories /forks URL however this is not supported via PyGithub yet.

timgates42:
http://developer.github.com/v3/repos/forks

Closes #1305


timgates42:
Ok will need to figure out how to add some unit tests - will do

github/Repository.py
```diff
@@ -1867,6 +1867,22 @@ def get_forks(self):
             None
         )
 
+    def create_fork(self, organization=None):
+        """
+        :calls: `POST /repos/:owner/:repo/forks <http://developer.github.com/v3/repos/forks>`_
```
s-t-e-v-e-n-k: Please document organiztion in the docstring.

Unset parameters must default to NotSet (see elsewhere in that file)

timgates42: thanks will do

github/Repository.py
```diff
@@ -1867,6 +1867,22 @@ def get_forks(self):
...
+        :rtype: :class:`github.Repository.Repository`
+        """
+        post_parameters = {
+        }
```
s-t-e-v-e-n-k: Also, this can be one line.

timgates42:
Thanks for your help and patience @s-t-e-v-e-n-k let me know if you spot anything else that needs to be fixed.

github/Repository.py
s-t-e-v-e-n-k: Just thinking about it more -- is it really organization, or is it destination of the fork?

github/Repository.py
```diff
@@ -1867,6 +1867,22 @@ def get_forks(self):
             None
         )
 
+    def create_fork(self, organization=github.GithubObject.NotSet):
+        """
+        :calls: `POST /repos/:owner/:repo/forks <http://developer.github.com/v3/repos/forks>`_
+        :param organization: string or "none" or "*"
+        :rtype: :class:`github.Repository.Repository`
+        """
+        post_parameters = {}
```
s-t-e-v-e-n-k: We also assert arguments are correct, sorry I failed to pick that up before.

github/Repository.py
timgates42: I just got the name from the GitHub API docs
image

I think GitHub often refers to a collection of repositories as an organization even though sometimes this is a single user.

timgates42: I guess if not specified then it is your user account otherwise you are forking into an organization - as presumably you can't fork into someone else user account.

github/Repository.py
timgates42: thanks will do

timgates42: added now

s-t-e-v-e-n-k:
@timgates42 Thanks for your great first contribution!

timgates42:
Thanks @s-t-e-v-e-n-k !

PR #1307, opened 2019-11-29
Fix class used returning pull request comments

lamflam:
When trying to pull in pull request comments I noticed that it was returning `IssueComment`s instead of `PullRequestComment`s so this just switched to use the proper type.


s-t-e-v-e-n-k:
Perhaps we need something to check the documented return type, but nice catch, thank you!

PR #1330, opened 2019-12-29
Fix PyPI wheel deployment

### LINKED ISSUE #1321: Publish sdist and bdist wheel
The benefits of wheels are well documented. See: https://pythonwheels.com/
This package is pure Python and publishing it as both source and as a wheel is simple.

It looks like there is config for Travis to upload wheels, but it doesn't appear to be happening for some reason. Is this something that we can fix or you'd like assistance in fixing?

sharkykh:
`distributions` should be under `deploy`, not `deploy.on`
https://docs.travis-ci.com/user/deployment/pypi/#uploading-different-distributions

Fixes #1321

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,11 @@ notifications:
 deploy:
   provider: pypi
   skip_existing: true
+  distributions: sdist bdist_wheel
   user: jacquev6
   password:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,11 +19,11 @@ notifications:
 deploy:
   provider: pypi
   skip_existing: true
+  distributions: sdist bdist_wheel
   user: jacquev6
   password:
     secure: eorsWtbj7JUBcy/Ovohg2yxyvyF4Sz9QNqtdPJhLSBzd1S01qoVLVAqsQhfZTxMlSOE4RCPOm+VodhV55oa1RboLah4DpDDW998J61QSo9im7Ch2GBkL3C6XqhWAFr6g4KqTG4h/6QTp5dF8vebXiMADmpshCMMpUxm3FccFY7k=
   on:
     tags: true
-    distributions: "sdist bdist_wheel"
```
sharkykh: I also don't believe the quotes are needed.

s-t-e-v-e-n-k:
Nice catch, thanks!

PR #1338, opened 2020-01-06
add endpoint for "transfer repo"

droopy4096:
To transfer repo from one owner to another one needs to call /repos/:owner/:repo/transfer as per https://developer.github.com/v3/repos/#transfer-a-repository . This PR aims to add that endpoint to API.

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -2089,6 +2089,21 @@ class Repository(github.GithubObject.CompletableGithubObject):
         )
         return Repository(self._requester, headers, data, completed=True)
 
+    def transfer(self, new_owner, team_ids=[]):
+        """
+        :calls: `POST /repos/:owner/:repo/transfer <https://developer.github.com/v3/repos/#transfer-a-repository>`_
+        :param new_owner: string or "none" or "*"
+        :rtype: :class:`github.Repository.Repository`
+        """
+        post_parameters = {}
+        if new_owner is not None:
+            post_parameters["new_owner"] = new_owner
+        post_parameters["team_ids"] = team_ids
+        headers, data = self._requester.requestJsonAndCheck(
+            "POST", self.url + "/transfer", input=post_parameters,
+        )
+        return Repository(self._requester, headers, data, completed=True)
+
     def get_git_blob(self, sha):
         """
         :calls: `GET /repos/:owner/:repo/git/blobs/:sha <http://developer.github.com/v3/git/blobs>`_
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -2097,7 +2097,8 @@ class Repository(github.GithubObject.CompletableGithubObject):
         :rtype: :class:`github.Repository.Repository`
         """
         assert new_owner is not None
-        assert isinstance(team_ids, list)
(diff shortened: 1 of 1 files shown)


github/Repository.py
```diff
@@ -2089,6 +2089,22 @@ def create_fork(self, organization=github.GithubObject.NotSet):
...
 
+    def transfer(self, new_owner, team_ids=[]):
+        """
+        :calls: `POST /repos/:owner/:repo/transfer <https://developer.github.com/v3/repos/#transfer-a-repository>`_
+        :param new_owner: string or "none" or "*". Required.
+        :param team_ids: list. Optional. Works only under organizations.
+        :rtype: :class:`github.Repository.Repository`
+        """
+        post_parameters = {}
+        if new_owner is not None:
```
s-t-e-v-e-n-k: We require this argument, so this check is pointless. Please add asserts for both arguments.

github/Repository.py
```diff
@@ -2089,6 +2089,22 @@ def create_fork(self, organization=github.GithubObject.NotSet):
...
+            post_parameters["new_owner"] = new_owner
```
s-t-e-v-e-n-k: And since it's required, we can initialise post_parameters with it.

github/Repository.py
droopy4096: not sure how to proceed with unit test for "transfer" as it requires GitHub permissions to 2 organizations etc. so not-so-trivial to have set up. Any pointers?

github/Repository.py
```diff
@@ -2089,6 +2089,21 @@ def create_fork(self, organization=github.GithubObject.NotSet):
...
         return Repository(self._requester, headers, data, completed=True)
 
+    def transfer(self, new_owner, team_ids=[]):
+        """
+        :calls: `POST /repos/:owner/:repo/transfer <https://developer.github.com/v3/repos/#transfer-a-repository>`_
+        :param new_owner: string or "none" or "*". Required.
+        :param team_ids: list. Optional. Works only under organizations.
+        :rtype: :class:`github.Repository.Repository`
+        """
+        assert new_owner is not None
```
s-t-e-v-e-n-k: ``assert isinstance(new_owner, str), new_owner``

Since other objects will also also match the current assert.

github/Repository.py
```diff
@@ -2089,6 +2089,21 @@ def create_fork(self, organization=github.GithubObject.NotSet):
...
+        assert isinstance(team_ids, list)
```
s-t-e-v-e-n-k: A list of ints? At this point, you're only asserting it's a list.

``assert all(inistance(element, int) for element in team_ids), team_ids``

s-t-e-v-e-n-k:
Furthermore, my apologies for taking so long to get back to this review -- LCA 2020 is incredibly distracting. :-)

github/Repository.py
s-t-e-v-e-n-k: This isn't required since you assert it's a str just below.

s-t-e-v-e-n-k:
@droopy4096 Hi, do you time to work on a test for this?

Issue #1335, opened 2020-01-02
Dismissal Restrictions Enabled When no Users/Teams Provided


AetherDeity:
Branch protection for organization repos allow for the inclusion of dismissal restrictions for who should be able to dismiss reviews. To disable this feature, you must pass an empty object when editing branch protection. The current function `edit_protection` in `Branch.py` excludes the field if no users or teams are provided, which is the same as setting both to an empty list. This has the effect of limiting who can dismiss a review to admin users.

Unfortunately, the fix is not as simple as including the field with an empty dict when neither users nor teams are provided, as this causes a validation error when used to edit branch protection for user owned repos.

```
422 {'message': 'Validation Failed', 'errors': ['Only organization repositories can have users and team restrictions'], 'documentation_url': 'https://developer.github.com/v3/repos/branches/#update-branch-protection'} 
```

I am happy to PR a fix, but am unsure what direction the maintainers would like to go.

1. Add org context to the Branch object when calling `get_branch` from a Repository object. This has the benefit of not introducing an API breaking change or additional arguments to `edit_protection`.

2. Add an option to the `edit_protection` function that allows the user to specify whether it is a branch or an org repo. This will require a default that will either need to be defaulted to true or false. Default of user repo would maintain current behavior and leave org repos misconfigured. Default to org repo would cause validation errors for user repos, thus introducing an API breaking change.

3. Something else.

s-t-e-v-e-n-k:
I'm still unhappy with parts of the branch protection API, but mostly at how brittle it is exported by GitHub.

Could we look at the Branch object and if it has users and teams dismissal restrictions set, then pass those if NotSet, otherwise ignore them, would that work?

AetherDeity:
The Branch object itself wouldn't have any branch protection attributes to check; that data is retrieved by the `get_protection` call, which returns a BranchProtection object. Given just the Branch object, there is currently no way that I know of to determine whether the call should or should not include the dismissal restrictions.

s-t-e-v-e-n-k:
Some of the attributes are populated when calling get_branch():

```
>>> repo = g.get_repo('PyGithub/PyGithub')
>>> b = repo.get_branch('master')
>>> b._rawData['protection']
{'enabled': True, 'required_status_checks': {'enforcement_level': 'off', 'contexts': []}}
```

AetherDeity:
[...]

PR #1339, opened 2020-01-07
Fix org dismissal restrictions

AetherDeity:
Addresses https://github.com/PyGithub/PyGithub/issues/1335

### DIFF
--- a/github/Branch.py
+++ b/github/Branch.py
@@ -72,6 +72,13 @@ class Branch(github.GithubObject.NonCompletableGithubObject):
         return self._protected.value
 
     @property
+    def protection(self):
+        """
+        :type: dict
+        """
+        return self._protection
+
+    @property
     def protection_url(self):
         """
         :type: string
@@ -83,6 +90,7 @@ class Branch(github.GithubObject.NonCompletableGithubObject):
         self._name = github.GithubObject.NotSet
         self._protection_url = github.GithubObject.NotSet
         self._protected = github.GithubObject.NotSet
+        self._protection = github.GithubObject.NotSet
 
     def _useAttributes(self, attributes):
         if "commit" in attributes:  # pragma no branch
@@ -97,6 +105,8 @@ class Branch(github.GithubObject.NonCompletableGithubObject):
             )
         if "protected" in attributes:  # pragma no branch
             self._protected = self._makeBoolAttribute(attributes["protected"])
+        if "protection" in attributes:  # pragma no branch
+            self._protection = self._makeDictAttribute(attributes["protection"])
 
     def get_protection(self):
         """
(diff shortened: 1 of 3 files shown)


github/Repository.py
```diff
@@ -1518,6 +1518,7 @@ def get_branch(self, branch):
         headers, data = self._requester.requestJsonAndCheck(
             "GET", self.url + "/branches/" + branch
         )
+        data['is_organization'] = self.organization is not None
```
AetherDeity: This change causes Repository object to be loaded in tests, negating lazy load https://github.com/PyGithub/PyGithub/blob/dd158556a838a180d966925c824100f320a5fcfc/tests/Branch.py#L45

I'm not sure if the test data needs rerecording or if there is a different fix for the resulting failed tests.

s-t-e-v-e-n-k: Yeah, I think you might want to duck under the covers there and use the underscored attribute.

github/Branch.py
```diff
@@ -196,6 +206,10 @@ def edit_protection(
             or required_approving_review_count is not github.GithubObject.NotSet
         ):
             post_parameters["required_pull_request_reviews"] = {}
+            if self._rawData.get("is_organization", False):
```
s-t-e-v-e-n-k: Please add a property to Branch that exposes it and check it here.

tests/GitRelease.py
```diff
@@ -51,9 +51,8 @@ def setUp(self):
         with open(self.content_path, "w") as zip_content:
             zip_content.write("Pedro for president.")
 
-        artifact = zipfile.ZipFile(self.artifact_path, "w")
-        artifact.write(self.content_path)
-        artifact.close()
+        with zipfile.ZipFile(self.artifact_path, "w") as artifact:
+            artifact.write(self.content_path)
```
s-t-e-v-e-n-k: This is a nice change, but can you split it out to a separate PR?

github/Branch.py
```diff
@@ -71,6 +71,13 @@ def protected(self):
         """
         return self._protected.value
 
+    @property
```
s-t-e-v-e-n-k: I mean, I get why you're adding it, but we also don't check it anywhere at all ... Not even tests.

github/Repository.py
AetherDeity: I did look at doing that, but as I understand the lazy loading, the organization key would not exist before completing the object. The result would be a lazy loaded Repository would always look like a user repo.

github/Branch.py
AetherDeity: I wasnt sure if it ought to be exposed as it isn't a native GitHub attribute, but I can certainly do that.

github/Branch.py
AetherDeity: As an attribute returned by GitHub, I thought it made sense to expose. I'm not sure if you are asking for this to be removed or for tests to check it.

tests/GitRelease.py
AetherDeity: Can do

PR #1342, opened 2020-01-08
Add get_review_comments to PullRequestReview

urishalit:
I tried adding a test for this, and frankly understood you guys will do it much faster than me... It seems quite simple if you know exactly what to put in the txt file.

Would appreciate if you can merge this and publish a new version, thanks.

### DIFF
--- a/github/PullRequestReview.py
+++ b/github/PullRequestReview.py
@@ -123,6 +123,18 @@ class PullRequestReview(github.GithubObject.CompletableGithubObject):
             input=post_parameters,
         )
 
+    def get_review_comments(self):
+        """
+        :calls: `GET /repos/:owner/:repo/pulls/:pull_number/reviews/:review_id/comments <https://developer.github.com/v3/pulls/reviews/#get-comments-for-a-single-review>`_
+        """
+        url_parameters = dict()
+        return github.PaginatedList.PaginatedList(
+            github.PullRequestComment.PullRequestComment,
+            self._requester,
+            self.pull_request_url + "/reviews/" + str(self.id) + "/comments",
+            url_parameters,
+        )
+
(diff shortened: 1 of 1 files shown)


urishalit:
@TravisBuddy Can someone help me with the test?

s-t-e-v-e-n-k:
This API endpoint is already exported as ``PullRequest.get_single_review_comments(id)``. Which also isn't tested (sigh!)

urishalit:
> This API endpoint is already exported as `PullRequest.get_single_review_comments(id)`. Which also isn't tested (sigh!)

Could be, but that isn't the straightforward way to do it. You have a review, you want to get the comments for it. Why not add this as well?

s-t-e-v-e-n-k:
It calls exactly the same API endpoint as your PR, and is already landed. To replicate your functionality just call ''pr.get_single_review_comments(pr.id)`` ?

s-t-e-v-e-n-k:
To put it another way, why would I merge code that looks 99% identical to mean you don't have to pass one argument?

urishalit:
What you are saying is this:

```
pull = repo.get_pull(123)
for review in pull.get_reviews():
     pull. get_single_review_comments(review.id)
```

While what I am suggesting is this:

```
pull = repo.get_pull(123)
for review in pull.get_reviews():
     review.get_review_comments()
```

I think my solution is more correct for a few reasons:

1. The comments are properties of the review not the pull, so it should be reflected as such.
2. If for instance I want to call function/store reviews in state - in your way I need to also store the pull with it (as from a review you can't know for which pull it is) - this is quite cumbersome.

But in the end, 1 is the reason - the objects should reflect the data model, the api is to get comments of a review not a pull thus the function should be on the review and not the pull

PR #1374, opened 2020-02-02
Add support to reply to a Pull request comment

### LINKED ISSUE #1375: Add support to reply to a comment in a PullRequest
https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply

rsn491:
Closes #1375

Adds support to reply to a Pull request comment:
-> https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply


github/PullRequest.py
```diff
@@ -402,6 +402,24 @@ def create_review_comment(self, body, commit_id, path, position):
             self._requester, headers, data, completed=True
         )
 
+    def create_review_comment_reply(self, comment_id, body):
+        """
+        :calls: `POST /repos/:owner/:repo/pulls/:pull_number/comments/:comment_id/replies <http://developer.github.com/v3/pulls/comments>`_
+        :param body: string
```
s-t-e-v-e-n-k: Please also document comment_id

rsn491: added missing docs

rsn491:
> Reflecting on this further, I'm wondering if we can make this API a little easier to use -- where do you fetch the comment ID from that's passed into this function?

The comment id can be the id of any comment in a reply sequence. Regardless of the id provided, the created comment will have the `in_reply_to_id` attr set to the top comment.

The API could potentially be simplified if we add this functionality as a `create_reply` method to the `PullRequestComment` class. Unfortunately, we would need to parse the pull request URL, since the Github API does not provide the PullRequest number when getting a PullRequest comment. 
What do you think? 

Note: Nevertheless, I can update the tests to show that if you reply to both the top comment or another comment in a reply chain, the code can always create a comment, and the created comment will have the `in_reply_to_id` attr pointing to the top comment.

s-t-e-v-e-n-k:
Okay, it seems a little awkward to me, but having said that, I think the awkward is coming from GitHub's implementation, and not your work.

rsn491:
After looking more closely at the `PullRequestComment`, I noticed that the `self.url`in the class corresponds to `/repos/:owner/:repo/pulls/comments/:comment_id`, which is a different base url than the reply API one `/repos/:owner/:repo/pulls/:pull_number/comments/:comment_id/replies` that requires the `:pull_number` in between. As so, if I implement this functionality in this class, I'm kind of breaking the abstraction and the convention followed by the framework so far.
Being said, probably better to keep it as it is, i.e., leave the reply functionality only in the `PullRequest` class, agree?

rsn491:
@s-t-e-v-e-n-k avoided parsing by using the `pull_request_url` property

s-t-e-v-e-n-k:
I'd rather not duplicate the API calls -- the two functions look almost identical. Pick the one you prefer and delete the other. Further more, you could change the first parameter to also take a PullRequestComment and grab the id from it rather than requiring an int. But thank you for the amazing work so far!

rsn491:
Sounds good... In that case, I'd rather keep the first API since it is more aligned with the Github API, and also keep accepting the comment `id` as a parameter for flexibility.

s-t-e-v-e-n-k:
Sorry I've dropped the ball on this. Can you either rebase on top of master, or merge master into this branch and I'll land it?

rsn491:
No worries... Sure! just rebased on`master`

s-t-e-v-e-n-k:
Bad me, bad! No cookie! Rebase again and tag me in a comment and I'll hit the button.

rsn491:
@s-t-e-v-e-n-k no worries... rebased!

PR #1379, opened 2020-02-04
Search repo issues by string label

### LINKED ISSUE #1377: Search repo issue by label without `get_label`
In order to search a repo's issues by label, you first need to fetch the label, then perform the search. This means we're doing an extra HTTP request (and using an extra API usage) just to search issues.
I am working on integration with a repo where I know the label exists. Either way, this is irrelevant as GitHub will return no results for a label that doesn't exist.

What I propose is to allow searching issues by the label strings removing the need for an extra API request. This means that PyGithub has the same behaviour for handling labels in `Repository.get_issues` as `Repository.create_issue`

huwcbjones:
Fixes #1377 

### The problem
Allows searching a repository for issues by label without having to do an HTTP request to fetch the label.

### The solution
1. Add a check to the assert to allow either `Label` or `str` types
2. When building URL params, fetch the name of the Label (if Label type), else the label

Note: code shamelessly copied from `Repository.create_issue`.

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -1120,7 +1120,7 @@ class Repository(github.GithubObject.CompletableGithubObject):
         :param title: string
         :param body: string
         :param assignee: string or :class:`github.NamedUser.NamedUser`
-        :param assignees: list (of string or :class:`github.NamedUser.NamedUser`)
+        :param assignees: list of string or :class:`github.NamedUser.NamedUser`
         :param milestone: :class:`github.Milestone.Milestone`
         :param labels: list of :class:`github.Label.Label`
         :rtype: :class:`github.Issue.Issue`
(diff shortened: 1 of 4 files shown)


github/Repository.py
```diff
@@ -2192,7 +2192,7 @@ def get_issues(
         :param state: string. `open`, `closed`, or `all`. If this is not set the GitHub API default behavior will be used. At the moment this is to return only open issues. This might change anytime on GitHub API side and it could be clever to explicitly specify the state value.
         :param assignee: string or :class:`github.NamedUser.NamedUser` or "none" or "*"
         :param mentioned: :class:`github.NamedUser.NamedUser`
-        :param labels: list of :class:`github.Label.Label`
+        :param labels: list (of string or :class:`github.Label.Label`)
```
s-t-e-v-e-n-k: I don't like the brackets, I think just remove those and say strings, and this is fine to land.

huwcbjones: Have removed brackets 👍

PR #1381, opened 2020-02-06
Add repo edit support for delete_branch_on_merge

testworksau:
A new parameter has been added to the repositories api, which allows the setting of the "Automatically delete head branches" setting shown in the screenshot below.

The parameter is defined here (create):
https://developer.github.com/v3/repos/#parameters-4

and here (edit):
https://developer.github.com/v3/repos/#input

image

Incidentally / FYI, the script to add missing attributes bombed out on my Windows environment (Python x64 3.7.5).

https://github.com/PyGithub/PyGithub/blob/master/CONTRIBUTING.md#adding-missing-attributes-for-a-githubobject

```
python scripts/add_attribute.py AuthenticatedUser delete_branch_on_merge bool                                                                                                                        Traceback (most recent call last):
  File "scripts/add_attribute.py", line 141, in <module>
    f.write(line + "\n")
TypeError: a bytes-like object is required, not 'str'
```

The end result was a saved, empty file.


tests/Repository.py
```diff
@@ -126,6 +126,7 @@ def testEditWithAllArguments(self):
             allow_squash_merge=True,
             allow_merge_commit=True,
             allow_rebase_merge=True,
+            delete_branch_on_merge=True,
```
s-t-e-v-e-n-k: My only comment here would be we don't actually assert on any of the changed parameters. To help increase coverage, can you ``assertTrue delete_branch_on_merge`` and add a FIXME comment to assert the rest?

testworksau: Done. 

Ironically, adding an assertion for this parameter found a recursion bug due to a copy / paste error :)

s-t-e-v-e-n-k: Untested code is broken code. :-)

PR #1388, opened 2020-02-10
Add tox

sfdye:
- [x] Migrate all virtualenvs to tox 
- [x] Support for python3.5 - Python 3.8
- [x] Add travis check for flake8/black/isort
- [x] Add a tox command `format` for manual formatting


tox.ini
```diff
@@ -0,0 +1,17 @@
+[tox]
+envlist = py{36,37,38},flake8
+
+[testenv]
+deps = -rtest-requirements.txt
+commands =
+	python setup.py test
```
s-t-e-v-e-n-k: This is the deprecated command that tox is supposed to be getting us away from. :-)

tox.ini
```diff
@@ -0,0 +1,17 @@
+[tox]
+envlist = py{36,37,38},flake8
```
s-t-e-v-e-n-k: This is missing py35.

.travis.yml
```diff
@@ -1,14 +1,19 @@
...
+  include:
+    - python: 3.5
+      env: TOXENV=py35
+    - python: 3.6
+      env: TOXENV=py36
+    - python: 3.7
+      env: TOXENV=py37
+    - python: 3.8
+      env: TOXENV=py38
+    - env: TOXENV=flake8
```
s-t-e-v-e-n-k: This to be reads like we run flake8 under 3.8, but tox.ini says 3.6. Can we pick one?

.travis.yml
```diff
@@ -1,14 +1,19 @@
...
 install:
-  - pip install codecov flake8
+  - pip install tox
 script:
-  - flake8
-  - coverage run --source github,tests setup.py test
```
s-t-e-v-e-n-k: Does this mean we lose coverage reports?

sfdye: Thanks for the feedback. Again this PR is specifically in Draft mode, which means it's not yet ready for review

s-t-e-v-e-n-k: It's never early for reviews! :-)

sfdye: The draft mode was mean to not disturb the reviewers 😬

tox.ini
sfdye: we might need to migrate to pytest first, but that will be a separate PR

.flake8
```diff
@@ -1,4 +0,0 @@
-[flake8]
```
sfdye: this is moved to tox.ini

sfdye:
@s-t-e-v-e-n-k Now it's ready for review :)

test-requirements.txt
```diff
@@ -1,3 +1,7 @@
 cryptography
 httpretty>=0.9.6
 parameterized>=0.7.0
+coverage==5.0.3
+black==19.10b0; python_version > '3.5'
```
s-t-e-v-e-n-k: I don't think we need the marker here, we require 3.5 or greater anyway

.travis.yml
```diff
@@ -1,14 +1,19 @@
...
+  include:
+    - python: 3.5
+      env: TOXENV=py35
+    - python: 3.6
+      env: TOXENV=py36
+    - python: 3.7
+      env: TOXENV=py37
+    - python: 3.8
+      env: TOXENV=py38
+    - env: TOXENV=lint
```
s-t-e-v-e-n-k: I remain concerned about which Python version Travis will use when tox is configured for 3.6, but if it works, then that's great too.

sfdye: It works. I set the base python to py36 for lint.

test-requirements.txt
sfdye: This is because black is only available for py3.6+

sfdye:
image

Now travis will fail either flake8/isort/black fails the check.

PR #1408, opened 2020-02-22
Adding media preview headers

UserXGnu:
Adding mandatory media preview header to get_label() get_labels() methods within Repository and Issue classes


s-t-e-v-e-n-k:
You also need to change the replay data since the headers sent will also include the new one you've added. Do you need help on how to do that?

Issue #1415, opened 2020-03-03
Team missing "get_repo_permissions"


tl-adrian-bridgett:
There is a "set_repo_permissions" call (great!) however there's no way to read the permissions.  GitHub does support this:

https://developer.github.com/v3/teams/#check-if-a-team-manages-a-repository

I'm especially interested in the full permissions that it describes just after that link (so that you can see admin/push/pull).

I had a quick look to see if I could send a PR (the call is easy, but lost my way figuring out how to convert that into a Repository!)

s-t-e-v-e-n-k:
What do you mean convert it into a Repository? I can see arguments both ways for having it on Repository or on Team. Where are you getting stuck?

tl-adrian-bridgett:
Hi, thanks for such a speedy response! :-)

The JSON returned from the "GET /orgs/:org/teams/:team_slug/repos/:owner/:repo" call looks like a repository (but I guess it's a slightly different view than that you'd receive from "GET /repos/:owner/:repo" as the permissions will reflect the team, I'm not sure what the other fields will be).

I was getting stuck figuring out how to return a Repository.  With a bit more time today I've figured out what to copy&paste so there's now a PR:
https://github.com/PyGithub/PyGithub/pull/1416

PR #1416, opened 2020-03-04
Add "get_repo_permission" to Team class (for #1415)

tl-adrian-bridgett:
This adds support in a way that matches GitHub API:
https://developer.github.com/v3/teams/#check-if-a-team-manages-a-repository

I've tried to stick to the coding standards but I'm not _too_ convinced.

BTW this account is going to disappear on Friday (change of job) and/or I may well stop being able to respond :-(


s-t-e-v-e-n-k:
I sort of envisioned it checking the status code returned by GitHub and returning a boolean, True if the team has permissions and False if not. There's a tox environment to run the formatting to help with that: ``tox -eformat``

WRT your account, that's disappointing, but Github accounts do support multiple email addresses

tl-adrian-bridgett:
Ah. I'm really after the full set of permissions (admin/push/pull etc) rather than just the boolean value. Hmm. Maybe it should be returning the Permission class though if possible.

s-t-e-v-e-n-k:
That sounds like a much better fit, indeed. Also, please add at least one test -- preferably two, to check the you don't have any permissions on this repo case. You could also probably extend the method to take a repo or a string as the repo name, but I'm not wedded to that idea.

tl-adrian-bridgett:
Good news and bad news
Good news - I've made the improvements (and added my personal email address to github)
Bad news - I can't get the tests to pass but not sure what's going on!

s-t-e-v-e-n-k:
This is going to be a fun rebase. :-P

The test failures are because you're making a new API call in the tests, and the replay data (tests/ReplayData/Team.testRepos.txt in this case) doesn't know about that the call, and so is matching with a different call.

If you get stuck, mention me, and I'll dig in further.

s-t-e-v-e-n-k:
Must remember to put "Fixes #1415" in your commit body when I merge it...

tl-adrian-bridgett:
Should be good now, I updated the replay data by using a temporary personal repo then merge/update the two new requests.  Thanks so much for all your help, I'd never have made it without you :-)

tests/Team.py
```diff
@@ -114,11 +116,15 @@ def testRepos(self):
...
         self.assertFalse(self.team.has_in_repos(repo))
+        self.assertIsNone(self.team.get_repo_permission(repo))
         self.team.add_to_repos(repo)
         self.assertListKeyEqual(
             self.team.get_repos(), lambda r: r.name, ["FatherBeaver"]
         )
         self.assertTrue(self.team.has_in_repos(repo))
+        self.assertIsInstance(
+            self.team.get_repo_permission(repo), github.Permissions.Permissions
+        )
```
s-t-e-v-e-n-k: I'd rather you assert something in the returned object, such as:

```
permissions = self.team.get_repo_permission(repo)
self.assertTrue(permissions.pull)
```

tl-adrian-bridgett: Sounds good, that also drops the need for the import.  I've force-pushed (first time I used github UI to resolve the Const conflict but now I've rebased and fixed it there instead)

PR #1409, opened 2020-02-25
Add support for editing project cards

scjody:
One more step towards full project support (https://github.com/PyGithub/PyGithub/issues/606)

https://developer.github.com/v3/projects/cards/#update-a-project-card

PR #1418, opened 2020-03-05
 Add support for editing project cards

scjody:
One more step towards full project support (https://github.com/PyGithub/PyGithub/issues/606)

https://developer.github.com/v3/projects/cards/#update-a-project-card

Revised version of https://github.com/PyGithub/PyGithub/pull/1409


tests/Project.py
```diff
@@ -187,3 +187,29 @@ def testCreateCardFromIssue(self):
         issue = self.repo.create_issue(title="Issue created by PyGithub")
         card2 = column.create_card(content_id=issue.id, content_type="Issue")
         self.assertEqual(card2.id, 16039106)
+
+    def testEditCardWithoutParameters(self):
+        project = self.repo.create_project(
```
s-t-e-v-e-n-k: I was checking into this yesterday, and I think you would be better placed grabbing the project by id directly and making use of that replay data. For example, as seen on line 149 in this file.

s-t-e-v-e-n-k:
Is this the last missing piece for #606? Deleting Project Cards landed yesterday.

scjody:
> Is this the last missing piece for #606? Deleting Project Cards landed yesterday.

It looks like we're still missing a few small things:
* [Create an organization project](https://developer.github.com/v3/projects/#create-an-organization-project)
* [Create a user project](https://developer.github.com/v3/projects/#create-a-user-project)
* [List user projects](https://developer.github.com/v3/projects/#list-user-projects)
* [Update a project](https://developer.github.com/v3/projects/#update-a-project)
* [Delete a project](https://developer.github.com/v3/projects/#delete-a-project)

I'll add this list to #606.

tests/Project.py
scjody: Thanks for the suggestion! As you can see that worked 😸

PR #1423, opened 2020-03-06
Switch to using pytest

s-t-e-v-e-n-k:
Due to the recent deprecation of setup.py test, we should move with the
times, and switch to a non-deprecated test runner.

Add configuration for pytest, and switch tox to using it. Remove
AllTests, since it is no longer required to locate test classes, and
rewrite tests/__main__.py as a pytest conftest plugin.

Also drop the test function from manage.sh, tox does a much better job.


setup.py
```diff
@@ -98,9 +98,7 @@
             "Programming Language :: Python :: 3.8",
             "Topic :: Software Development",
         ],
-        test_suite="tests.AllTests",
         python_requires=">=3.5",
         install_requires=["deprecated", "pyjwt", "requests>=2.14.0"],
         extras_require={"integrations": ["cryptography"]},
-        tests_require=["cryptography", "httpretty>=0.9.6", "parameterized>=0.7.0"],
```
sfdye: I think we can probably keep the `tests_require`, as some downstream usage may depend on PyGithub for these dependencies

tests/conftest.py
```diff
@@ -28,27 +22,21 @@
...
-    if "--auth_with_token" in argv:
+def pytest_configure(config):
+    if config.getoption("record"):
+        Framework.activateRecordMode()
+    if config.getoption("auth_with_token"):
         Framework.activateTokenAuthMode()
-        argv = [arg for arg in argv if arg != "--auth_with_token"]
-
-    if "--auth_with_jwt" in argv:
+    if config.getoption("auth_with_jwt"):
```
sfdye: Does the record and replay testing framework still work? 

Maybe we should also update the docs: https://github.com/PyGithub/PyGithub/blob/08b1c474b32d3f32bc64a59a7e77d87c43527e1f/CONTRIBUTING.md#L73

sfdye:
💯 This is so good @s-t-e-v-e-n-k

tests/conftest.py
s-t-e-v-e-n-k: Embarrassing truth time: I don't use it. However, I did check that the right methods in Framework are called if you call pytest with --record and such.

setup.py
s-t-e-v-e-n-k: Even if this this list is outdated compared to the previous test-requirements.txt? :-)

sfdye: tests_require is not the same as requirement.txt, it should provide the minimum version required
https://packaging.python.org/discussions/install-requires-vs-requirements/

tests/conftest.py
s-t-e-v-e-n-k: Excellent point about updating the docs, I'm doing so.

sfdye: I see, anyway we should update all reference of `python -m test` to `pytest`

s-t-e-v-e-n-k:
@sfdye Forced pushed with the docs updated and tests_require resurrected.

sfdye:
Awesome, have you tested the -k to run specific class or method?

s-t-e-v-e-n-k:
```
(py36) steven@wrecked:~/PyGithub% pytest -k RequiredStatusChecks
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.3.5, py-1.8.1, pluggy-0.13.1
rootdir: /home/steven/PyGithub, inifile: pytest.ini
plugins: cov-2.8.1
collected 607 items / 604 deselected / 3 selected                              

tests/Branch.py ..                                                       [ 66%]
tests/RequiredStatusChecks.py .                                          [100%]

====================== 3 passed, 604 deselected in 0.74s =======================
(py36) steven@wrecked:~/PyGithub% pytest tests/RequiredStatusChecks.py -k testAttributes 
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.3.5, py-1.8.1, pluggy-0.13.1
rootdir: /home/steven/PyGithub, inifile: pytest.ini
plugins: cov-2.8.1
collected 1 item                                                               

tests/RequiredStatusChecks.py .                                          [100%]

============================== 1 passed in 0.05s ===============================
```

-k is pretty cool.

tests/conftest.py
s-t-e-v-e-n-k: Am I being judged for using pdb and list comprehensions in a text editor to generate my replay date? :laughing:

test-requirements.txt
```diff
@@ -1,4 +1,5 @@
 cryptography
 httpretty>=0.9.6
 parameterized>=0.7.0
-coverage>=5.0.3
```
sfdye: One more thing, maybe we should be explicitly installing coverage rather relying transitive dep from pytesy (my guess)

s-t-e-v-e-n-k: pytest-cov still requires coverage<5, and I daresay will always use coverage, so I'd rather pytest-cov pulling it in so the correct version is installed.

s-t-e-v-e-n-k: Due to pip's resolver, if we include coverage and get out of step, we could hit the issue of installing a version of coverage that pytest-cov does not support.

PR #1428, opened 2020-03-09
Don't transform completely in PullRequest.*assignees

### LINKED ISSUE #1406: add_to_assignees() breaks create_review_request()
Given a PullRequest object `pull`, the following sequence of operations leads to a 404 error:

1. `pull.add_to_assignees("fpoli")`
2. `pull.create_review_request(reviewers=["some-user"])`

However, the following works fine:

1. `deepclone(pull).add_to_assignees("fpoli")`
2. `pull.create_review_request(reviewers=["some-user"])`

Possible reason: internally, `add_to_assignees` calls the `_useAttributes` defined in the `Issue` class, which replaces the `self._url` of the pull request with the url of an _issue_. Then, the `create_review_request` call fails because it sends a POST request to a non-existing endpoint like <https://api.github.com/repos/viperproject/viperserver/issues/10/requested_reviewers>, which should be <https://api.github.com/repos/viperproject/viperserver/pulls/10/requested_reviewers>.

s-t-e-v-e-n-k:
The two methods PullRequest.add_to_assignees() and
PullRequest.remove_from_assignees() are special, in that they call
POST/DELETE on the issue_url of the pull request, and then transform all
their attributes based on what was returned. This has a number of
effects, none of them any good for calling new methods on the same pull
request object. Instead, only pull out the changed attribute and use
that. Also change the returned URL for the POST call in the replay data
to match what GitHub returns.

Fixes #1406


sfdye:
black failed 😄

s-t-e-v-e-n-k:
@sfdye Force pushing dismisses reviews :-( But that's what I get for trying to push up a PR at 10pm :-P

PR #1433, opened 2020-03-11
Add method for fetching pull requests associated with a commit

### LINKED ISSUE #1414: List pull requests based on commit
There's a functionality on the Github API that is apparently not supported on PyGithub -- https://developer.github.com/v3/repos/commits/#list-pull-requests-associated-with-commit

## Motivation

I'm doing an auto-versioning type of thing on my repo, and I need to walk through all the pull requests searching for labels. I can get all the commits since the last release using `<repo>.get_commits(since=date_from_last_release)`. It would be incredibly useful if those commits came with their associated pull requests - that way I could simply check for each label on those.

I couldn't find this functionality on the documentation or reading the code, and I couldn't find another decent way to do this on the current PyGithub implementation.

## Caveats

This feature seems to be in preview period and the API might change, so it's probably not worth doing this right now (if ever).

mawkee:
## Closes Feature Request
- Closes #1414

## Explanation
GitHub implements a direct link between a commit and all pull requests associated with that commit through a preview API, as explained here https://developer.github.com/v3/repos/commits/#list-pull-requests-associated-with-commit

## Notes
The test actually gets a new commit (instead of using the one set on the test class instance). That's because that commit has no corresponding PRs. If preferred, I can reimplement the other tests pointing to another commit that might cover - or I can just fake it on the test replay data.


tests/Commit.py
```diff
@@ -150,3 +150,11 @@ def testCreateStatusWithAllParameters(self):
...
         )
         self.assertEqual(status.description, "Status successfuly created by PyGithub")
+
+    def testGetPulls(self):
+        commit = (
+            self.g.get_user()
+            .get_repo("PyGithub")
+            .get_commit("e44d11d565c022496544dd6ed1f19a8d718c2b0c")
+        )
+        self.assertListKeyEqual(commit.get_pulls(), lambda c: c.id, [385799287])
```
s-t-e-v-e-n-k: Could I ask that you assert on PR number, not the ID? Other than that, this looks fantastic!

mawkee: Sure!

mawkee: Done. Can you re-check please?

PR #1434, opened 2020-03-12
Add support to edit and delete a project

ahhda:
Implements edit and delete project part of https://github.com/PyGithub/PyGithub/issues/606#issuecomment-595864374

<details>
<summary>
Manual testing
</summary>

Before editing the project
Screenshot 2020-03-12 at 9 49 53 PM

After editing the project
Screenshot 2020-03-12 at 9 55 52 PM

After deleting the project
Screenshot 2020-03-12 at 9 56 23 PM
</details>


ahhda:
I have created a single pull request for both the changes since the changes were similar.

Let me know if this needs to be broken into two parts. Thanks! 😇

github/Project.py
```diff
@@ -140,6 +140,60 @@ def url(self):
...
 
+    def delete(self):
+        """
+        :calls: `DELETE /projects/:project_id <https://developer.github.com/v3/projects/#delete-a-project>`_
+        :rtype: None
+        """
+        headers, data = self._requester.requestJsonAndCheck(
+            "DELETE", self.url, headers={"Accept": Consts.mediaTypeProjectsPreview}
+        )
+
```
s-t-e-v-e-n-k: Other delete methods usually return True and check the status returned by GitHub, could you do the same here? You can then ``self.assertTrrue()`` in the test case.

tests/Project1434.py
```diff
@@ -0,0 +1,45 @@
...
+    def setUp(self):
+        super().setUp()
+
+    def testDelete(self):
+        project = self.g.get_project(4102095)
+        project.delete()
+
+    def testEditWithoutParameters(self):
+        project = self.g.get_project(4101939)
+        project.edit()
```
s-t-e-v-e-n-k: Assert the name or something here to just to check at least one of the attributes.

tests/Project1434.py
```diff
@@ -0,0 +1,45 @@
...
+from . import Framework
+
+
+class Project1434(Framework.TestCase):
+    def setUp(self):
+        super().setUp()
+
+    def testDelete(self):
+        project = self.g.get_project(4102095)
+        project.delete()
```
s-t-e-v-e-n-k: As above -- we should be asserting the function return type too.

github/Project.py
ahhda: I checked the documentation and the result of the call. The call to [the above URL](https://developer.github.com/v3/projects/#delete-a-project), doesn't return a True or False. It just gets us `None`.

I checked the other methods as well, none of them return any data.

https://github.com/PyGithub/PyGithub/blob/b7894ea00c9993f5d129091b15ce0232da1e1aea/github/GitRef.py#L68-L75
https://github.com/PyGithub/PyGithub/blob/b7894ea00c9993f5d129091b15ce0232da1e1aea/github/Branch.py#L250-L256
https://github.com/PyGithub/PyGithub/blob/fc4eff3185af49239712dcb32779a6c51b788136/github/Migration.py#L154-L162

Let me know if there is a way around this.

tests/Project1434.py
ahhda: As mentioned here, https://github.com/PyGithub/PyGithub/pull/1434#discussion_r392459033, the delete method doesn't return a boolean. Not sure if it would be possible without that.

Checked other tests as well.

https://github.com/PyGithub/PyGithub/blob/37122cf2bbde825e57e78ee37171bab150802eca/tests/Reaction.py#L54-L55

https://github.com/PyGithub/PyGithub/blob/37122cf2bbde825e57e78ee37171bab150802eca/tests/GitRef.py#L69-L70

https://github.com/PyGithub/PyGithub/blob/37122cf2bbde825e57e78ee37171bab150802eca/tests/GistComment.py#L68-L69

tests/Project1434.py
ahhda: Resolved. Added a check to test the project name.

github/Project.py
s-t-e-v-e-n-k: Right, I was confusing it with another method, sorry.

PR #1438, opened 2020-03-15
Use pytest to parametrize tests

mgorny:
Use built-in pytest test case parametrization support over external
'parameterize' package.  The latter is not well maintained, and has
known Python 3.8 failures unsolved since November 2019.

Since pytest fixtures are incompatible with unittest-style tests,
rewrite the relevant test case to use pytest-style asserts.  This also
makes the resulting code simpler, as we no longer have to pass TestCase
to the helper classes.

### DIFF
--- a/tests/Framework.py
+++ b/tests/Framework.py
@@ -150,8 +150,7 @@ class RecordingHttpsConnection(RecordingConnection):
 
 
 class ReplayingConnection:
-    def __init__(self, testCase, file, protocol, host, port, *args, **kwds):
-        self.__testCase = testCase
+    def __init__(self, file, protocol, host, port, *args, **kwds):
         self.__file = file
         self.__protocol = protocol
         self.__host = host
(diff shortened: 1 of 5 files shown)


s-t-e-v-e-n-k:
I think this is great idea! I'm not clear on what the root cause of the failures are, though :-(

sfdye:
I'd really want to merge this, after Travis is green

mgorny:
I'm sorry, I have failed horribly and for some reason forsake to test it properly. Working on a fix right now.

mgorny:
Ok, it finally passes now ;-).

tests/Framework.py
```diff
@@ -171,25 +170,20 @@ def request(self, verb, url, input, headers):
...
-                    json.loads(expectedInput),
-                )
+                assert json.loads(
+                    input.replace("\n", "").replace("\r", "")
+                ) == json.loads(expectedInput)
             else:
-                self.__testCase.assertEqual(
-                    input.replace("\n", "").replace("\r", ""), expectedInput
-                )
+                assert input.replace("\n", "").replace("\r", "") == expectedInput
```
s-t-e-v-e-n-k: It seems we do the replace here no matter if the input is JSON or not, would you mind refactoring it so we do it unilaterally? Should also make the asserting for the json.loads() rather more readable too.

mgorny: Done.

s-t-e-v-e-n-k:
Thanks! It looks great.

PR #1444, opened 2020-03-24
Slightly clean up tox.ini

s-t-e-v-e-n-k:
Move the 3 from env list, since we can save a whole 4 characters! Also
drop a flake8 ignore that we no longer need.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     lint,
-    py{35,36,37,38},
+    py3{5,6,7,8},
     docs
 
 [testenv]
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     lint,
-    py{35,36,37,38},
+    py3{5,6,7,8},
```
sfdye: I am against, what if python4 is released?

s-t-e-v-e-n-k: In which case, I'd be tempted to nest the braces: py{3{7,8,9},40} , but okay, I shall try and stop myself from injecting shell expansion into our tox config file. :-)

PR #1449, opened 2020-03-30
Add create_repository_dispatch

### LINKED ISSUE #1299: Support for repository dispatch event
Preview feature useful for triggering github actions.

POST /repos/:owner/:repo/dispatches
https://developer.github.com/v3/repos/#create-a-repository-dispatch-event

christopher-dG:
Closes #1299 

The code is working fine but I'm not quite sure what kind of tests I should be adding or where. Some guidance there would be very much appreciated!

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -1373,6 +1373,27 @@ class Repository(github.GithubObject.CompletableGithubObject):
             self._requester, headers, data, completed=True
         )
 
+    def create_repository_dispatch(
+        self, event_type, client_payload=github.GithubObject.NotSet
+    ):
+        """
+        :calls: POST /repos/:owner/:repo/dispatches <https://developer.github.com/v3/repos/#create-a-repository-dispatch-event>
+        :param event_type: string
+        :param client_payload: dict
+        :rtype: bool
+        """
+        assert isinstance(event_type, str), event_type
+        assert client_payload is github.GithubObject.NotSet or isinstance(
+            client_payload, dict
+        ), client_payload
+        post_parameters = {"event_type": event_type}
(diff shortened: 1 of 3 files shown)


github/Repository.py
```diff
@@ -1373,6 +1373,28 @@ def __create_pull(self, **kwds):
...
+            client_payload, dict
+        ), client_payload
+        post_parameters = {"event_type": event_type}
+        if client_payload is not github.GithubObject.NotSet:
+            post_parameters["client_payload"] = client_payload
+
+        headers, data = self._requester.requestJsonAndCheck(
+            "POST", self.url + "/dispatches", input=post_parameters
+        )
+        return None
```
s-t-e-v-e-n-k: This returns a 204, and we already return a truthy value in that case. See other places in this very file for "return status == 204".

github/Repository.py
```diff
@@ -1373,6 +1373,28 @@ def __create_pull(self, **kwds):
...
         )
 
+    def create_repository_dispatch(
+        self, event_type, client_payload=github.GithubObject.NotSet
+    ):
+        """
+        :calls: POST /repos/:owner/:repo/dispatches <https://developer.github.com/v3/repos/#create-a-repository-dispatch-event>
+        :param event_type: string
+        :param client_payload: dict
+        :rtype None
```
s-t-e-v-e-n-k: I think this is missing a : ?

s-t-e-v-e-n-k: And should be updated when you change the code too, of course.

christopher-dG:
Added a test, I think I did it right.

PR #1450, opened 2020-04-02
Export IncompletableObject in the github namespace

s-t-e-v-e-n-k:
When IncompletableObject was added, it was only added to
GithubException, and not imported directly into the github namespace
like other exceptions. Correct that, and clean up the resultant now
unrequired import.


s-t-e-v-e-n-k:
@sfdye Where did the Travis check go? Are we planning to switch to GitHub Checks?

sfdye:
I don’t mind switching to GH Action, if there is a working PR I can review :)

s-t-e-v-e-n-k:
This one? :-)

sfdye:
Maybe a separate one just for that purpose?

PR #1456, opened 2020-04-07
GitHub Action to lint Python

cclauss:
https://github.com/cclauss/PyGithub/actions to replace Travis CI which appears to be disabled.

Codespell issues at #1459

Pytest issues...
=========================== short test summary info ==========================
FAILED tests/GithubIntegration.py::GithubIntegration::testCreateJWT - NotImpl...
FAILED tests/GithubIntegration.py::GithubIntegration::testGetAccessToken - No...
FAILED tests/GithubIntegration.py::GithubIntegration::test_get_installation
FAILED tests/GithubIntegration.py::GithubIntegration::test_get_installation_custom_base_url
======================== 4 failed, 620 passed in 24.69s ========================

### DIFF
--- a/.github/workflows/.lint_python.yml
+++ b/.github/workflows/.lint_python.yml
@@ -17,10 +17,10 @@ jobs:
       - if: matrix.python-version >= 3.6
         run: |
           pip install black
-          black --check . || true
+          black --check .
       - run: codespell --quiet-level=2 || true  # --ignore-words-list="" --skip=""
       - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-      - run: isort --recursive . || true
+      - run: isort --recursive .
       - run: pip install -r requirements.txt
-      - run: mypy --ignore-missing-imports . || true
+      - run: mypy --ignore-missing-imports .
       - run: pytest . || true
--- a/.github/workflows/.lint_python.yml
+++ b/.github/workflows/.lint_python.yml
@@ -1,0 +1,26 @@
+name: lint_python
+on:
+  pull_request:
+  push:
+  #  branches: [master]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.5, 3.6, 3.7, 3.8]  # , pypy3]
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@master
+      - run: pip install codespell flake8 isort mypy pytest
+      - if: matrix.python-version >= 3.6
+        run: |
+          pip install black
+          black --check . || true
+      - run: codespell --quiet-level=2 || true  # --ignore-words-list="" --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --recursive . || true
+      - run: pip install -r requirements.txt
+      - run: mypy --ignore-missing-imports . || true
+      - run: pytest . || true
--- a/.github/workflows/GitHub_Action.yml
+++ b/.github/workflows/GitHub_Action.yml
@@ -17,9 +17,9 @@ jobs:
         run: |
           pip install black
           black --check .
-      - run: codespell --quiet-level=2 || true  # --ignore-words-list="" --skip=""
+      - run: codespell --ignore-words-list="bloaded,nto,pullrequest,tim,wan" --quiet-level=2 --skip="./tests/*" || true
       - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
       - run: isort --recursive .
       - run: pip install httpretty -r requirements.txt
       - run: mypy --ignore-missing-imports .
-      - run: pytest .
+      - run: pytest . || true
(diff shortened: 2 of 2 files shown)


.github/workflows/.lint_python.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  pull_request:
+  push:
+  #  branches: [master]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
s-t-e-v-e-n-k, line 11: We surely don't need to test across all three platforms.

.github/workflows/.lint_python.yml
```diff
@@ -0,0 +1,26 @@
+name: lint_python
```
s-t-e-v-e-n-k, line 1: Maybe just lint?

.github/workflows/.lint_python.yml
```diff
@@ -0,0 +1,26 @@
...
+  lint_python:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.5, 3.6, 3.7, 3.8]  # , pypy3]
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@master
+      - run: pip install codespell flake8 isort mypy pytest
```
s-t-e-v-e-n-k, line 16: I think you should use our requirements files so they don't get out of sync. Further, those support environment markers.

.github/workflows/.lint_python.yml
```diff
@@ -0,0 +1,26 @@
...
+      - if: matrix.python-version >= 3.6
+        run: |
+          pip install black
+          black --check . || true
+      - run: codespell --quiet-level=2 || true  # --ignore-words-list="" --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --recursive . || true
+      - run: pip install -r requirements.txt
+      - run: mypy --ignore-missing-imports . || true
+      - run: pytest . || true
```
s-t-e-v-e-n-k, line 26: This also means the name is a lie, since we are no longer linting, but running the test suite.

Further more, could you rely on the tox environment we have for this very purpose?

.github/workflows/.lint_python.yml
cclauss: These are not runtime requirements.  We do not need end users to install these.  We only need these dependencies at test time.

.github/workflows/.lint_python.yml
cclauss: GitHub_Action

.github/workflows/.lint_python.yml
cclauss: Why tox?  It is so old school.  Why not an automatic https://pre-commit.com ?

.github/workflows/.lint_python.yml
cclauss: Dropped macOS and Windows

.github/workflows/.lint_python.yml
s-t-e-v-e-n-k: Our tox environment uses pre-commit, and duplicates most of that. Also, way to win minds when you call something old school that we switched to less than 2 months ago ... :-P

.github/workflows/.lint_python.yml
gschaffner: I believe that dev dependencies are typically in `requirements.txt` while end-user ones are in `setup.py`. `pip install pygithub` does not look at `requirements.txt` but only `install_requires` in `setup.py`.

https://packaging.python.org/discussions/install-requires-vs-requirements/

cclauss: Cool!  I have never seen that.  I usually see `requirements_dev.txt` and `requirements.txt` but I am good with the model you suggest.

.github/workflows/GitHub_Action.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/setup-python@master
+      - run: pip install codespell flake8 isort mypy pytest
+      - if: matrix.python-version >= 3.6
+        run: |
+          pip install black
+          black --check .
+      - run: codespell --ignore-words-list="bloaded,nto,pullrequest,pullrequests,thi,tim,wan" --quiet-level=2 --skip="./tests/*" || true
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --recursive .
+      - run: pip install httpretty -r requirements.txt
```
s-t-e-v-e-n-k, line 23: I think the test failures are due to not installing the test requirements. httpretty and pytest is included in them, so you can drop it from the above steps:

``pip install -r requirements.txt -r test-requirements.txt``

.github/workflows/GitHub_Action.yml
```diff
@@ -0,0 +1,25 @@
+name: GitHub_Action
```
s-t-e-v-e-n-k, line 1: How about "check" ?

cclauss:
Closing in favor of  #1464

PR #1464, opened 2020-04-13
Add GitHub Action workflow for checks

s-t-e-v-e-n-k:
To integrate better with the GitHub workflow, add a workflow for
checking pushes and PRs. This doesn't remove the Travis configuration,
which is still required for releases.

Co-Authored-By: Christian Clauss <cclauss@me.com>

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,16 @@
+name: check
+on:
+  pull_request:
+  push:
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+      - run: pip install tox -r requirements.txt -r test-requirements.txt
+      - run: tox -elint
+      - run: pytest --cov=github


.github/workflows/check.yml
```diff
@@ -0,0 +1,17 @@
+name: check
+on:
+  pull_request:
+  push:
+jobs:
+  pre-commit:
```
cclauss, line 6:
```suggestion
  check:
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+      - run: pip install tox pre-commit
```
cclauss, line 14:
```suggestion
      - run: pip install tox
```
tox already installs pre-commit.

.github/workflows/check.yml
```diff
@@ -0,0 +1,17 @@
...
+      - run: pip install -r requirements.txt -r test-requirements.txt
```
cclauss, line 15: It is usually safer to edit `test-requirements.txt` to have the first line of be `-r requirements.txt` but this will work as well.

Issue #1057, opened 2019-03-05
Missing api for selecting a column in the Project board


shyam334:
Hello Guys,

I was using the library to automate a bit of the Project board feature, and noticed that there isn't a `get_column` method. I thought I'd report it as an issue/feature request.
The intent is to select a particular column in the Project board.

Corresponding GH api : `GET /projects/columns/:column_id`

https://developer.github.com/v3/projects/columns/#get-a-project-column
https://developer.github.com/enterprise/2.16/v3/projects/columns/#get-a-project-column

Thanks.

viv2793:
@sfdye @shyam334 Can anyone assign this to me ?

PR #1466, opened 2020-04-14
Get a project column by id

FlorentClarret:
This PR resolves the https://github.com/PyGithub/PyGithub/issues/1057
I also added some tests for the ProjectColumn class


FlorentClarret:
I just noticed that other ProjectColumn related methods are missing like update and delete. @sfdye do you think we should keep adding those methods to the main class or we can add them somewhere else to avoid putting everything in the main class? As static method to the ProjectColumn class maybe?

sfdye:
Let’s add to the main for now. Will revisit the refactoring later. 

cc @s-t-e-v-e-n-k

s-t-e-v-e-n-k:
ProjectColumn should match ProjectCard -- I think the delete() and edit() methods would be incredibly similar.

FlorentClarret:
Oh my bad, I didn't see them, you're right. I will add them in another PR

FlorentClarret:
Thanks for the review ! If you want me to add something else to this PR, please tell me ;)

PR #1487, opened 2020-04-26
Use Github action to upload to PyPI on new release

sfdye:
- Added `secrets.PYPI_PASSWORD` to project settings
- Use twine to build artifact and upload to PyPI on new release

TODO

- [ ] Test twine with new patch release (1.47.1)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -16,7 +16,7 @@ jobs:
           python-version: "${{ matrix.python-version }}"
       - name: "Install dependencies"
         run: |
-          pip install --upgrade pip
+          python -m pip install --upgrade pip
           pip install tox tox-gh-actions
       - name: Test with tox
         run: tox
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,24 @@
+name: Publish to PyPI
+on:
+  release:
+    types: [created]
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: '3.x'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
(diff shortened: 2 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           python-version: "${{ matrix.python-version }}"
       - name: "Install dependencies"
         run: |
-          pip install --upgrade pip
+          python -m pip install --upgrade pip
```
sfdye, line 19: this is preferred way by latest pip

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,24 @@
+name: Publish to PyPI
+on:
+  release:
+    types: [created]
```
sfdye, line 4: Although I am not entirely sure the difference betwewen `created` and `published`, let's stick with the [example](https://github.com/actions/starter-workflows/blob/8c4fc57de83ba017cd877a7fad3f4dd665c5c4b3/ci/python-publish.yml#L8) for now

sfdye:
Merging this to test the new workflow.

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
```
s-t-e-v-e-n-k, line 7: There's a workflow that does most of the heavy lifting: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

sfdye: I am aware, it's too complicated. Let me see if this simple approach works.

sfdye:
It works! on first try 🎉 🎉 🎉 

image

image

I love Github workflow ❤️

s-t-e-v-e-n-k:
Wow! Nice work! Actually, one thing I forgot to point out is we should also update the docs. I'll look at hooking codecov tomorrow and then we can drop travis.

sfdye:
What do you I am doing right now :)

I will leave codecov to you 👍

PR #1489, opened 2020-04-26
Improve type stubs

emilazy:
Some changes to help PyGithub work properly with mypy:

* Install the `.pyi` files and a `py.typed` marker file to allow type checkers to find the type stubs when checking other code that uses PyGithub, per [the mypy documentation/PEP 561](https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages).
* Fill out some missing type stubs and superclasses.
* Make `PaginatedList` generic on the element type.

I was able to successfully type-check [a script using PyGithub](https://github.com/NixOS/nixpkgs/pull/86074) with these changes.


s-t-e-v-e-n-k:
Wow! This looks amazing! What were you using to find all of these issues, and are we able to integrate it with CI?

emilazy:
I just did `mypy --strict my-script.py` and fixed things until it passed :) I used some awful shell one-liners to partially automate looking for bare `PaginatedList` and inserting the superclass declarations, but nothing fancier than that.

I guess the best way to test it would be to make sure a simple example script importing PyGithub passes? That would test the PEP 561 file placement to make sure other packages can use the type stubs. The PaginatedList/GithubObject changes seem harder to test other than by relying on them in an example; the previous type stubs weren't incorrect so much as they were too underspecified to check useful scripts (because e.g. `release.<GitRelease method>() for release in repo.get_releases()` would error out as `release` has type `GithubObject`).

sfdye:
Wow, thanks a ton @emilazy 

TIL `mypy --strict`

sfdye:
This is now published to PyPI:
 https://pypi.org/project/PyGithub/1.51/

PR #1497, opened 2020-05-01
Move, edit and delete project columns

FlorentClarret:
This PR adds missing methods for the project columns


s-t-e-v-e-n-k:
This is looking fantastic! Would you mind also adding the new methods to the type hints file -- github/ProjectColumn.pyi in this case.

FlorentClarret:
Oops... I'm not used to pyi files and always forget them. It should be ok now

sfdye:
In hindsight, we should document about `.pyi` files in the contributing docs and we need a PR template!

PR #1502, opened 2020-05-02
Add create_repository_dispatch to typing files

christopher-dG:
Forgot to do this in my earlier PR.

I would really love to have the type annotations in the original `.py` files, it would be a lot easier to remember that way 🙂


sfdye:
Merging type stubs into source code might be something we consider in the future.

Here in an interesting read:
https://pyfound.blogspot.com/2020/04/the-path-forward-for-typing-python.html

PR #1503, opened 2020-05-04
Improve type signature for create_from_raw_data

sophiebits:
Previously you could call it only on Repository or NamedUser. Now you can call it with any GithubObject subclass -- and the type is carried through to the return value.

FWIW, this is preventing me from upgrading to 1.51.

### DIFF
--- a/github/MainClass.pyi
+++ b/github/MainClass.pyi
@@ -1,13 +1,13 @@
 from datetime import datetime
 from io import BytesIO
-from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, TypeVar, Union
 
 from github.AuthenticatedUser import AuthenticatedUser
 from github.Commit import Commit
 from github.ContentFile import ContentFile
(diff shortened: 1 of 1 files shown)


github/MainClass.pyi
```diff
@@ -46,10 +48,10 @@ class Github:
     def __set_per_page(self, value: int) -> None: ...
     def create_from_raw_data(
         self,
-        klass: Union[Type[Repository], Type[NamedUser]],
+        klass: Type[TGithubObject],
         raw_data: Dict[str, Any],
         headers: Dict[str, Union[str, int]] = ...,
-    ) -> Union[Repository, NamedUser]: ...
+    ) -> TGithubObject: ...
     def dump(self, obj: Repository, file: BytesIO, protocol: int = ...) -> None: ...
```
s-t-e-v-e-n-k, line 55: Should this also change from Repository to TGithibObject as well?

sophiebits: I'm not using it in my code so didn't notice, but I suppose so! Update pushed.

PR #1506, opened 2020-05-05
Raise a FutureWarning on use of client_{id,secret}

### LINKED ISSUE #1500: Deprecating API authentication through query parameters
I'm using `g = Github("access_token")` on PyGithub-1.50 (latest) and got the below message from github. I am I doing something wrong?

```
On May 1st, 2020 at 16:31 (UTC) your personal access token using python-requests/2.22.0 was used as part of a query parameter to access an endpoint through the GitHub API:

[removed]

Please use the Authorization HTTP header instead, as using the `access_token` query parameter is deprecated. If this token is being used by an app you don't have control over, be aware that it may stop working as a result of this deprecation.

Depending on your API usage, we'll be sending you this email reminder on a monthly basis for each token and User-Agent used in API calls made on your behalf.
Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.

Visit https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param for more information about suggested workarounds and removal dates.
```

s-t-e-v-e-n-k:
client_id and client_secret are sent to GitHub encoded on the query
string, which GitHub has firmly deprecated, so much so as to email
people who are doing so. To discourage its use, raise a FutureWarning if
they are specified.

Fixes #1500

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -51,6 +51,7 @@
 import datetime
 import pickle
 import time
+import warnings
 
 import jwt
 import requests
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
@sfdye I've tested it by hand, or are you going to ask me nicely to write a unit test?

sfdye:
That's good enough

PR #1521, opened 2020-05-18
Drastically increase coverage by checking repr()

s-t-e-v-e-n-k:
A lot of our missing coverage is caused by not checking repr(), which is
trivial enough to test against -- do so, and also clean up existing repr
tests to look the same by calling repr(). Where it was trivial, add a
few assertions about missing attributes.


s-t-e-v-e-n-k:
From 42 files completely covered to 71.

s-t-e-v-e-n-k:
Not 71, 72.

PR #1565, opened 2020-06-18
Small documentation correction in Repository.py

mark-bromell:
Class `github.Clone.Clone` doesn't exist and should reference `github.Clones.Clones` instead.

Short and simple PR, fixing this makes the documentation more accurate, and makes navigation in readthedocs easier.


s-t-e-v-e-n-k:
Nice catch, thanks!

PR #1574, opened 2020-06-24
User format to enable Hex user name

henrykironde:
This allows user names that correspond to Hex values like 0000

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -253,7 +253,7 @@ class Github(object):
             github.Event.Event, self.__requester, "/events", None
         )
 
-    def get_user(self, login=github.GithubObject.NotSet):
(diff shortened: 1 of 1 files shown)


github/MainClass.py
```diff
@@ -253,7 +253,7 @@ def get_events(self):
             github.Event.Event, self.__requester, "/events", None
         )
 
-    def get_user(self, login=github.GithubObject.NotSet):
+    def get_user(self, login="{0}".format(github.GithubObject.NotSet)):
```
s-t-e-v-e-n-k, line 256: I daresay this isn't the correct place, and we should cast login to str on line 269.

s-t-e-v-e-n-k:
And this won't in fact fix the issue you're trying to work around. The assert() to str is for type safety.

PR #1580, opened 2020-06-24
Add storing to PaginatedList

OlegYurchik:
I add storing parameter in functions search_code and search_commits. It is work - I tried it in my server. In future I can add in other methods which used PaginatedList as return object


s-t-e-v-e-n-k:
Right, I see. PaginatedList should probably be turned into a generator that yields the next element rather than massively growing an array -- but I don't think this is the right approach.

s-t-e-v-e-n-k:
This is going to require more thought here. Do you mind if I close this, and we move to discussing this on the issue you opened?

Issue #1586, opened 2020-06-27
More support for Deployments


colbygallup:
I am trying to create a server that uses [GitHub Deployments](https://developer.github.com/v3/repos/deployments/) for a continuous deployment (similar to [Heroku's automatic GitHub deploys](https://devcenter.heroku.com/articles/github-integration)). PyGithub allows me to get and create deployments, but doesn't seem to support deletion or deployment statuses. Without these functions, it will be more difficult to display feedback about deployments within a repository.

Are there any existing plans to add this functionality? If not, I am willing to work on this.

sfdye:
No plans yet, PR welcome

PR #1588, opened 2020-06-27
Add support for deployment statuses

colbygallup:
Fixes issue #1586


colbygallup:
@s-t-e-v-e-n-k I added the requested test. Need any other changes?

tests/DeploymentStatus.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class DeploymentStatus(Framework.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.deployment = (
+            self.g.get_user().get_repo("PyGithub").get_deployment(242997115)
+        )
+        self.status = next(
+            s for s in self.deployment.get_statuses() if s.id == 344110026
```
s-t-e-v-e-n-k, line 37: There's a handy get_status() method, and you already have the id, there's no need to loop through them.

You should also add a test for get_statuses() though -- just asserting on the returned ids would be enough, such as the Workflow.testGetRunsWithObjects unit test.

colbygallup: I added a `Deployment.get_status()` before, but removed it once I realized that it was part of a feature preview:
```python
gh.get_user().get_repo("PyGithub").get_deployment(242997115).get_status("344427441") 
  File "C:\Users\colby\PyGithub\github\Deployment.py", line 162, in get_status
    "GET", self.url + "/statuses/" + str(id_)
  File "C:\Users\colby\PyGithub\github\Requester.py", line 319, in requestJsonAndCheck
    verb, url, parameters, headers, input, self.__customConnection(url)
  File "C:\Users\colby\PyGithub\github\Requester.py", line 342, in __check
github.GithubException.GithubException: 415 {"message": "If you would like to help us test the Deployments API enhancement during its preview period, you must specify a custom media type in the 'Accept' header. Please 
see the docs for full details.", "documentation_url": "https://developer.github.com/v3/repos/deployments/#get-a-single-deployment-status"}
```

s-t-e-v-e-n-k: We already support a lot of API previews, so feel free to add it -- Repository.create_label() can give you some ideas.

colbygallup: Sounds good. I have the new test done already; I'll add the `get_status()` code back in and it'll be good for another review.

PR #1605, opened 2020-07-15
Add support for requested_reviewers on PullRequest

npmccallum:
This is helpful for detecting when someone has asked for a rereview on a
pull request review.


npmccallum:
I just realized that there is already `get_review_requests()`. So you can ignore this.

PR #1606, opened 2020-07-16
Check for updates to GitHub Actions every weekday

jlosito:
Now that dependabot is native to GitHub and the fact that this project makes use of GitHub actions, I think it would be nice if dependabot were to send automatic PR's of version bumps to outdated GitHub actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
sfdye, line 6: Good idea, probably less often?

s-t-e-v-e-n-k:
I'd be happy with weekly -- and just for our own edification, do you have an example PR of this on another project?

sfdye:
@jlosito Thank you for this great idea.

I've use dependabot.com to auto generated the PR for us in#1607. So closing this.

PR #1616, opened 2020-07-21
more flexible header splitting

emiride:
When using get_authorizations() from AuthenticatedUser headers["link"] is:
` '<https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/>; rel="deprecation"; type="text/html"' ` and thus it fails. With this approach, it will just discard `type="text/html"` part and will not affect other calls.

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -230,7 +230,7 @@ class PaginatedList(PaginatedListBase):
         if "link" in headers:
             linkHeaders = headers["link"].split(", ")
             for linkHeader in linkHeaders:
-                (url, rel) = linkHeader.split("; ")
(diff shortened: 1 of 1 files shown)


github/PaginatedList.py
```diff
@@ -230,7 +230,7 @@ def __parseLinkHeader(self, headers):
         if "link" in headers:
             linkHeaders = headers["link"].split(", ")
             for linkHeader in linkHeaders:
-                (url, rel) = linkHeader.split("; ")
+                url, rel, *rest = linkHeader.split("; ")
```
sfdye, line 233:
```suggestion
                url, rel, _ = linkHeader.split("; ")
```

Issue #1622, opened 2020-07-26
create_repo() does not handle repo names with " " (White-spaces) 


ionicc:
create_repo() right now throws an expected 
```
github.GithubException.UnknownObjectException: 404 {"message": "Not Found", "documentation_url": "https://developer.github.com/v3/repos/#get%22%7D
```
when creating repositories using `create_repo` with names than have " " (Spaces) in them.

This should be handled either by: 
* Replacing the spaces with a hyphen "-" 
* By throwing a better exception which explains the problem.

I would personally lean towards the "-" replacement as users would anyways replace the spaces with hyphens.

PR #1623, opened 2020-07-26
Handle white-spaces in name in Organization and AuthenticatedUser 

ionicc:
Closes: #1622

* Added stripping of leading and ending white spaces for in name in AuthenticatedUser and Organization
* Added replacement of white space in between the text with "-"

### DIFF
--- a/github/AuthenticatedUser.py
+++ b/github/AuthenticatedUser.py
@@ -644,6 +644,7 @@ class AuthenticatedUser(github.GithubObject.CompletableGithubObject):
         assert delete_branch_on_merge is github.GithubObject.NotSet or isinstance(
             delete_branch_on_merge, bool
         ), delete_branch_on_merge
+        name = name.lstrip().rstrip().replace(" ", "-") if " " in name else name
         post_parameters = {
             "name": name,
         }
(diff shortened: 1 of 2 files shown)


ionicc:
@s-t-e-v-e-n-k By ugly, Are you referring to the implementation? Simple replace would not deal with the leading and ending whitespaces. Also, strip would strip the spaces between the org name as well.

s-t-e-v-e-n-k:
@ionicc strip() doesn't work like that.

```
>>> foo = '    test string '
>>> foo.replace(" ", "-")
'----test-string-'
>>> foo.strip().replace(" ", "-")
'test-string'
>>> 
```

Replacing spaces with a - I can understand because spaces will get URL encoded, but why strip(), that is just working around issues that callers should be dealing with.

ionicc:
> strip() doesn't work like that.
@s-t-e-v-e-n-k  My bad on that, Didn't see the implementation of strip() before implementing lstrip and rstrip. But as you said, Giving a more descriptive exception would be good.

jesseli2002:
Question - might be a slippery slope argument, but there's a bunch of prohibited characters for a repository name on Github (most symbols, I imagine most Unicode characters too. Brief experimentation suggests that Github combines any contiguous string of these into a single dash.)

If validation is going to be done for spaces, wouldn't that imply validation should be done for all other characters?

ionicc:
@jesseli2002 True. Maybe instead of validation, Meaningful exceptions should be given rather than getting `UnknownObjectException` on any validation issue.

Issue #1624, opened 2020-07-27
Add Support for manual workflow trigger using workflow_dispatch


wrecker:
Github Actions recently added support for triggering Workflows using the `workflow_dispatch` event [Blog Post](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/)

Docs: https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event

I have a PR that adds support for this API Endpoint.

PR #1625, opened 2020-07-27
Add suport for workflow dispatch event

wrecker:
Fixes: #1624


wrecker:
One thing I was not able to verify is the ability to trigger the workflow with a commit SHA. I tried the API with curl too and still couldn't make it work. I suspect github's implementation of the API might be the problem.

github/Workflow.py
```diff
@@ -106,6 +106,31 @@ def badge_url(self):
         self._completeIfNotSet(self._badge_url)
         return self._badge_url.value
 
+    def create_dispatch(self, ref, inputs):
```
s-t-e-v-e-n-k, line 109: If inputs is not required, default it to NotSet.

github/Workflow.py
```diff
@@ -106,6 +106,31 @@ def badge_url(self):
...
+        :param inputs: dict
+        :rtype: bool
+        """
+        assert (
+            isinstance(ref, github.Branch.Branch)
+            or isinstance(ref, github.Tag.Tag)
+            or isinstance(ref, github.Commit.Commit)
+            or isinstance(ref, str)
+        ), ref
+        assert isinstance(inputs, dict) or inputs is None, inputs
```
s-t-e-v-e-n-k, line 122: Or is NotSet, not None.

github/Workflow.pyi
```diff
@@ -2,21 +2,26 @@ from datetime import datetime
...
 from github.PaginatedList import PaginatedList
+from github.Tag import Tag
 from github.WorkflowRun import WorkflowRun
 
 class Workflow(CompletableGithubObject):
     def __repr__(self) -> str: ...
     def _initAttributes(self) -> None: ...
     def _useAttributes(self, attributes: Dict[str, Any]) -> None: ...
+    def create_dispatch(
+        self, ref: Union[str, Branch, Commit, Tag], inputs: Union[dict, None]
```
s-t-e-v-e-n-k, line 17: Can you expand dict to Dict[] with some types for the keys and values?

tests/Workflow.py
```diff
@@ -81,3 +81,39 @@ def testGetRunsWithStrings(self):
...
+
+        self.assertTrue(workflow.create_dispatch(branch, dispatch_inputs))
+
+    def testCreateDispatchWithTag(self):
+        dispatch_inputs = {"logLevel": "Warning", "message": "Log Message"}
+        workflow = self.g.get_repo("wrecker/PyGithub").get_workflow(
+            "manual_dispatch.yml"
+        )
+        tags = self.g.get_repo("wrecker/PyGithub").get_tags()
+        for tag in tags:
```
s-t-e-v-e-n-k, line 102: tag = [t for t in tags if t.name == "workflow_dispatch_tag"]

tests/Workflow.py
```diff
@@ -81,3 +81,39 @@ def testGetRunsWithStrings(self):
...
+
+    def testCreateDispatchWithBranch(self):
+        dispatch_inputs = {"logLevel": "Warning", "message": "Log Message"}
+        workflow = self.g.get_repo("wrecker/PyGithub").get_workflow(
+            "manual_dispatch.yml"
+        )
+        branch = self.g.get_repo("wrecker/PyGithub").get_branch(
+            "workflow_dispatch_branch"
+        )
+
```
s-t-e-v-e-n-k, line 93: Would you mind dropping the vertical white space here and in the other tests you've added?

github/Workflow.py
wrecker: Done.

github/Workflow.py
wrecker: Done

tests/Workflow.py
wrecker: Done. Though I will add this form makes the API go and fetch all pages of tags from the repo. Its not that relevant here because we are using the recorded data file.

tests/Workflow.py
wrecker: Done.

github/Workflow.pyi
wrecker: Based on my reading of the Github docs, I am assuming that the values can be one of str, int or Float. 
I have updated it to read `inputs: Union[Dict[str, Union[str, int, float]], _NotSetType]`

wrecker:
@s-t-e-v-e-n-k Thanks for the review & approval. anything else needed from me...

Issue #1621, opened 2020-07-26
Get status of checks on a Pull Request


ChaiBapchya:
Currently, PyGithub API doesn't allow to check for status of the checks run on a PR.

For e.g. in case of a PR https://github.com/apache/incubator-mxnet/pull/18785
There are 12 status checks.
There should be an API that exposes the status of each of those 12 checks or overall status of that PR.

ChaiBapchya:
Related issue https://github.com/PyGithub/PyGithub/issues/1063

wrecker:
@s-t-e-v-e-n-k @sfdye I have made significant progress on implementing the Github Checks API. For now I have most of the READ actions implemented based on this: https://docs.github.com/en/rest/reference/checks

- [x] CheckSuite
- [x] CheckRun
- [ ] CheckRunOutput
- [ ] CheckRunAnnotation
- [ ] GithubApp

Whats your preferred way of sending PRs? One big monolith PR (with individual commits) or split PRs?

sfdye:
Of course individual PR is easier to review, but both are welcome

wrecker:
@sfdye The first of the PRs is in.

cristianmtr:
Is this available? I can't find it

ChaiBapchya:
Don't think its supported yet. There was some PR opened on aug 2 but got closed. so ya.. we're all waiting :)

Croydon:
Please re-ope this then :)

Seluj78:
Hey ! If you upgrade to `1.54`, check suite and runs are supported !

If you want an example, I use it here: https://github.com/AFPy/PyDocTeur/pull/47

ChaiBapchya:
While I can see check-suites works with 1.54
I can see that the number returned is incorrect
For example:
For this commit: https://github.com/apache/incubator-mxnet/pull/18785/commits/232538f4f4f704fd5d01b97b66655bab5393942b
There were 14 checks made.
But the result returned by the api for this specific commit is: 
```
>>> commit
Commit(sha="232538f4f4f704fd5d01b97b66655bab5393942b")
>>> for a in cs:
...      print(a)
...
CheckSuite(url="https://api.github.com/repos/apache/incubator-mxnet/check-suites/1063117829", id=1063117829)
CheckSuite(url="https://api.github.com/repos/apache/incubator-mxnet/check-suites/1064967098", id=1064967098)
```

s-t-e-v-e-n-k:
[...]

PR #1631, opened 2020-08-03
Add initial support for Github Apps.

wrecker:
Needed for #1621


sfdye:
It seems there is an isort issue, you can get around by `tox -e lint` locally

wrecker:
Ahh.. I can't `tox -e lint` because it keeps looking for python3.6. My dev environment is ubuntu20.04 which doesn't have 3.6 in its package repo. I have to look into running 3.6 with docker.

github/GithubApp.py
```diff
@@ -0,0 +1,172 @@
...
+#                                                                              #
+################################################################################
+
+import github.GithubObject
+import github.NamedUser
+
+
+class GithubApp(github.GithubObject.CompletableGithubObject):
+    """
+    This class represents check suites. The reference can be found here https://docs.github.com/en/rest/reference/apps
```
s-t-e-v-e-n-k, line 31: This class does not represent check suites.

wrecker: :grimacing: Oops!

wrecker:
@s-t-e-v-e-n-k Can I get a OK / NOT OK for this PR. my next set of PRs reference the changes in this one.

PR #1636, opened 2020-08-04
Correct **kwargs typing for search_*

### LINKED ISSUE #1633: Typing error on Github.search_issues
7b20b13d0b853e813b201edf1daf159ca8bc983f set the type of all kwargs to `Dict[str, Any]`. This is incorrect, as `repo` is `str`.

My assumption is that the author of #1517 didn't understand the the way typed **kwargs work, and meant to simply assign it to `Any`.

Mypy error message: `error: Argument "repo" to "search_issues" of "Github" has incompatible type "str"; expected "Dict[str, Any]"  [arg-type]`

s-t-e-v-e-n-k:
It turns out typing for **kwargs is always Dict[str, Type], so only the
type is required.

Fixes #1633


github/Requester.pyi
```diff
@@ -17,7 +17,7 @@ class HTTPRequestsConnectionClass:
         strict: bool = ...,
         timeout: Optional[int] = ...,
         retry: Any = ...,
-        **kwargs: Dict[str, str]
+        **kwargs: str
```
sfdye, line 20: not `Any`?

s-t-e-v-e-n-k: This is HTTP{,S}ConnectionClass, not search_* -- it only gets strings as keyword arguments. I was naughty and fixed it in the same commit, bad me.

sfdye: But isn't this a Dict?

PR #1643, opened 2020-08-05
Fix Repository.create_repository_dispatch type signature

christopher-dG:
Not really sure why I initially thought this had to be `str -> str`. The endpoint accepts anything that can be encoded into JSON. Technically it accepts numbers, bools, and None as keys too, so perhaps the key type should be a Union... but 99% of people will only use strings so maybe this is ok.


s-t-e-v-e-n-k:
Probably 'Any', rather than 'object'

PR #1658, opened 2020-08-17
Add method get_team_membership for user 

talmachani:
Implemented the get_team_membership with tests.


github/Team.pyi
```diff
@@ -17,6 +18,7 @@ class Team(CompletableGithubObject):
         self, member: NamedUser, role: Union[str, _NotSetType] = ...
     ) -> None: ...
     def add_to_members(self, member: NamedUser) -> None: ...
+    def get_team_membership(self, member) -> Membership: ...
```
s-t-e-v-e-n-k, line 21: You need to specify the type of member too.

tests/Team.py
```diff
@@ -105,6 +105,20 @@ def testMembers(self):
         self.assertRaises(AssertionError, self.team.add_membership, user, "admin")
         self.team.remove_membership(user)
 
+    def testTeamMembership(self):
+        user = self.g.get_user("jacquev6")
+        self.assertListKeyEqual(self.team.get_members(), None, [])
+        self.assertFalse(self.team.has_in_members(user))
+        self.team.add_to_members(user)
```
s-t-e-v-e-n-k, line 112: add_to_members() API call is deprecated. Use `add_membership` instead.

tests/Team.py
```diff
@@ -105,6 +105,20 @@ def testMembers(self):
         self.assertRaises(AssertionError, self.team.add_membership, user, "admin")
         self.team.remove_membership(user)
 
+    def testTeamMembership(self):
+        user = self.g.get_user("jacquev6")
+        self.assertListKeyEqual(self.team.get_members(), None, [])
```
s-t-e-v-e-n-k, line 110: This feels awkward, just:

``self.assertEqual(list(self.team.get_members()), [])``

github/Team.pyi
talmachani: Add Union[str,NamedUser]

tests/Team.py
talmachani: changed both in test and ReplayData

tests/Team.py
talmachani: done.

PR #1667, opened 2020-08-21
Add link to more examples

seansan:
really helpful ;)


s-t-e-v-e-n-k:
We already upload the examples to Readthedocs, so maybe we should link there? https://pygithub.readthedocs.io/en/latest/examples.html

PR #1676, opened 2020-08-27
Issue 1655: Add authorize credentials list and removal on organization under SAML

AlexandreODelisle:
The objective is to implement the two available methods to list and remove credential ids linked to an Organization that have been Authorize by the user to act as authentication on the Organization.

Added the two endpoints available under the Organization on the GitHub API
https://developer.github.com/v3/orgs/#list-saml-sso-authorizations-for-an-organization
https://developer.github.com/v3/orgs/#remove-a-saml-sso-authorization-for-an-organization

Also needed to create a new Object to handle the return payload of the API since the raw data differs from the Authorization Object..

Added one unit test to list the credentials under an Organization, but I do not have a Organization that is SAML protected that I can use to display in the unittest.. So I added on a personnal organization with no SAML, so an empty List is validated.


github/AuthorizationOrganization.pyi
```diff
@@ -0,0 +1,21 @@
...
+    @property
+    def credential_type(self) -> str: ...
+    @property
+    def credential_authorized_at(self) -> str: ...
+    @property
+    def credential_accessed_at(self) -> str: ...
+    @property
+    def token_last_eight(self) -> str: ...
+    @property
+    def scopes(self) -> list: ...
```
s-t-e-v-e-n-k, line 21: List[] including the types, such as List[str], please.

tests/Organization1655.py
```diff
@@ -0,0 +1,35 @@
...
+#                                                                              #
+# You should have received a copy of the GNU Lesser General Public License     #
+# along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
+#                                                                              #
+################################################################################
+
+from . import Framework
+
+
+class Organization1655(Framework.TestCase):
```
s-t-e-v-e-n-k, line 28: Can we have a better name, such as OrganizationAuthorization?

tests/Organization1655.py
```diff
@@ -0,0 +1,35 @@
...
+    def setUp(self):
+        super().setUp()
+        self.org = self.g.get_organization("ThoughtCamera")
+
+    def testGetCredentialsAuth(self):
+        credentials_list = self.org.get_credential_authorizations()
+        self.assertListEqual(credentials_list, [])
```
s-t-e-v-e-n-k, line 35: You also need to test delete, but I'd also prefer we return something so we can verify the returned attributes.

tests/Organization1655.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+############################ Copyrights and license ############################
+#                                                                              #
+# Copyright 2020 Alexandre Delisle <alexodelisle  @gmail.com>                  #
```
s-t-e-v-e-n-k, line 5: space in the email address.

s-t-e-v-e-n-k (changes requested):
It's a good start, but requires more before merging. Thanks!

tests/Organization1655.py
AlexandreODelisle: I will need to wait in that case, since the only Organization I have available at the moment,with SAML SSO, is private. I can not display the credentials list in the --record. I will ask GitHub sales for a trial for Entreprise Org hosted on GitHub.

AlexandreODelisle:
@s-t-e-v-e-n-k , finally got the coverage to pass. I think I got all the reviews, since there is now test for the Attributes and __repr__, also got an Enterprise Team so the credential list also has non redacted information.
Thanks

github/Organization.py
```diff
@@ -1156,6 +1157,33 @@ def get_migrations(self):
             headers={"Accept": Consts.mediaTypeMigrationPreview},
         )
 
+    def get_credential_authorizations(self):
+        """
+        :calls: `GET /orgs/:org/credential-authorizations <https://developer.github.com/v3/orgs>`_
+        :rtype: list
```
s-t-e-v-e-n-k, line 1163: list of :class:`github.AuthorizationOrganization.AuthorizationOrganization`_

(I'm not certain of the formatting, but roughly that, please)

github/Organization.pyi
```diff
@@ -204,3 +204,7 @@ class Organization(CompletableGithubObject):
     def updated_at(self) -> datetime: ...
     @property
     def url(self) -> str: ...
+    def remove_credential_authorization(
+        self,
+        credential_id: Union[int, _NotSetType] = ...,
+    ) -> None: ...
```
s-t-e-v-e-n-k, line 210: The return type is not None.

github/AuthorizationOrganization.py
```diff
@@ -0,0 +1,126 @@
...
+# along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
+#                                                                              #
+################################################################################
+
+import github.GithubObject
+
+
+class AuthorizationOrganization(github.GithubObject.NonCompletableGithubObject):
+    """
+    This class represents a credential id.
```
s-t-e-v-e-n-k, line 30: I'd like more detail here, along with a link to the documentation of the object.

s-t-e-v-e-n-k (changes requested):
This is looking like a great start, I have some concerns.

alexef:
Any chance this gets some traction?

EnricoMi:
@AlexandreODelisle this is awesome, can we get this over the finish line? Happy to approve once review comments are resolved.

alambike:
Is it possible to do something with this to resume this pull request and be able to have this functionality in PyGithub? Is it correct to create a new PR with these changes to be able to work on it? cc/ @EnricoMi , @AlexandreODelisle

AlexandreODelisle:
Please feel free to go ahead with it.

Personally, I am no longer using SAML SSO on a regular basis for authentication, and also another py package.

I am sorry for the delay in response.

lionas32:
Picked this up here: https://github.com/PyGithub/PyGithub/pull/3392

PR #1679, opened 2020-09-02
Format with new black

s-t-e-v-e-n-k:
Black is very opinionated, but sometimes those opinions change. Run the
new black version across the codebase, and lock down the version used by
pre-commit so we don't get surprised like this again.


sfdye:
@s-t-e-v-e-n-k Could you add this commit to https://github.com/PyGithub/PyGithub/blob/master/.git-blame-ignore-revs?

PR #1682, opened 2020-09-04
Added support for `Repository.get_workflow_runs` parameters

zacker150:
The `/repos/:owner/:repo/actions/runs`  endpoint has 4 optional parameters which can be used to narrow down the search results. 

> Name | Type | Description
> -- | -- | --
> actor | string | Returns someone's workflow runs. Use the login for the user who created the push associated with the check suite or workflow run.
> branch | string | Returns workflow runs associated with a branch. Use the name of the branch of the push.
> event | string | Returns workflow run triggered by the event you specify. For example, push, pull_request or issue. For more information, see "Events that trigger workflows" in the GitHub Help documentation.
> status | string | Returns workflow runs associated with the check run status or conclusion you specify. For example, a conclusion can be success or a status can be completed. For more information, see the status and conclusion options available in "Create a check run."

This PR adds support for those parameters


s-t-e-v-e-n-k (changes requested):
This is looking like a great start! You don't need a setUp() method if you don't perform any setup. Further more, this could be more easily expressed as 3 test methods -- calling get_workflow_runs() with no arguments, calling it with all suitable object arguments and calling it with all string arguments. I'd also recommend picking only one repo and setting it to self.repo in setUp just to stop repeating code so much.

zacker150:
I've consolidated the unit tests as requested.

s-t-e-v-e-n-k (approved):

Issue #1615, opened 2020-07-20
get_user does not support retrieving user information by ID


natanlao:
I'm writing a script that interfaces with both ZenHub and GitHub (using PyGithub). ZenHub only identifies users by their user id (databaseId), so if I want to find the name of the user associated with an id, I have to query GitHub's API. There's an 
[undocumented endpoint](https://stackoverflow.com/questions/11976393/get-github-username-by-id) for this (`/user/:id`). 

A quick example. (My user id is 23219746.) I'm on `PyGithub==1.43.5` but I can see this is still the case on v1.51.

```pycon
>>> import github
>>> g = github.Github("...")
>>> g.get_user(g.get_user().id)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/natan/DataBiosphere/azul/.venv/lib/python3.8/site-packages/github/MainClass.py", line 226, in get_user
    assert login is github.GithubObject.NotSet or isinstance(login, str), login
AssertionError: 23219746
```

Does PyGithub support undocumented endpoints? If so, it seems trivial to add support for this:

```diff
diff --git a/github/MainClass.py b/github/MainClass.py
index cadd1ff5..b450492a 100644
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -261,18 +261,25 @@ class Github(object):
         :param login: string
         :rtype: :class:`github.NamedUser.NamedUser` or :class:`github.AuthenticatedUser.AuthenticatedUser`
         """
-        assert login is github.GithubObject.NotSet or isinstance(login, str), login
+        assert login is github.GithubObject.NotSet or isinstance(login, str) or isinstance(login, int), login
         if login is github.GithubObject.NotSet:
             return AuthenticatedUser.AuthenticatedUser(
                 self.__requester, {}, {"url": "/user"}, completed=False
             )
-        else:
+        elif isinstance(login, str):
             headers, data = self.__requester.requestJsonAndCheck(
                 "GET", "/users/" + login
             )
             return github.NamedUser.NamedUser(
                 self.__requester, headers, data, completed=True
             )
+        else:
+            headers, data = self.__requester.requestJsonAndCheck(
+                "GET", "/user/" + login
+            )
+            return github.NamedUser.NamedUser(
+                self.__requester, headers, data, completed=True
+            )
 
     def get_users(self, since=github.GithubObject.NotSet):
         """
```

s-t-e-v-e-n-k:
[...]

PR #1691, opened 2020-09-12
Added a method for getting a user by their id

Brikaa:
This addresses issue #1615; as @s-t-e-v-e-n-k has said, I have made a new method in the Github class called get_user_by_id() which takes in the integer user id as a parameter.


bachp (approved):
Useful functionallity to have.

s-t-e-v-e-n-k (approved):

PR #1669, opened 2020-08-21
Add parent_team_id support

alexnorell:
I noticed that `Organization.create_team()` and `Team.edit()` didn't have the ability to set the parent team ID.

I haven't added any tests, but I can if wanted.

PR #1696, opened 2020-09-20
Create Team with a Team Parent and delete Team from Organization object

AlexandreODelisle:
- Making it possible to create a Team and bind it to a Parent Team.
- Also adding the possiblity to delete a Team from the Organization object instead of using only the Team Object to delete a Team.

Since the first initiative is an available option in the GitHub API, I added the parameter to the create_team() function of the Organization object, the parent team being an _int_ changed in _str_ at the url  build time. 

For the second, added the remove function like the get_team*() of the Organization object, where we can remove by name or team_id


AlexandreODelisle:
Oups, duplicate of https://github.com/PyGithub/PyGithub/pull/1669

s-t-e-v-e-n-k:
Closing as a duplicate of #1669, thanks for your contribution!

Issue #1421, opened 2020-03-06
Support new "Delete Reactions" endpoints


wesleimp:
GitHub Developer API announcement:
https://developer.github.com/changes/2020-02-26-new-delete-reactions-endpoints/

These new endpoints need to be supported:

- [ ] Delete a commit comment reaction
  - `DELETE /repos/:owner/:repo/comments/:comment_id/reactions/:reaction_id`
  - Docs: https://developer.github.com/v3/reactions/#delete-a-commit-comment-reaction

- [ ] Delete an issue reaction
   - `DELETE /repos/:owner/:repo/issues/:issue_number/reactions/:reaction_id`
   - Docs: https://developer.github.com/v3/reactions/#delete-an-issue-reaction

- [ ] Delete a reaction to a commit comment
  - `DELETE /repos/:owner/:repo/issues/comments/:comment_id/reactions/:reaction_id`
  - Docs: https://developer.github.com/v3/reactions/#delete-an-issue-comment-reaction

- [ ] Delete a pull request comment reaction
  - `DELETE /repos/:owner/:repo/pulls/comments/:comment_id/reactions/:reaction_id`
  - Docs: https://developer.github.com/v3/reactions/#delete-a-pull-request-comment-reaction

- [ ] Delete team discussion reaction
  - `DELETE /orgs/:org/teams/:team_slug/discussions/:discussion_number/reactions/:reaction_id`
  - Docs: https://developer.github.com/v3/reactions/#delete-team-discussion-reaction

- [ ] Delete team discussion comment reaction
  - `DELETE /orgs/:org/teams/:team_slug/discussions/:discussion_number/comments/:comment_number/reactions/:reaction_id`
  - Docs: https://developer.github.com/v3/reactions/#delete-team-discussion-comment-reaction

gjcairo:
Hey, I will work on this.

s-t-e-v-e-n-k:
Closed via merging #1708

PR #1708, opened 2020-09-28
Add support for deleting reactions

changhc:
Implements #1421 .

Note that only the following endpoints are implemented:
* DELETE /repos/:owner/:repo/comments/:comment_id/reactions/:reaction_id
* DELETE /repos/:owner/:repo/issues/:issue_number/reactions/:reaction_id
* DELETE /repos/:owner/:repo/issues/comments/:comment_id/reactions/:reaction_id
* DELETE /repos/:owner/:repo/pulls/comments/:comment_id/reactions/:reaction_id

For team-related endpoints, all reaction-related endpoints are not yet built according to #1249 . They'll be implemented together if reactions for teams are prioritised. I think the issue mentioned in #1249 no longer exists. The schema looks identical to `github.Reaction.Reaction`. Please comment on this part.


github/PullRequestComment.py
```diff
@@ -223,6 +224,22 @@ def create_reaction(self, reaction_type):
         )
         return github.Reaction.Reaction(self._requester, headers, data, completed=True)
 
+    def delete_reaction(self, reaction_id):
+        """
+        :calls: `DELETE /repos/:owner/:repo/pulls/comments/:comment_id/reactions/:reaction_id
+                <https://developer.github.com/v3/reactions/#delete-a-pull-request-comment-reaction>`_
+        :param reaction_id: integer
+        :rtype: :class:`github.Reaction.Reaction`
```
s-t-e-v-e-n-k, line 232: This is not correct.

github/CommitComment.py
```diff
@@ -189,6 +190,21 @@ def create_reaction(self, reaction_type):
...
+        :param reaction_id: integer
+        :rtype: None
+        """
+        assert isinstance(reaction_id, int), reaction_id
+        headers, data = self._requester.requestJsonAndCheck(
+            "DELETE",
+            self.url + "/reactions/" + str(reaction_id),
+            headers={"Accept": Consts.mediaTypeReactionsPreview},
+        )
+        return None
```
s-t-e-v-e-n-k, line 206: Most delete actions check if the status returned is 204. Can you do that here?

tests/Issue.py
```diff
@@ -263,6 +264,9 @@ def testCreateReaction(self):
         self.assertEqual(reaction.id, 16917472)
         self.assertEqual(reaction.content, "hooray")
 
+    def testDeleteReaction(self):
+        self.issue.delete_reaction(85740167)
```
s-t-e-v-e-n-k, line 268: With the above change, this can now be `self.assertTrue(self.issue.delete_reaction(85740167))`

s-t-e-v-e-n-k (changes requested):
This is looking great, just some small nitpicks.

github/CommitComment.py
changhc: I can do this by calling `requestJson` instead. I actually followed the `delete` method of this class defined in previous lines. I guess the code is a bit old so that the method doesn't check the returned status. Do you think we should take them as tech debt and update these methods as well (in a separate PR)?

tests/Issue.py
pascal-hofmann: Code was updated accordingly. Resolving this conversation.

s-t-e-v-e-n-k (approved):

Issue #1127, opened 2019-05-26
Mock for testing


ahmad88me:
Is there a way to fake GitHub APIs like some kind of a mock? 
something like 
```
g = Github(fake=True)
repo = g.get_repo('non-existing/repo', default={'name': 'repo', ...})
```

nuarhu:
Any information on this? Or maybe a bit more inline documentation in the test classes could already help to decide whether they can be used for such tests.

ahmad88me:
@nuarhu actually the test cases use the actual github APIs. This mock suggestions wouldn't call GitHub APIs.

steveoh:
It would be great to be able to test using some classes without calling actual api's. I'm trying to create a new issue and it is not very intuitive.

ChaiBapchya:
@ahmad88me did you get a workaround for mocking Github APIs? I scoured the internet but to no avail.

ahmad88me:
@ChaiBapchya I didn't get around it yet. I will try to do so soon.

ChaiBapchya:
Sure thanks. @ahmad88me any update or any pointers?

ahmad88me:
@ChaiBapchya  
@steveoh 
@nuarhu 

If you are still interested in the mock option, I've created this pull request.

https://github.com/PyGithub/PyGithub/pull/1720

kwk:
Hello. I'm very interested in this for many reasons:

* I want to run tests without having access to github, e.g. when working offline.
* I want to be able to create "golden files". That is a mode where you actually call the real Github API and store the results for later use in tests. This is helpful in case the API changes.
* I don't want to pass along credentials just to run tests on github.

[...]

PR #1720, opened 2020-10-14
Add Mock option instead of calling the live API. 

ahmad88me:
This is related to issue https://github.com/PyGithub/PyGithub/issues/1127
I've added this sample. What do you think?

The idea is to pass a dictionary of the API response. 
```github.Github(self.oauth_token, retry=self.retry, mock=mock_dict)```

Further steps would be to simplify  the `mock_dict` further by providing default attributes values. A see an example of `mock_dict` see the file `/tests/FrameworkMock.py` and `tests/Mock.py`


s-t-e-v-e-n-k (changes requested):
But we already do mock responses? That's what the replay data in the test machinery is for. What is the problem is you're trying to solve with this?

ahmad88me:
@s-t-e-v-e-n-k  From my understanding, the mock responses are text files used for the automated tests for PyGithub. The mock here is to be used by the users for PyGithub to simulate the different responses/cases. Let me know what do you think.
Best,

s-t-e-v-e-n-k:
But why do we need to do that? If you want to mock objects, you can do it directly without having to call into this library.

ahmad88me:
@s-t-e-v-e-n-k For testing. I have a project which relies on PyGithub. And, I don't want to run the tests on the live Github when I am doing the tests. So the mock makes is easier. Basically, I can test everything in my project without changing anything in the code (except for the part where I pass the mock). This also seems relevant for others (see https://github.com/PyGithub/PyGithub/issues/1127). What do you think?

s-t-e-v-e-n-k:
Okay, I see, but carrying around more code in this project to make mocking easier, I'm not going to do. I'd strongly suggest you look at https://docs.python.org/3/library/unittest.mock.html

matks:
Hi, I just wanted to say I have the exact same need:

> I have a project which relies on PyGithub. And, I don't want to run the tests on the live Github when I am doing the tests. So the mock makes is easier.

I also understand "but carrying around more code in this project to make mocking easier" 👍 that's a relevant statement. On the other hand I see more and more API SDK / API Client provide an official (and usually easy to setup) dedicated mock for the API they are wrapping. See for example [Guzzle PHP Client tools for mocking](https://docs.guzzlephp.org/en/stable/testing.html)

So indeed it's not a must-have for PyGithub but it would be a nice-to-have 😉 to promote the best practice that is ... to test one's code 😄 

Another way to address this need could be a basic how-to-mock-the-client section on https://pygithub.readthedocs.io/ . I am actually learning Python so I am looking for such resources.

ahmad88me:
@matks Thanks for your comment. As for now, I will be maintaining my fork (with the mock) as it is important for my project. Just in case you need to use it.

Issue #1556, opened 2020-06-11
Possibility of boilerplate reduction


JacekPliszka:
I wonder if it was possible to reduce boilerplate... something like this:

```
class Artifact(
    github.GithubObject.CompletableGithubObject,
    metaclass=github.GithubObject.CompletableGithubObjectMeta
):
    """
    This class represents Artifacts.
    The reference can be found here https://developer.github.com/v3/actions/artifacts/
    """
    _ATTRIBUTES = {
        "id": "Int",
        "node_id": "String",
        "name": "String",
        "size_in_bytes": "Int",
        "url": "String",
        "archive_download_url": "String",
        "expired": "Bool",
        "created_at": "Datetime",
        "expires_at": "Datetime"
    }
```
Then
```
class CompletableGithubObjectMeta(type):
    def __new__(cls, name, bases, dct):
        instance = super().__new__(cls, name, bases, dct)
        for attr in dct.get('_ATTRIBUTES') or []:
            def fget(self, attr_=attr):
                _attr = getattr(self, '_' + attr_)
                self._completeIfNotSet(_attr)
                return _attr.value
            setattr(instance, attr, property(fget))
        return instance
```
and
```
class GithubObject(object):
    def _initAttributes(self):
        for attribute in self._ATTRIBUTES:
            setattr(self, '_' + attribute, NotSet)

    def _useAttributes(self, attributes):
        for attribute, type_ in self._ATTRIBUTES.items():
            if attribute in attributes: # pragma no branch
                make_method = getattr(self, f'_make{type_}Attribute')
                setattr(self, '_' + attribute, make_method(attributes[attribute]))
```

And some form of .pyi autogeneration.

What do you think?

JacekPliszka:
this is Artifacts implementation 

https://github.com/JacekPliszka/PyGithub/pull/1

s-t-e-v-e-n-k:
The boilerplate reduction is an interesting idea, but I think it also loses us the docstrings mentioning type of the attributes, which is important.

JacekPliszka:
docstring is not a problem as you can modify it in meta.

```
    _ATTRIBUTES = {
        "id": ("Int", "doctring for id"),
        "node_id": ("String","docstring for node_id"),
        "name": ("String", "docstring for name"),
...
    }
```
The only difference is that you need to import the class for it to work (sphinx does import).
But there is pyi which can be generated.

So the question is if there is any use of the .py file that is not importing it?

[...]

Issue #1373, opened 2020-01-29
GitHub Actions support


gschaffner:
Docs here: https://developer.github.com/v3/actions/

Are there plans to add support for GitHub Actions? The API is still in beta, but I would suspect that it won't change much in the future.

parzhitsky:
Isn't Actions still in beta?

gschaffner:
Looks like the beta ended in November [[source](https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/)]. The API is definitely still in beta though.

sfdye:
https://developer.github.com/changes/2020-03-23-ActionsGA/

sfdye:
Pin this issue to get more attention

s-t-e-v-e-n-k:
I won't split this issue up into six, but documenting progress as I get to them.

- [ ] Artifacts
- [ ] Secrets
- [ ] Self-hosted runners
- [ ] Workflows
- [ ] Workflow jobs
- [ ] Workflow runs

cclauss:
#1456 and #1457

marns93:
Are there any plans, when the Actions API will be supported?

sfdye:
Not yet, but we are open to any granular PR to implement the checklist items ^

s-t-e-v-e-n-k:
I have a local branch for Workflows, it requires some polish, tests and more importantly some free time.

sfdye:
I can take `Secrets` when I have some time.

3wnbr1:
Is somebody already working on Artifacts ? Otherwise I can take it

sfdye:
@3wnbr1 Feel free

s-t-e-v-e-n-k:
I have a local branch that needs some more love for ``Workflow Runs``.

JacekPliszka:
@3wnbr1 what is the status of your artifacts?

I have list of artifacts and downloading artifact working but I have no tests and also my implementation need modification.

salomvary:
@s-t-e-v-e-n-k [`WorkflowRun`](https://github.com/PyGithub/PyGithub/blob/master/github/WorkflowRun.py) seems to be missing `workflow_id` attribute: https://docs.github.com/en/rest/reference/actions#list-workflow-runs-for-a-repository

salomvary:
Another small thing: `repo.get_workflow(id_or_name)` fails if the id is `int`:

```
    "GET", self.url + "/actions/workflows/" + id_or_name
[...]

PR #1734, opened 2020-11-03
Support for artifacts API

YannickJadoul:
This builds further on @JacekPliszka's work, but without the metaclasses to automatically add fields to a new class (cfr. #1556; which seems interesting, but unrelated to this actual functionality). See also #1373.

This still needs tests, but I cannot immediately find any artifacts belonging to this repository. Any idea what I can use to test downloading artifacts?

Things to review/consider:
- Following the status "302 Found" redirect, needed during the download of artifacts, is handled manually, since existing tests break when adding this to `Requester.__requestRaw` (see @JacekPliszka's f69c763, and my a7fd8f2). Is this something that should be preferably be added to `Requester.__requestRaw`? I can do an effort to fix those failing tests, then.
- I had to refactor the methods of `Requester` quite a bit, in order to be able to download a binary file without trying to decode it to a unicode string. While this seems like a lot of changes, I think they should be reasonably focused. See 99bbfbe and 78e7ae3.
- https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#artifacts mentions that setting "`accept: application/vnd.github.v3+json` is recommended", yet this is not done for querying JSON in general. Should I do this for the new methods, or is this a separate PR to get this done for everything?


github/Artifact.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+
+        return data
+
+    def delete(self):
+        """
+        :calls: `DELETE /repos/:owner/:repo/actions/artifacts/:artifact_id <https://developer.github.com/v3/actions/artifacts/#delete-an-artifact>`_
+        :rtype: None
+        """
+        headers, data = self._requester.requestJsonAndCheck("DELETE", self.url)
```
s-t-e-v-e-n-k, line 137: requestJson() and also check the returned status code is 204, like other examples in the code base.

github/Artifact.pyi
```diff
@@ -0,0 +1,31 @@
...
+from typing import Any, Dict, Union
+
+from github.GithubObject import CompletableGithubObject
+
+class Artifact(CompletableGithubObject):
+    def __repr__(self) -> str: ...
+    def _initAttributes(self) -> None: ...
+    def _useAttributes(self, attributes: Dict[str, Any]) -> None: ...
+    def download(self, archive_format: str) -> bytes: ...
+    def delete(self) -> None: ...
```
s-t-e-v-e-n-k, line 11: The above means this turns to bool.

github/WorkflowRun.py
```diff
@@ -92,6 +95,14 @@ def conclusion(self):
         self._completeIfNotSet(self._conclusion)
         return self._conclusion.value
 
+    @property
+    def workflow_id(self):
```
s-t-e-v-e-n-k, line 99: This is unrelated to the Artifacts API, I'd prefer it was split out to its own tiny PR.

s-t-e-v-e-n-k (changes requested):
This is looking like a great start, however the decode changes are incredibly far reaching, more so than I would have thought, you can't use __requestRaw? I have some other concerns inline.

github/WorkflowRun.py
YannickJadoul: Good point. Done: #1737

github/Artifact.py
YannickJadoul: OK, sure. FYI, most of the things I found when I looked for `DELETE` were of this form, though? E.g.:

https://github.com/PyGithub/PyGithub/blob/1bb18ab52e9a5997774c9aed4e1521fe382e5b21/github/PullRequestComment.py#L172-L177

So that's probably why I ended up with this.

YannickJadoul:
> This is looking like a great start, however the decode changes are incredibly far reaching, more so than I would have thought, you can't use __requestRaw?

I've tried to clean up a little bit further, yes. However, the main problem is that accessing the `requests.Response.text` attribute (inside `RequestsResonse`, see [here](https://github.com/PyGithub/PyGithub/blob/1bb18ab52e9a5997774c9aed4e1521fe382e5b21/github/Requester.py#L75)) tries to decode to Unicode (see [here in the Requests docs](https://requests.readthedocs.io/en/latest/api/#requests.Response.text)) and fails in the case of a zip file. To get `bytes`, we need to access [`requests.Response.content`](https://requests.readthedocs.io/en/master/api/#requests.Response.content). So we cannot directly use `__requestRaw` as it is now.

github/Artifact.py
YannickJadoul: I have changed it, at any rate; see 8ec83c32fa0be91d539a33803b0bddef99528f6d

github/Artifact.py
```diff
@@ -0,0 +1,174 @@
...
+#                                                                              #
+# You should have received a copy of the GNU Lesser General Public License     #
+# along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
+#                                                                              #
+################################################################################
+
+import github.GithubObject
+
+
+class Artifact(github.GithubObject.CompletableGithubObject):
```
yixinguo2, line 29: Hi, before seeing your PR, I also did something similar and [created a PR](https://github.com/PyGithub/PyGithub/pull/1985). Please take a look and see if there are any useful parts that you could incorporate like some of the tests...

bauergeorg:
Could you reopen this pull request?

PR #1738, opened 2020-11-04
Remove older MainClass.get_installation()

s-t-e-v-e-n-k:
There are two get_installation() methods in MainClass, with different
arguments. get_installation(owner, repo) is documented and tested, and
get_installation(id) is neither. Remove it to avoid confusion.


pascal-hofmann (approved):
LGTM. Should we document the removal?

s-t-e-v-e-n-k:
I will do so in the changelog when I make a release.

PR #1743, opened 2020-11-04
Fix #1742 - incorrect typehint for `Installation.id`

### LINKED ISSUE #1742: `Installation.id` type hint incorrect
```python
def id(self) -> _ValuedAttribute: ...
```
should be an int according to the Github docs
```python
def id(self) -> int: ...
```

huwcbjones:
Noticed this whilst I was playing with the GithubApp workflows and wondered why mypy was complaining.

`GithubIntegration.get_access_token` takes the `Installation.id` which is an int


github/Installation.pyi
```diff
@@ -8,5 +8,5 @@ class Installation(NonCompletableGithubObject):
     def _initAttributes(self) -> None: ...
     def _useAttributes(self, attributes: Dict[str, Any]) -> None: ...
     @property
-    def id(self) -> _ValuedAttribute: ...
```
s-t-e-v-e-n-k, line 11: You can drop the import on line 3 too, please.

huwcbjones: Ta, missed that

s-t-e-v-e-n-k (approved):

Issue #1746, opened 2020-11-05
Support repository transfer


ryannowarm:
Please add support for [repo transfer](https://developer.github.com/v3/repos/#transfer-a-repository)

tubaman:
Here's an initial PR: https://github.com/PyGithub/PyGithub/pull/1749

falkoner:
Any chance to reopen this request? Still not addressed.

EnricoMi:
Sure, contribution welcome! Easiest might be to pick up #1749 and go from there.

PR #1749, opened 2020-11-10
add transfer method

tubaman:


github/Repository.py
```diff
@@ -1393,6 +1393,28 @@ def __create_pull(self, **kwds):
             self._requester, headers, data, completed=True
         )
 
+    def create_transfer_request(
+        self, new_owner, teams=github.GithubObject.NotSet
+    ):
+        """
+        :calls: POST /repos/:owner/:repo/transfer <https://developer.github.com/v3/repos/#transfer-a-repository>
+        :param new_owner: string
```
s-t-e-v-e-n-k, line 1401: I mean, we could also have this accept an Organization, or a NamedUser ...

github/Repository.py
```diff
@@ -1393,6 +1393,28 @@ def __create_pull(self, **kwds):
...
+        :param teams: list
```
s-t-e-v-e-n-k, line 1402: A list of what?

github/Repository.py
```diff
@@ -1393,6 +1393,28 @@ def __create_pull(self, **kwds):
             self._requester, headers, data, completed=True
         )
 
+    def create_transfer_request(
```
s-t-e-v-e-n-k, line 1396: This doesn't match the return type then? Either this creates a request to move the repository, which will happen at some point in the future, and the user then needs to go and use get_repo on the new name, or this should be called transfer and return the new repository object.

s-t-e-v-e-n-k (changes requested):
I have some concerns about the code changes, but it's looking like a good first step. This also requires tests and the method signature added to Repository.pyi for type checking

tubaman:
Agreed with all the comments above.  Unfortunately I don't have time to work this PR any further.  Just seemed like functionality and I wanted to [raise that issue](https://github.com/PyGithub/PyGithub/issues/1746) and perhaps give someone a head start with this PR

PR #1757, opened 2020-11-19
Pin requests to <2.25 as well

s-t-e-v-e-n-k:
requests 2.26 and its requirements break httpretty (see
gabrielfalcao/HTTPretty#409), so pin for the time being. Thanks to
Sébastien Besson and Dhruv Manilawala for the debugging assistance.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-requests>=2.14.0
+requests>=2.14.0,<2.25
 pyjwt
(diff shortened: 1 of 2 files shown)


gabrielfalcao:
Hello folks, thanks for your patience and to @vphilippon for fixing this. 
I just released HTTPretty 1.0.3 with the fix.

PR #1763, opened 2020-11-25
Revert "Pin requests to <2.25 as well (#1757)"

sbesson:
This reverts commit d159425f36dc7f68766cc980e262e9287e5f111c.

The dependency pinning was primarily motivated by the breakage of `httpretty` by the last minor release of `requests`. As mentioned in https://github.com/gabrielfalcao/HTTPretty/issues/409#issuecomment-733313173, the latest patch release of `httpretty` should restore compatibility with the new semantics of `requests 2.25.0`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ if __name__ == "__main__":
             "Topic :: Software Development",
         ],
         python_requires=">=3.5",
-        install_requires=["deprecated", "pyjwt", "requests>=2.14.0,<2.25"],
+        install_requires=["deprecated", "pyjwt", "requests>=2.14.0"],
         extras_require={"integrations": ["cryptography"]},
         tests_require=["cryptography", "httpretty>=0.9.6"],
     )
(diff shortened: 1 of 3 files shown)


s-t-e-v-e-n-k:
We should probably also bump httpretty to >= 1.0.3 just to sure here too.

sbesson:
998c906 should set the minimum version of `httpretty` for the tests

setup.py
```diff
@@ -100,7 +100,7 @@
             "Topic :: Software Development",
         ],
         python_requires=">=3.5",
-        install_requires=["deprecated", "pyjwt", "requests>=2.14.0,<2.25"],
+        install_requires=["deprecated", "pyjwt", "requests>=2.14.0"],
         extras_require={"integrations": ["cryptography"]},
         tests_require=["cryptography", "httpretty>=0.9.6"],
```
s-t-e-v-e-n-k, line 105: This doesn't match test-requirements

sbesson: I hadn't realized the test requirements were duplicated in two places. Should be fixed by 91098f6

sfdye (approved):

akosthekiss:
Hi, is there any chance to have a release that contains this change? (The latest release contains the pinning and gives some headaches. Not that it cannot be worked around. Just asking.)

sfdye:
I think we could do a patch release, what do you think? @s-t-e-v-e-n-k

tnguyen14:
Hi, when will this change be released?

PR #1646, opened 2020-08-09
Add Support for Check Suites

wrecker:
~~NOTE: This PR includes changes from PR #1631~~ Cleaned up.

PR #1764, opened 2020-11-25
[Rebase] Add support for Check Suites

dhruvmanila:
Continuing on the work done by @wrecker in #1646 

As mentioned in the latest review: https://github.com/PyGithub/PyGithub/pull/1646#pullrequestreview-517322517
- [x] Properties are 'type' not 'rtype'
- [x] This actually calls a URL, so this needs more detail.
- [x] Fixed merge conflicts.

***NOTE: I will add other CheckSuite API endpoints in a couple of days.***

@sbesson Should the other endpoints be included in this PR and thus, in the next release?


dhruvmanila:
It also seems that the Checks API graduated from their preview period. I shall remove the preview headers soon.

sbesson:
@dhruvmanila thanks for doing this, I have quickly tested this PR in the context of our library and at least I can confirm all the check suites endpoints we would need on our side are present and functional. 💯 

As per the other endpoints, I think you referred to [list check runs in a check suite](https://docs.github.com/en/free-pro-team@latest/rest/reference/checks#list-check-runs-in-a-check-suite) in the previous PR which makes sense. Is there anything else?

dhruvmanila:
This one [creates a check suite](https://docs.github.com/en/free-pro-team@latest/rest/reference/checks#create-a-check-suite) and this one [updates repository preferences for check suites](https://docs.github.com/en/free-pro-team@latest/rest/reference/checks#update-repository-preferences-for-check-suites) and the last one you mentioned.

By default, GitHub automatically creates a check suite each time code is pushed to a repository, changing the preferences for each app allows you to be in control of when you want the check suite to be created. I don't know where this could be useful.

github/CheckSuite.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        self._completeIfNotSet(self._url)
+        return self._url.value
+
+    def rerequest(self):
+        """
+        :calls: `POST /repos/:owner/:repo/check-suites/:check_suite_id/rerequest <https://docs.github.com/en/rest/reference/checks#rerequest-a-check-suite>`_
+        :rtype: bool
+        """
+        request_headers = {"Accept": "application/vnd.github.v3+json"}
```
s-t-e-v-e-n-k, line 169: Is this from removing the preview headers? Since isn't this the default?

github/Commit.py
```diff
@@ -293,6 +294,34 @@ def get_check_runs(
...
+        assert app_id is github.GithubObject.NotSet or isinstance(app_id, int), app_id
+        assert check_name is github.GithubObject.NotSet or isinstance(
+            check_name, str
+        ), check_name
+        parameters = dict()
+        if app_id is not github.GithubObject.NotSet:
+            parameters["app_id"] = app_id
+        if check_name is not github.GithubObject.NotSet:
+            parameters["check_name"] = check_name
+        request_headers = {"Accept": "application/vnd.github.v3+json"}
```
s-t-e-v-e-n-k, line 315: As above.

github/Commit.pyi
```diff
@@ -41,8 +42,13 @@ class Commit(CompletableGithubObject):
...
     @property
     def files(self) -> List[File]: ...
+    def get_check_suites(
+        self,
+        app_id: Union[_NotSetType, int],
+        check_name: Union[_NotSetType, str],
+    ) -> PaginatedList[CheckSuite]: ...
     def get_combined_status(self) -> CommitCombinedStatus: ...
-    def get_comments(self) -> PaginatedList[GitCommit]: ...
+    def get_comments(self) -> PaginatedList[CommitComment]: ...
```
s-t-e-v-e-n-k, line 51: Wasn't that already merged?

s-t-e-v-e-n-k (changes requested):
Make certain that you attribute @wrecker in your commit if you haven't. I have some concerns which I've highlighted inline, but other than that, thanks for addressing my comments from the previous PR.

github/Commit.pyi
dhruvmanila: No, I rebased #1646 on the latest upstream which does not include that change: https://github.com/PyGithub/PyGithub/blob/master/github/Commit.pyi#L45

github/CheckSuite.py
dhruvmanila: Yes, these are the default headers after removing the preview headers. If you want I can remove them, I will also have to remove them from the replay data files.

dhruvmanila:
I pulled down the commits made by @wrecker and rebased it on the latest upstream. So, all the commits still attribute @wrecker as their author. The three commits made by @wrecker in this pull request:
- https://github.com/PyGithub/PyGithub/pull/1764/commits/9bd5c9f6d0024adb7f69f3db48b4e462c0183360
- https://github.com/PyGithub/PyGithub/pull/1764/commits/2df00d1a7d4e8ebe5d3177413d573e50018a5d1b
- https://github.com/PyGithub/PyGithub/pull/1764/commits/897744c13b07d305054ed6995a5e0011afdb41c6

github/CheckSuite.py
s-t-e-v-e-n-k: I see, they are not in the default headers. Objection removed.

github/Commit.pyi
s-t-e-v-e-n-k: Okay, let me break that out into a quick PR.

s-t-e-v-e-n-k: And meged

dhruvmanila: Done ✅

dhruvmanila:
@s-t-e-v-e-n-k @sbesson All endpoints have been added, please take a look at the new changes made.

s-t-e-v-e-n-k (approved):

PR #1774, opened 2020-12-07
Switch all string addition to using f-strings

s-t-e-v-e-n-k:
Now that we no longer support Python 3.5, move to the best feature 3.6+
gives us, f-strings! This will likely be a bit of a moving target if
older code gets merged.

### DIFF
--- a/github/ContentFile.py
+++ b/github/ContentFile.py
@@ -56,7 +56,7 @@ class ContentFile(github.GithubObject.CompletableGithubObject):
         """
         :type: bytes
         """
-        assert self.encoding == "base64", "unsupported encoding: %s" % self.encoding
+        assert self.encoding == "base64", f"unsupported encoding: {self.encoding}"
         return base64.b64decode(bytearray(self.content, "utf-8"))
 
     @property
--- a/github/AuthenticatedUser.py
+++ b/github/AuthenticatedUser.py
@@ -386,7 +386,7 @@ class AuthenticatedUser(github.GithubObject.CompletableGithubObject):
         """
         assert isinstance(following, github.NamedUser.NamedUser), following
         headers, data = self._requester.requestJsonAndCheck(
-            "PUT", "/user/following/" + following._identity
+            "PUT", f"/user/following/{following._identity}"
         )
 
     def add_to_starred(self, starred):
@@ -397,7 +397,7 @@ class AuthenticatedUser(github.GithubObject.CompletableGithubObject):
         """
         assert isinstance(starred, github.Repository.Repository), starred
         headers, data = self._requester.requestJsonAndCheck(
-            "PUT", "/user/starred/" + starred._identity
+            "PUT", f"/user/starred/{starred._identity}"
         )
 
     def add_to_subscriptions(self, subscription):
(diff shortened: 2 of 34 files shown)


sfdye:
Let's use [pre-commit](https://github.com/asottile/pyupgrade) to enforce this

pascal-hofmann (approved):

s-t-e-v-e-n-k:
@sfdye Good poiint, I'll add that in this PR, and then merge it when CI is happy again.

s-t-e-v-e-n-k:
Wait a second, how does pyupgrade enforce this?

sfdye:
if you use the --python36 option, it will enforce f-string i think

henryiii:
It doesn't enforce f-strings everywhere, just where it thinks they are better. [ref](https://github.com/asottile/pyupgrade#f-strings). It would be good to use f-strings as much as possible, because they are faster. :)

sfdye:
> It doesn't enforce f-strings _everywhere_, just where it thinks they are better. [ref](https://github.com/asottile/pyupgrade#f-strings). It would be good to use f-strings as much as possible, because they are faster. :)

That's the more accurate way to put it 😄

s-t-e-v-e-n-k:
@sfdye pyupgrade did some far-reaching changes, but shall we merge this and cut a .1 release?

sfdye:
The changes are non just f-string, more like applying pyupgrade on whole codebase. Could you update the PR title/description and add the commit sha our ignore list 😄

github/ContentFile.py
```diff
@@ -58,7 +56,7 @@ def decoded_content(self):
         """
         :type: bytes
         """
-        assert self.encoding == "base64", "unsupported encoding: %s" % self.encoding
```
sfdye, line 61: need to fix this

github/PaginatedList.py
```diff
@@ -53,12 +51,10 @@ def __getitem__(self, index):
             return self._Slice(self, index)
 
     def __iter__(self):
-        for element in self.__elements:
-            yield element
+        yield from self.__elements
         while self._couldGrow():
             newElements = self._grow()
-            for element in newElements:
-                yield element
```
sfdye, line 61: sweet

s-t-e-v-e-n-k:
So the problem is, if we disentangle this, we get a bunch of changes that pyuppgrade makes calling .format(). I can try again if you wish.

github/ContentFile.py
s-t-e-v-e-n-k: Our assert messages are a mess anyway, but I'll put it back.

sfdye:
> So the problem is, if we disentangle this, we get a bunch of changes that pyuppgrade makes calling .format(). I can try again if you wish.

What is the problem? I don't understand

s-t-e-v-e-n-k:
> What is the problem? I don't understand

Objection removed -- I'll push up a new PR with pyupgrade. Given the formative changes that it will introduce, like yield form, I don't think we should add it to .git-blame-ignore-revs

sfdye:
I feel most changes (remove unicode literal, removing shebang/convert to f-string) are inconsequential

Issue #1788, opened 2020-12-17
Package is missing multiple metadata fields


ssbarnea:
The package publishes on https://pypi.org/project/PyGithub/ lacks multiple metadata fields like link to source code, issue tracker and so on, making very hard to go from there to github.

I will raise a PR to address this issue. Once done pypi will also display number of starts, issues and open PRs automatically, which is an essential part for evaluating how well maintained is a project.

PR #1789, opened 2020-12-17
Add missing links to project metadata

ssbarnea:
Fixes: #1788


ssbarnea:
It was big surprise to me to see that the project did not had metadata inside setup.cfg and was using the old setup.py approach but I assumed a move to the new packaging would be out of scope. Let me know you need help with that and I can propose another change to address that boring bit too.

Issue #1781, opened 2020-12-10
Way to get all the installed GitHub apps in the org


prathameshnyt:
Can you please let me know how I can all the properties for a list of all the installed GitHub apps in the organizattion. 

[Organization.get_installations](https://pygithub.readthedocs.io/en/latest/github_objects/Organization.html#github.Organization.Organization.get_installations ) is not returning all the fields from mentioned in [list-app-installations-for-an-organization](https://developer.github.com/v3/orgs/#list-app-installations-for-an-organization) I can get some information using `get_app` but that requires slug. Let me know if I am missing something here.

dhruvmanila:
It's because the `Installation` object is incomplete. I am working on the fix in #1791

Issue #1776, opened 2020-12-08
Return MainClass.Github.get_installation(id)


atodorov:
Removed in https://github.com/PyGithub/PyGithub/pull/1738/ but we've got an application which uses it and inheriting the base class and adding this method results it lots of boilerplate code due to using `self.__requester` attribute which isn't available in the inherited class.

s-t-e-v-e-n-k:
Are you able to switch the application to using the other get_installation() method?

atodorov:
> Are you able to switch the application to using the other get_installation() method?

IDK really. At the top of my head I think no. The use case is that we have the application ID and using that and the removed method we get an `Installation` object and then iterate on `inst.get_repos()` to find all of the authorized repositories for this installation.

dhruvmanila:
I have taken a look at the installation object and the `GithubIntegration` class and I propose the following changes:

`GithubIntegration` should only be used to generate the JWT or installation access token for the app/installation. No endpoints should be provided in that class. Once the user has generated the appropriate credentials, they should be passed to the main `Github` class and make all the API calls through the instance. The current `GithubIntegration.get_installation()` function [should actually be in the `Repository` class.](https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#get-a-repository-installation-for-the-authenticated-app)

The function `Github.get_installation()` is indeed a [valid endpoint](https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#get-an-installation-for-the-authenticated-app) which should be provided with the `installation_id` parameter and the `Github` object should be initialized with the installation access token. As I mentioned before, the `installation` object is missing a lot of attributes.

@s-t-e-v-e-n-k I can implement this change but I need a go/no-go.

s-t-e-v-e-n-k:
@dhruvmanila Absolutely, please go for it. If you do re-add get_installation() it needs a docstring and tests, of which it had neither before.

dhruvmanila:
@s-t-e-v-e-n-k I want to discuss something: As per [my comment here](https://github.com/PyGithub/PyGithub/issues/1730#issuecomment-723556956), I have been using the class to programmatically get the installation ID for a specific installation in the repository (`GithubIntegration.get_installation(owner, repo)`) and using that I generate the installation access token and then make all the API calls as an installation. Now, with the proposed change of moving the method to `Repository` object, the process will become like so:
1. Instantiate a `GithubIntegration` object and get the JWT.
[...]

PR #1791, opened 2020-12-18
Update Installation object with attributes and related methods

dhruvmanila:
- [x] Add missing attributes in the Installation object
- [x] Add `get_installation_by_id()` method in the Main class (tested locally)
- [x] Move `GithubIntegration.get_installation()` to `Github.get_installation_by_repo()` (tested locally)
- [x] Update docstring for the mentioned methods
- [x] Add test for Installation object attributes
- [x] Add test for `Github.get_installation_by_id()`
- [x] Add test for `Github.get_installation_by_repo()`

Also, the `Installation.get_repos()` method should be moved to the main class as `Github.get_installation_repos()` because that method requires an installation access token and also the `Installation` object is `NonCompletableGithubObject`.

Fixes: #1781 
Fixes: #1776


github/Installation.py
```diff
@@ -77,34 +152,92 @@ def target_type(self):
...
         return self._target_type.value
 
+    @property
+    def updated_at(self):
+        """
+        :type: datetime.datetime
+        """
+        return self._updated_at.value
+
     def get_repos(self):
```
dhruvmanila, line 162: I also don't think this method belongs in the `Installation` object as that is a `NonCompletable` and also the API call does not depend on any of the Installation attributes. It should be moved to the `github.Github.get_installation_repos`.

dhruvmanila:
@s-t-e-v-e-n-k Hey, can you take a look at this? Thanks.

github/MainClass.py
```diff
@@ -881,23 +917,3 @@ def get_access_token(self, installation_id, user_id=None):
         raise GithubException.GithubException(
             status=response.status_code, data=response.text
         )
-
-    def get_installation(self, owner, repo):
```
dhruvmanila, line 885: Hey, should we add a deprecation warning for this method instead of removing it completely?

github/MainClass.py
```diff
@@ -806,6 +806,42 @@ def get_app(self, slug=github.GithubObject.NotSet):
             headers, data = self.__requester.requestJsonAndCheck("GET", "/apps/" + slug)
             return GithubApp.GithubApp(self.__requester, headers, data, completed=True)
 
+    def get_installation_by_id(self, installation_id):
+        """
+        You must use a JWT to access this endpoint which can be generated using the
+        method `create_jwt()` from :class:`github.GithubIntegration`.
```
s-t-e-v-e-n-k, line 812: There's a :meth: link, I *think* there's an existing example, means the method and the class are both links.

s-t-e-v-e-n-k (changes requested):
My biggest concern with this is it is breaks all users of Installation since it pretty much rewrites it. Can you see any sort of migration plan here?

dhruvmanila:
> My biggest concern with this is it is breaks all users of Installation since it pretty much rewrites it. Can you see any sort of migration plan here?

I will restore the method in `GithubIntegration` class which will emit a deprecation warning and will use the updated method in the main class. I think that's the only method which will be deprecated and about bringing back the `get_installation(id)` from #1776 the name has been changed from `get_installation` to `get_installation_by_id` to avoid conflicts with the updated method `get_installation_by_repo`.

dhruvmanila:
Any updates here?

dhruvmanila:
Is this still relevant or should I close this?

megamorf:
I'd like to see this feature implemented as well. I was just in the process of automating the monthly permissions check on our org and was surprised to see that I cannot really query installed apps properly.

github/MainClass.py
```diff
@@ -789,6 +790,42 @@ def get_app(self, slug=github.GithubObject.NotSet):
             headers, data = self.__requester.requestJsonAndCheck("GET", f"/apps/{slug}")
             return GithubApp.GithubApp(self.__requester, headers, data, completed=True)
 
+    def get_installation_by_id(self, installation_id):
```
EnricoMi, line 793: This is now implemented by `GithubInstallation.get_app_installation(installation_id)`.

github/MainClass.py
```diff
@@ -789,6 +790,42 @@ def get_app(self, slug=github.GithubObject.NotSet):
...
+        """
+        assert isinstance(installation_id, int), installation_id
+        headers, data = self.__requester.requestJsonAndCheck(
+            "GET", f"/app/installations/{installation_id}"
+        )
+        return Installation.Installation(
+            self.__requester, headers, data, completed=True
+        )
+
+    def get_installation_by_repo(self, owner, repo):
```
EnricoMi, line 810: This is now implemented by `GithubIntegration.get_repo_installation(owner, repo)`.

github/Installation.py
EnricoMi: Class `github.Github` is already cluttered with lots of class specific methods.

Why is `Github.get_installation_repos` better than `installation.get_repos`? With the latter it is clear that this requires an installation auth, while `Github` can be used with lots of different authentications, but that method works only with an installation auth.

dhruvmanila:
Honestly, I've actually lost context regarding this. I might need to look at it from scratch to pick it up but I don't want to spend time if it has become irrelevant.

If anyone else wants to continue the work done here, please feel free to do so. I'm happy to help :)

Issue #1796, opened 2020-12-22
Syntax error due to pyjwt-2.0.0


PaulLesur:
Hello ! 
We are facing syntax errors using PyGithub since pyjwt version 2.0.0.

```
Traceback (most recent call last):
  File "ci-tools/get_corresponding_tag.py", line 4, in <module>
    from github import Github
  File "/opt/circleci/.pyenv/versions/3.5.2/lib/python3.5/site-packages/github/__init__.py", line 58, in <module>
    from github.MainClass import Github, GithubIntegration
  File "/opt/circleci/.pyenv/versions/3.5.2/lib/python3.5/site-packages/github/MainClass.py", line 57, in <module>
    import jwt
  File "/opt/circleci/.pyenv/versions/3.5.2/lib/python3.5/site-packages/jwt/__init__.py", line 1, in <module>
    from .api_jws import (
  File "/opt/circleci/.pyenv/versions/3.5.2/lib/python3.5/site-packages/jwt/api_jws.py", line 134
    **kwargs,
            ^
SyntaxError: invalid syntax
```

A temporary solution is to install this dependency with a version inferior to 2.0.0

jambon69:
:+1:

gturbat:
:+1:

xupefei:
We are seeing this issue too in Python 3.5. For some reason, `pip3` choose to install `pyjwt 2.0.0` even when https://pypi.org/project/PyJWT/ says 2.0.0 "Requires: Python >=3.6".

PR #1797, opened 2020-12-22
Fix pyjwt version

Hackatosh:
Pyjwt version is not fixed, which can cause syntax error for user since the release of breaking changes from pyjwt (https://pypi.org/project/PyJWT/#history)

Related Issue : https://github.com/PyGithub/PyGithub/issues/1796

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 requests>=2.14.0
-pyjwt
+pyjwt<2.0
 sphinx<3
 sphinx-rtd-theme<0.6
 Deprecated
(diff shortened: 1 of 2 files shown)


LouisJULIEN (approved):

gturbat (approved):

s-t-e-v-e-n-k:
Also need to restrict it in setup.py as well

Hackatosh:
> Also need to restrict it in setup.py as well

I have made the change, thanks

austinsasko (approved):

PR #1807, opened 2020-12-29
Allow adding attributes at the end of the list

### LINKED ISSUE #1806: Adding new attribute fails in case new name is the last in the list
### Problem Statement

```bash
$ python scripts/add_attribute.py Permissions triage bool
  File "<...>\PyGithub\scripts\add_attribute.py", line 124, in <module>
    line = lines[i].rstrip()
IndexError: list index out of range
```
--> Adding a new attribute at the end of the existing list of attributes in class `Permissions` fails.
--> In this case the last attribute name was "push", so "triage" comes last.
https://github.com/PyGithub/PyGithub/blob/34d097ce473601624722b90fc5d0396011dd3acb/github/Permissions.py#L63-L72

### Solution Approach

In case the new attribute name will result in adding it at the end of the list of attributes, then the processing within the script at https://github.com/PyGithub/PyGithub/blob/master/scripts/add_attribute.py#L89 was already processing the next source code line which already contains the `_initAttributes` function.
Subsequently at https://github.com/PyGithub/PyGithub/blob/master/scripts/add_attribute.py#L122 `inInit` is set to `False`, but only checked again after reading already the next line. This means the following code block will never again notice the place of the `_initAttributes` and fails at the end of the file due to endless loop.

Problem can be fixed by conditionally remembering we already reached the `_initAttributes` function, so replace:
https://github.com/PyGithub/PyGithub/blob/34d097ce473601624722b90fc5d0396011dd3acb/scripts/add_attribute.py#L122

with

```python
inInit = True if line == "    def _initAttributes(self):" else False
```

karsten-wagner:
This intents to fix #1806

### DIFF
--- a/scripts/add_attribute.py
+++ b/scripts/add_attribute.py
@@ -119,7 +119,7 @@ while not added:
 
 added = False
 
-inInit = False
+inInit = True if line == "    def _initAttributes(self):" else False
 while not added:
     line = lines[i].rstrip()
     i += 1
--- a/scripts/add_attribute.py
+++ b/scripts/add_attribute.py
@@ -119,7 +119,7 @@ while not added:
 
 added = False
 
-inInit = True if line == "    def _initAttributes(self):" else False
+inInit = line.endswith("def _initAttributes(self):")
 while not added:
     line = lines[i].rstrip()
     i += 1


scripts/add_attribute.py
```diff
@@ -119,7 +119,7 @@
 
 added = False
 
-inInit = False
+inInit = True if line == "    def _initAttributes(self):" else False
```
s-t-e-v-e-n-k, line 122: This is awkward. How about?

`inInit = line.endswith("def _initAttributes(self):")`

karsten-wagner: @s-t-e-v-e-n-k , sure - updated.
What a pitty, there are so view chances to use ternary operators...

s-t-e-v-e-n-k: Haha! I am not against them, but only when absolutely required :-)

Issue #1812, opened 2021-01-01
Code coverage report on codecov.io incomplete


karsten-wagner:
PyGithub has 107 files in folder `/github` at the time of writing, but only 51 files are covered in codecov.io report.
Compare: https://codecov.io/gh/PyGithub/PyGithub/tree/master/github

This seems to be an issue on codecov.io, since the coverage report contains all 107 classes after the test.
Compare: https://codecov.io/api/gh/PyGithub/PyGithub/download/build?path=v4/raw/2020-12-24/F473F251CFE431584CB1C976534196B2/34d097ce473601624722b90fc5d0396011dd3acb/bc973103-0c9d-4aab-9009-458491840c75.txt

Can you please either open a support request with codecov for your project, or apply a manual workaround?
I can provide a PR which was reflecting all files based on manually applying the path fixes needed to match the codecov.io conventions...

PR #1813, opened 2021-01-01
Manually fixing paths for codecov.io to cover all project files

karsten-wagner:
This fix will change the codecov.io configuration to replace automatic path handling with manual fixes. This is needed to properly reflect all project files in the codecov.io report, since with automatic path fixes it is covering only roughly 50% of the code (compare number of files in the project and those reflected on codecov.io).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,8 @@
 comment:
   after_n_builds: 4
+
+codecov:
+  disable_default_path_fixes: true # manual fix properly covers all files
+
+fixes:
+  - "::github/" # manual fix properly covers all files
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,7 @@ comment:
   after_n_builds: 4
 
 codecov:
-  disable_default_path_fixes: true # manual fix properly covers all files
+  disable_default_path_fixes: true # Automatic detection does not discover all files
 
 fixes:
   - "::github/"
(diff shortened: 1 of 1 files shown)


karsten-wagner:
This PR aims to fix #1812.

Feel free to apply or to address this to codecov.io to have them fix the automatic path fixing...

karsten-wagner:
Note all files are reported unchanged in the code coverage comment, but the number of covered files increases from 51 to 107.

codecov.yml
```diff
@@ -1,2 +1,8 @@
 comment:
   after_n_builds: 4
+
+codecov:
+  disable_default_path_fixes: true # manual fix properly covers all files
```
s-t-e-v-e-n-k, line 5:
```suggestion
  disable_default_path_fixes: true # Automatic detection does not discover all files
```

The comment isn't clear to me

codecov.yml
```diff
@@ -1,2 +1,8 @@
...
+
+fixes:
+  - "::github/" # manual fix properly covers all files
```
s-t-e-v-e-n-k, line 8:
```suggestion
  - "::github/"
```

I don't think we need a comment here, the one above makes it clear (to me, at least)

karsten-wagner:
@s-t-e-v-e-n-k , sure - reworked the comment(s) as per your suggestion...

PR #1831, opened 2021-01-29
Make datetime objects timezone-aware

jwodder:
Timestamps returned by the GitHub API include timezone information, yet PyGithub throws this information away.  This can lead to wrong results in client code, as naïve timestamps are generally assumed to use the system time zone, yet the majority of timestamps returned by the API are in UTC.  This PR fixes the library to retain timezone information when parsing timestamps and always produce timezone-aware `datetime` objects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 on:
   push:
+    branches: master
   pull_request:
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: CI
 on:
   push:
-    branches: master
   pull_request:
 jobs:
   test:
--- a/tests/GithubObject.py
+++ b/tests/GithubObject.py
@@ -1,0 +1,94 @@
+############################ Copyrights and license ############################
+#                                                                              #
+# Copyright 2023 Enrico Minack <github@enrico.minack.dev>                      #
+#                                                                              #
+# This file is part of PyGithub.                                               #
+# http://pygithub.readthedocs.io/                                              #
+#                                                                              #
+# PyGithub is free software: you can redistribute it and/or modify it under    #
+# the terms of the GNU Lesser General Public License as published by the Free  #
+# Software Foundation, either version 3 of the License, or (at your option)    #
+# any later version.                                                           #
+#                                                                              #
+# PyGithub is distributed in the hope that it will be useful, but WITHOUT ANY  #
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS    #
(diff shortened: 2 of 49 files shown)


sylvainmouquet:
@sfdye @s-t-e-v-e-n-k

could you give us information about the merge of this PR in master please ?

jwodder:
@sfdye @s-t-e-v-e-n-k What do I need to do to get this PR merged?

jwodder:
@sfdye @s-t-e-v-e-n-k Is there any interest in ever merging this?

EnricoMi:
@jwodder would be good to add some test cases for this: https://github.com/jwodder/PyGithub/pull/1

EnricoMi:
This might be a breaking change for some users and require a major version bump? Or should this behaviour be configurable (enabled by default) so users can disable it when it hits them?

See https://github.com/PyGithub/PyGithub/issues/1905#issuecomment-881933460.

EnricoMi:
@jwodder what about using `dateutil.parser`? https://github.com/jwodder/PyGithub/pull/2

.github/workflows/ci.yml
```diff
@@ -1,7 +1,6 @@
 name: CI
 on:
   push:
-    branches: master
```
EnricoMi, line 4: you should revert this line

tests/GithubObject.py
```diff
@@ -0,0 +1,120 @@
...
+                "2021-01-23T12:34:56-06:30",
+                datetime(
+                    2021,
+                    1,
+                    23,
+                    12,
+                    34,
+                    56,
+                    tzinfo=timezone(timedelta(hours=-6, minutes=-30)),
+                ),
```
EnricoMi, line 67: Maybe black is fine with this line:
```suggestion
                datetime(
                    2021, 1, 23, 12, 34, 56, tzinfo=timezone(timedelta(minutes=-390))
                ),
```

EnricoMi:
@s-t-e-v-e-n-k @sfdye are you happy with releasing changes that potentially break user code in a minor version bump (like v1.59.0), or should this go into v2.0.0?

Returning timezoned datetime will break user code that compares those with naive datetimes: https://github.com/PyGithub/PyGithub/issues/1905#issuecomment-881933460

JLLeitschuh:
I generally prefer "more correct" but that breaks things potentially, over "broken/incorrect" but stable. Principle of least surprise for new users.

EnricoMi:
I will use this to revert https://github.com/PyGithub/PyGithub/pull/2561/commits/1b228ae67a113997154b3aefb38324a066675e0e#diff-984677619e2408123d5a48bb0e08edc8ca94aef5d17c8b8a5ee35dd8e6430b76R310-R311.

EnricoMi:
@jwodder don't bother with the conflicts, I'll resolve them soon.

EnricoMi:
Fixed by #2565

PR #1850, opened 2021-02-12
Updating links to Github documentation

fhightower:
There are a couple of links to http://developer.github.com/v3/repos/keys/ and I believe these should be going to http://developer.github.com/v3/repos/#deploy-keys. I've updated the docstrings in this file appropriately.


jobveldhuis (approved):
Well spotted. Probably some left over bits of old Github documentation. Changes look good to me.

PR #1857, opened 2021-02-19
Add default arguments to Workflow method type annotations

jwodder:
The type annotations for `Workflow.create_dispatch()` and `Workflow.get_runs()` do not include the fact that some of the arguments have default values.  As a result, user code that calls, say, `workflow.get_runs()` without specifying any arguments will not validate with mypy.  This PR fixes that.


jobveldhuis (approved):
LGTM

PR #1887, opened 2021-03-23
Export headers in GithubException

### LINKED ISSUE #1814: Handling hitting abuse detection limits and using Retry-After
Hey

Thanks a lot for the library. It really helped us with our CI setup.

However recently we have started hitting the abuse detection limits imposed by GH https://docs.github.com/en/free-pro-team@latest/rest/guides/best-practices-for-integrators#dealing-with-abuse-rate-limits

When this happens the library seems to just throw an exception. How can I get the Retry-After header? It is part of the exception?

s-t-e-v-e-n-k:
Since the headers that led to an exception are also useful, firstly pass
them into the constructor, and then export them in a property. Test one
specific use case to make sure of coverage.

Fixes #1814


sbesson:
Hi @s-t-e-v-e-n-k, just came across this as PyGithub 1.55 got released yesterday and our calls to `GitHubException` constructors started failing. One suggestion: at the moment, this is listed under `Bug Fixes & Improvements` in https://github.com/PyGithub/PyGithub/blob/v1.55/doc/changes.rst but I suspect it might be worth mentioning in `Breaking Changes` as well since client code will need to be adjusted to pass an extra field.

s-t-e-v-e-n-k:
Good point, I'll edit the release notes

PR #1890, opened 2021-03-23
Adjust to Github API changes regarding emails

### LINKED ISSUE #1852: AuthenticatedUser.get_emails() returned data changed to list of dicts
According to the [provided type information](https://github.com/PyGithub/PyGithub/blob/master/github/AuthenticatedUser.pyi#L111) for `AuthenticatedUser`, the `get_emails` function returns a list of string.  However, in my experience with version 1.54.1 of this module, it appears to return a list of dictionaries that looks like the following:

```
[
  {
    "email": "xxx@yyy.com",
    "primary": false,
    "verified": true,
    "visibility": null
  },
  ...
]
```

MeggyCal:
Closes #1852


github/AuthenticatedUser.py
```diff
@@ -373,7 +374,9 @@ def add_to_emails(self, *emails):
         :rtype: None
         """
         assert all(isinstance(element, str) for element in emails), emails
-        post_parameters = emails
+        post_parameters = {
+            "emails": emails
+        }
```
s-t-e-v-e-n-k, line 379: Let me guess, GitHub also requires the emails key now. Bah. (I'm annoyed at GitHub changing the API without notice, rather than this change.)

github/AuthenticatedUser.py
```diff
@@ -763,10 +766,14 @@ def get_authorizations(self):
...
         :calls: `GET /user/emails <http://developer.github.com/v3/users/emails>`_
-        :rtype: list of string
+        :rtype: list of namedtuples with members email, primary, verified and visibility
         """
         headers, data = self._requester.requestJsonAndCheck("GET", "/user/emails")
-        return data
+        emaildata = []
+        itemdata = namedtuple("EmailData", "email primary verified visibility")
+        for item in data:
+            emaildata.append(itemdata._make(item.values()))
```
s-t-e-v-e-n-k, line 775: `return [itemdata._make(item.values()) for item in data]` saves having to populate a second list, we can just return a list comprehension.

github/AuthenticatedUser.py
```diff
@@ -763,10 +766,14 @@ def get_authorizations(self):
     def get_emails(self):
         """
         :calls: `GET /user/emails <http://developer.github.com/v3/users/emails>`_
-        :rtype: list of string
+        :rtype: list of namedtuples with members email, primary, verified and visibility
         """
         headers, data = self._requester.requestJsonAndCheck("GET", "/user/emails")
-        return data
+        emaildata = []
+        itemdata = namedtuple("EmailData", "email primary verified visibility")
```
s-t-e-v-e-n-k, line 773: data[0].keys(), since dictionary ordering is Python version independent.

tests/AuthenticatedUser.py
```diff
@@ -101,22 +101,28 @@ def testEditWithAllArguments(self):
     def testEmails(self):
         self.assertEqual(
             self.user.get_emails(),
-            ["vincent@vincent-jacques.net", "github.com@vincent-jacques.net"],
+            [
+                EmailData(email="vincent@vincent-jacques.net", primary=True, verified=True, visibility="private"),
```
s-t-e-v-e-n-k, line 105: Surely this is NameError? You need to call the object returned. Have a look at other examples that call a method using assertListKeyEquals, and assert on .email only -- maybe checking the first item has private visibility and so on.

Comparing complex data structures is not easy with base unittest.

s-t-e-v-e-n-k (changes requested):
This is looking like a great start, I have some concerns inline.

github/AuthenticatedUser.py
MeggyCal: I think so from https://docs.github.com/en/rest/reference/users#add-an-email-address-for-the-authenticated-user

tests/AuthenticatedUser.py
MeggyCal: I am checking only for emails. Maybe checking for one "primary" would make sense, but nothing more.

tests/ReplayData/AuthenticatedUser.testEmails.txt
```diff
@@ -7,18 +7,18 @@ None
 None
 200
 [('status', '200 OK'), ('x-ratelimit-remaining', '4934'), ('content-length', '64'), ('server', 'nginx/1.0.13'), ('connection', 'keep-alive'), ('x-ratelimit-limit', '5000'), ('etag', '"ea6dacf29569317ccf460b4bb07075e5"'), ('date', 'Sun, 20 May 2012 12:41:39 GMT'), ('content-type', 'application/json; charset=utf-8')]
-["vincent@vincent-jacques.net","github.com@vincent-jacques.net"]
+[{"email": "vincent@vincent-jacques.net", "primary": True, "verified": True, "visibility": "private"},{"email": "github.com@vincent-jacques.net", "primary": False, "verified": True, "visibility": None}]
```
s-t-e-v-e-n-k, line 10: I understand the failure now -- this is the Python representation of the data, but the test machinery expects JSON, so that data[0] error is masking a JSON decode error.

```
[{"email": "vincent@vincent-jacques.net", "primary": true, "verified": true, "visibility": "private"}, {"email": "github.com@vincent-jacques.net", "primary": false, "verified": true, "visibility": null}]
```

is the correct JSON representation that should go here -- and in the other replay data.

MeggyCal: OK, thanks, I will fix it.

tests/AuthenticatedUser.py
```diff
@@ -100,12 +100,12 @@ def testEditWithAllArguments(self):
 
     def testEmails(self):
         self.assertEqual(
-            self.user.get_emails(),
+            [item.email for item in self.user.get_emails()],
```
s-t-e-v-e-n-k, line 103: You probably need to pull this out of the list comprehension so you check at least primary and hopefully the other two for the first element.

s-t-e-v-e-n-k (approved):

Issue #1832, opened 2021-01-30
allow PyJWT 2+


valentijnscholten:
other libraries are moving to PyJWT2+ as requirement, is it possible to update pygithub as well? currently we can't use for example pygithub together with django-social-core

s-t-e-v-e-n-k:
It's on my list, just a lack of free time to tackle it. :-(

fabaff:
I'm looking forward to see. The NixOS package will break if we update `PyJWT`.

fabaff:
Any progress?

It's blocking the PyJWT package update on NixOS. I guess that it affecting other distributions as well.

samamorgan:
What breaking changes does PyJWT>2 introduce?

s-t-e-v-e-n-k:
https://github.com/PyGithub/PyGithub/issues/1796 is what caused me to pin it, I just haven't time to dig into it

samamorgan:
@s-t-e-v-e-n-k the issues mentioned there are related to Python 3.5 specifically. I'm assuming these syntax errors occur due to syntax changes with Python 3.6. Which makes sense considering the drop of Python 3.5 support.

Isn't the current version of PyGithub dropping Python 3.5 support as well? That was mentioned in the latest release.

s-t-e-v-e-n-k:
The current version has dropped support for 3.5, yes. I'll propose a PR to drop the pin soon.

PR #1891, opened 2021-03-23
Update PyJWT Version to 2.0+

samamorgan:
Most other libraries that use PyJWT and have dropped Python 3.5 support now require version 2.0+, which creates an impossible to resolve dependency issue.

- Updated setup and requirements
- Fixed an issue with a test


samamorgan:
Fixes #1832

PR #1894, opened 2021-03-25
key_id could be int on Github Enterprise

### LINKED ISSUE #1878: Bug in repr(PublicKey)
hi there.

there is a bug in "repo.get_public_key".

i found this bug by trying to use "repo.create_secret", which failed because of this issue

```
repo.get_public_key()
  File "<input>", line 1, in <module>
  File "/Users/sschultchen/PycharmProjects/IAC_Main/venv/lib/python3.9/site-packages/github/PublicKey.py", line 55, in __repr__
    return self.get__repr__({"key_id": self._key_id.value, "key": self._key.value})
  File "/Users/sschultchen/PycharmProjects/IAC_Main/venv/lib/python3.9/site-packages/github/GithubObject.py", line 62, in value
    raise GithubException.BadAttributeException(
github.GithubException.BadAttributeException: (1, <class 'str'>, None)
```

i guess the reason for this might be that the pygithub implementation assumes that "key_id" is a string, but it actually is a integer.

at least, the github api gives me an integer, and not a string for this attribute.

MeggyCal:
Closes #1878


tests/PublicKey.py
```diff
@@ -33,9 +33,9 @@
 class PublicKey(Framework.TestCase):
     def setUp(self):
         super().setUp()
-        self.public_key = self.g.get_user().get_repo("PyGithub").get_public_key()
```
s-t-e-v-e-n-k, line 36: If setUp() now does nothing except call the superclass, you can remove it.

s-t-e-v-e-n-k (changes requested):
Looking excellent, just one minor niggle.

github/PublicKey.py
```diff
@@ -78,7 +78,10 @@ def _useAttributes(self, attributes):
         if "key" in attributes:  # pragma no branch
             self._key = self._makeStringAttribute(attributes["key"])
         if "key_id" in attributes:  # pragma no branch
-            self._key_id = self._makeStringAttribute(attributes["key_id"])
+            if type(attribute["key_id"]) == str:
```
s-t-e-v-e-n-k, line 81: "attributes" :-)

s-t-e-v-e-n-k (approved):

Issue #1898, opened 2021-03-28
Access the `assets` field in releases response to avoid extra requests


green-green-avk:
1. A response to `GET /repos/{owner}/{repo}/releases` already contains an assets list for each release.
2. **GitHub** has [Rate limiting](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting).

So, it's better to have an ability to access this list.

green-green-avk:
Pull request: <https://github.com/PyGithub/PyGithub/pull/1899>

Felixoid:
The issue is not solved yet and the PR wasn't merged

PR #1899, opened 2021-03-28
A property to access the `assets` field of release (#1898)

green-green-avk:
... in order to avoid extra requests.

<https://github.com/PyGithub/PyGithub/issues/1898>


github/GitRelease.py
```diff
@@ -170,6 +170,16 @@ def zipball_url(self):
         self._completeIfNotSet(self._zipball_url)
         return self._zipball_url.value
 
+    @property
+    def assets(self):
+        """
+        Already returned assets info (no additional requests will be done).
```
s-t-e-v-e-n-k, line 176: Just the :type: here, please

s-t-e-v-e-n-k (changes requested):
This is looking like a good start, I have one niggle inline, and this requires a test -- if the existing replay data used in tests/GitRelease.py already has assets, then you can assert that and you're golden.

github/GitRelease.py
green-green-avk: Where to put a note on the difference between `get_assets()` and `assets` though?

PR #1907, opened 2021-04-04
Created a module with a set of functions allowing offline conversion of DB IDs to GraphQL IDs and back

KOLANICH:


KOLANICH:
Because of the changes in GH infrastructure the modules become irrelevant.

PR #1909, opened 2021-04-06
Fix __log mutating requestHeaders

nmiculinic:
The requester log mutates the passed request headers if the logging level includes logging. This leads to the following bug:

* log level is debug
* we perform github operation on repo
* we get HTTP 301, redirection. The request is logged..and the log method mutates the requestHeaders
* requester tries again, now to the redirected URL...but with mutated requestHeaders
* The request fails, since the auth token has been sanitized

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -51,6 +51,7 @@
 ################################################################################
 
 import base64
+import copy
 import json
 import logging
 import mimetypes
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
That is a very good catch, thanks. Do you think it's possible to write a unit test for it?

nmiculinic:
Currently I'm working on other tasks and am using forked pyGithub (with this fix)...if I find time I'll hopefully write unit-tests for this.

s-t-e-v-e-n-k:
Thanks for your work, but I've written a PR that does this, and also adds a unittest.

PR #1926, opened 2021-04-23
Do not import from unpackaged paths in typing

nakato:
The tests path is not part of the shipped project, thus this type cannot
be checked by users of this project.

The correct type in place of the type imported from tests appears to be
Optional[
    Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]
]

### DIFF
--- a/github/Requester.pyi
+++ b/github/Requester.pyi
@@ -5,7 +5,6 @@ from typing import Any, Callable, Dict, Iterator, Optional, Tuple, Union
 from requests.models import Response
 
 from github.GithubObject import GithubObject
-from tests.Framework import ReplayingHttpsConnection
 
 # from urllib3.util.retry import Retry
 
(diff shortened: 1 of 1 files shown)


github/Requester.pyi
```diff
@@ -57,7 +56,9 @@ class Requester:
     def __authenticate(
         self, url: str, responseHeader: Dict[str, Any], parameters: Dict[str, Any],
     ) -> None: ...
-    def __customConnection(self, url: str,) -> Optional[ReplayingHttpsConnection]: ...
+    def __customConnection(
+        self, url: str,
+    ) -> Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]]: ...
```
s-t-e-v-e-n-k, line 61: Both of those classes are a subclass of ReplayingConnection, does it also work if you specify that instead of the union of HTTP{,S}?

PR #1952, opened 2021-05-15
Use generic for expected_type in BadAttributeException

nakato:
The expected_type return value is bound to expected_type argument on
init.  This will type the return of expected_type to the type passed to
the init function.

I'm choosing to not bind TypeVar to anything explicit, as in theory
the exception should not be limited to any specific expected type, and
the union quickly balloons and was already significant.

This fixes the implicit Any generic dict in Type[dict], that turns into
Type[dict[Any, Any]] and is disallowed when setting
--disallow-any-generics (this still allows explicitly defined Any
generics).  The generic also means consumption of the returned value
doesn't require handling all listed types in the Union as the type will
be specifically the type of expected_type.

Included is a couple lines of black formatting as they occurred on
write and are insignificant in number.

### DIFF
--- a/github/GithubException.pyi
+++ b/github/GithubException.pyi
@@ -1,39 +1,30 @@
-from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, Generic, List, Optional, Tuple, TypeVar, Union
 
 class GithubException(Exception):
-    def __init__(self, status: Union[int, str], data: Any, headers: Optional[Dict[str, str]]) -> None: ...
+    def __init__(
+        self, status: Union[int, str], data: Any, headers: Optional[Dict[str, str]]
+    ) -> None: ...
     def __str__(self) -> str: ...
     @property
     def data(self) -> Dict[str, Union[str, List[str], List[Dict[str, str]]]]: ...
     @property
     def status(self) -> int: ...
-
     @property
     def headers(self) -> Union[None, Dict[str, str]]: ...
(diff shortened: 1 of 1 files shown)


github/GithubException.pyi
```diff
@@ -1,39 +1,30 @@
...
     @property
     def status(self) -> int: ...
-
     @property
     def headers(self) -> Union[None, Dict[str, str]]: ...
 
-class BadAttributeException(GithubException):
+T = TypeVar("T")
+
+class BadAttributeException(GithubException, Generic[T]):
```
s-t-e-v-e-n-k, line 17: Why does the class inherent from Generic[T]? I thought this class didn't wrap the attribute.

nakato: The Generic is making the BadAttributeException be a generic class instead of a concrete class.  We cannot use the TypeVar inside the class without binding TypeVar T to the class.

So instead of having `BadAttributeException` we now have `BadAttributeException[T]`.

When initializing with `BadAttributeException(_, T, _)`, this now means `expected_value` will be typed as `expected_type(self) -> T`.

So if it is initialized with `str` for expected type, then `expected_value`'s type is `expected_type(self) -> str`.


This documentation should describe it better than I can.
https://mypy.readthedocs.io/en/stable/generics.html#defining-sub-classes-of-generic-classes

nakato:
Closing this, I'm going to take a second pass at this later.

PR #1954, opened 2021-05-17
Add type to OrderedDict

nakato:
OrderedDict needs to be typed just like Dict needs to be typed.  The
OrderedDict "input" variable key and value are used to append to a
string without any processing, so it seems unlikely that something other
than a string would be valid.

### DIFF
--- a/github/Requester.pyi
+++ b/github/Requester.pyi
@@ -158,7 +158,7 @@ class Requester:
         url: str,
         parameters: Optional[Dict[str, Any]] = ...,
         headers: Optional[Dict[str, Any]] = ...,
-        input: Optional[OrderedDict] = ...,
+        input: Optional[OrderedDict[str, str]] = ...,
         cnx: Optional[
             Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]
         ] = ...,
@@ -169,7 +169,7 @@ class Requester:
         url: str,
         parameters: Optional[Dict[str, Any]] = ...,
         headers: Optional[Dict[str, Any]] = ...,
-        input: Optional[OrderedDict] = ...,
+        input: Optional[OrderedDict[str, str]] = ...,
     ) -> Tuple[Dict[str, Any], Optional[Dict[str, Any]]]: ...
     @classmethod
     def resetConnectionClasses(cls) -> None: ...


s-t-e-v-e-n-k:
I'm not certain here, but I *think* input's value can also be bytes. key should always be string. I hope.

nakato:
> I'm not certain here, but I _think_ input's value can also be bytes. key should always be string. I hope.

Wouldn't that be a TypeError?
```python3
>>> a = "str"
>>> a += b"bytes"
TypeError: can only concatenate str (not "bytes") to str
>>> 
```

s-t-e-v-e-n-k:
Not that input's values can be string and bytes  -- but I was right, you need to check more closely, a quick grep showed me at least one call site where the value is also a boolean.

nakato:
> a quick grep showed me at least one call site where the value is also a boolean.

How does that work?  Python throws a TypeError Exception for `a += True` too.

*__request_encode* does not do any processing to input, and that calls the encode function defined inside [*requestMultipart*](https://github.com/PyGithub/PyGithub/blob/d1644e3350a8f8e87a84777c3d551718e383b9e9/github/Requester.py#L464), it doesn't do any pre-processing on input either.

```
            encoded_input = ""
            for name, value in input.items():
                encoded_input += f"--{boundary}{eol}"
                encoded_input += f'Content-Disposition: form-data; name="{name}"{eol}'
                encoded_input += eol
                encoded_input += value + eol
            encoded_input += f"--{boundary}--{eol}"
            return f"multipart/form-data; boundary={boundary}", encoded_input
```

nakato:
> but I was right, you need to check more closely, a quick grep showed me at least one call site where the value is also a boolean.

Where?  I don't see that use.

```
% grep -Ri requestMultipart github tests                                                     
github/Repository.py:        headers, output = self._requester.requestMultipartAndCheck(
github/Requester.py:    def requestMultipartAndCheck(
github/Requester.py:            *self.requestMultipart(
github/Requester.py:    def requestMultipart(
github/Requester.pyi:    def requestMultipart(
github/Requester.pyi:    def requestMultipartAndCheck(
```

The only use of it I see, including grepping the tests, is in [Repository.py](https://github.com/PyGithub/PyGithub/blob/d1644e3350a8f8e87a84777c3d551718e383b9e9/github/Repository.py#L3536), and that is gated with (runtime) asserts that ensure everything used to create *post_parameters* are str.

s-t-e-v-e-n-k:
These request methods all look the damned same, I swear ....

PR #1972, opened 2021-06-15
Add two_factor_authentication in AuthenticatedUser.

### LINKED ISSUE #1971: Missing two_factor_authentication attribute in AuthenticatedUser
`AuthenticatedUser` class should have a `two_factor_authentication` attribute.

Reference: https://docs.github.com/en/rest/reference/users#get-the-authenticated-user

Hanaasagi:
closes #1971


tests/AuthenticatedUser.py
```diff
@@ -76,6 +76,7 @@ def testAttributes(self):
         self.assertEqual(self.user.url, "https://api.github.com/users/jacquev6")
         self.assertEqual(self.user.node_id, "MDQ6VXNlcjMyNzE0Ng==")
         self.assertEqual(repr(self.user), 'AuthenticatedUser(login="jacquev6")')
+        self.assertEqual(self.user.two_factor_authentication, True)
```
s-t-e-v-e-n-k, line 79: self.assertTrue() please

s-t-e-v-e-n-k:
Further more, please run `tox -elint`, that's the cause of the 3.6 CI failure.

Hanaasagi:
Here is the error message when I run the `tox -elint` in my local. The `types-requests` and `types-jwt` is not installed. I'm not sure why the master branch could pass this test.

图片

Hanaasagi:
I try a new Python 3.6.13 virtualenv, install the newest tox(3.23.1) in my local. The master branch could not pass the lint. Missing the `types-jwt` and `types-request` too.

Hanaasagi:
I find the difference!. The pre success CI uses the mypy==0.812. See the Action log: https://github.com/PyGithub/PyGithub/runs/2724801619#step:5:10. And now it uses the mypy==0.902. It has a breaking change: 

> Mypy now only ships with type stubs for stdlib modules (plus typing_extensions and mypy_extensions). If you use third-party libraries that don’t include stubs or inline type annotations, you can explicitly install stub packages, usually called types-<name>.

s-t-e-v-e-n-k:
You can revert that tox.ini change, I have a separate PR up that I will merge that adds that.

Hanaasagi:
Done.

Issue #1855, opened 2021-02-17
Include GitHub App usage in doc/examples


swinton:
It would be great to have an example showing how to use PyGithub with [GitHub Apps](https://docs.github.com/en/developers/apps/creating-a-github-app).

GitHub Apps are a "first class vehicle" for integrating with GitHub, and [offer several benefits over OAuth / Personal Access Tokens](https://docs.github.com/en/developers/apps/migrating-oauth-apps-to-github-apps#reasons-for-switching-to-github-apps).

I have [some demo code](https://github.com/swinton/pygithub-playground/blob/main/demo.py) that works, and I'd be happy to submit a PR for this example, if that would be a welcome addition.

goldstar611:
I second this. Having some good reference scripts, even if they're 10 lines each, would be very helpful. I'm actually in the process of migrating from your `github-app-demo.py` repo to this repo :)

s-t-e-v-e-n-k:
Please do submit a PR!

PR #1986, opened 2021-06-30
Support full GitHub app authentication

dblanchette:
This adds support for GitHub App authentication using the main class which is simpler for users (see #1855)

```python
import Github

github_client = Github(app_id=MY_APP_ID, app_private_key=MY_PRIVATE_KEY)
```

Also, the credentials are refreshed when needed (tokens lasts for one hour). This should happen transparently to the users, as the token expiration is checked at every call.

Caveats and other notes:
- Uses the first found installation for an app. This means that it only works in the context of private apps
- I'm not sure how to add tests correctly and there does not seem to be any for `GithubIntegration`, so I skipped that part. Manual testing works and existing tests are not broken
- I could not resist fixing some type stubs along the way


github/Consts.py
```diff
@@ -125,3 +125,12 @@
 
 # https://developer.github.com/changes/2018-10-16-deployments-environments-states-and-auto-inactive-updates/
 deploymentStatusEnhancementsPreview = "application/vnd.github.flash-preview+json"
+
+
+DEFAULT_BASE_URL = "https://api.github.com"
```
dblanchette, line 130: Moved those here since they are used in 2 files now (`MainClass` and  `GithubIntegration`)

github/GithubException.py
```diff
@@ -38,7 +38,7 @@ class GithubException(Exception):
     Some other types of exceptions might be raised by underlying libraries, for example for network-related issues.
     """
 
-    def __init__(self, status, data, headers):
+    def __init__(self, status, data, headers=None):
```
dblanchette, line 41: I had a 404 while testing and had an exception because the headers parameters was not passed.

github/GithubIntegration.py
```diff
@@ -0,0 +1,140 @@
+import time
```
dblanchette, line 1: Moved from `MainClass`, it's already big as it is and that does not really belong there.

github/GithubIntegration.py
```diff
@@ -0,0 +1,140 @@
...
+            )
+        elif response.status_code == 404:
+            raise GithubException.UnknownObjectException(
+                status=response.status_code, data=response.text
+            )
+        raise GithubException.GithubException(
+            status=response.status_code, data=response.text
+        )
+
+    @deprecated.deprecated("Use get_repo_installation")
```
dblanchette, line 88: Let me know if that does not suit you. I feel that this call is badly named

github/GithubIntegration.py
```diff
@@ -0,0 +1,140 @@
...
+            f"{self.base_url}/repos/{owner}/{repo}/installation",
+            headers=headers,
+        )
+        response_dict = response.json()
+        return github.Installation.Installation(None, headers, response_dict, True)
+
+    def get_repo_installation(self, owner, repo):
+        return self.get_installation(owner, repo)
+
+    def get_installations(self):
```
dblanchette, line 112: This is the part I added in that file. Users having a private app with more than one installations can use this call as the main class always uses the first found installation.

github/Requester.py
```diff
@@ -304,6 +310,9 @@ def __init__(
     ):
         self._initializeDebugFeature()
 
+        self.__installation_authorization = None
```
dblanchette, line 313: The heart of the changes I made is here!

dblanchette (commented):
Added some comments for easier reviewing.

maxdanilov:
also very interested in seeing this implemented, are there any updates on this feature?

simkimsia:
Will this be integrated?

vitalii94lukianenko:
Hi, Could you merge this PR?

ThomasJRyan:
Would definitely appreciate this feature being merged in

github/MainClass.pyi
```diff
@@ -32,14 +30,15 @@ class Github:
...
+        app_private_key: Optional[str] = ...,
         base_url: str = ...,
         timeout: int = ...,
-        client_id: Optional[str] = ...,
-        client_secret: Optional[str] = ...,
         user_agent: str = ...,
         per_page: int = ...,
         verify: bool = ...,
         retry: Any = ...,
+        pool_size: Optional[int]=None,
```
EnricoMi, line 41: Should be ` = ...,` to not repeat the default value (see all other arguments in that signature).

EnricoMi: Same in `Requester.pyi`.

github/GithubException.py
EnricoMi: I think fixing the calling code as in #2079 is better than changing the signature.

github/GithubIntegration.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        return github.PaginatedList.PaginatedList(
+            contentClass=github.Installation.Installation,
+            requester=github.Requester.Requester(
+                login_or_token=None,
+                password=None,
+                jwt=self.create_jwt(),
+                app_id=None,
+                app_private_key=None,
+                base_url=Consts.DEFAULT_BASE_URL,
```
ThomasJRyan, line 125: If I'm not mistaken, this line makes the `base_url` passed in through the `__init__` useless. Makes it impossible to use it with Github Enterprise
```suggestion
                base_url=self.base_url,
```

dblanchette:
Thanks for the review! I made the requested changes.

github/GithubIntegration.py
dblanchette: Good catch!

dblanchette:
@EnricoMi @ThomasJRyan Anything else I should do?

github/ApplicationOAuth.pyi
```diff
@@ -13,12 +13,12 @@ class ApplicationOAuth(NonCompletableGithubObject):
     def client_secret(self) -> str: ...
     def get_login_url(
         self,
-        redirect_uri: Optional[str],
-        state: Optional[str],
-        login: Optional[str]
+        redirect_uri: Optional[str]=None,
```
EnricoMi, line 16: `.pyi` files usually do not have default values but `...` instead, default values are defined in the respective `.py` files, same issue in other files

github/Requester.py
```diff
@@ -349,6 +360,38 @@ def __init__(
         self.__userAgent = user_agent
         self.__verify = verify
 
+    def _must_refresh_token(self) -> bool:
```
EnricoMi, line 363: Should the token be refreshed some time before `expires_at` to make sure that after calling this method the request does not exceed `expired_at`? Or will e request with an expired token be retried anyway?

dblanchette: I put a 20 seconds delay (see `ACCESS_TOKEN_REFRESH_THRESHOLD_SECONDS`) to make sure there is still enough time to make the request.

github/ApplicationOAuth.pyi
dblanchette: Sorry about those, I had never used pyi files before, so I'm used to adding the hints where the default value is. Should all be fixed now.

github/Requester.py
EnricoMi: right, I misread the logic.

github/CheckSuite.pyi
```diff
@@ -47,5 +48,8 @@ class CheckSuite(CompletableGithubObject):
     def url(self) -> str: ...
     def rerequest(self) -> bool: ...
     def get_check_runs(
-        self, check_name: str, status: str, filter: str
+        self,
+        check_name: Union[str, _NotSetType]=NotSet,
```
EnricoMi, line 52: here are still some more default values in a `.pyi` file

github/Commit.pyi
```diff
@@ -45,8 +45,8 @@ class Commit(CompletableGithubObject):
     def files(self) -> List[File]: ...
     def get_check_suites(
         self,
-        app_id: Union[_NotSetType, int],
-        check_name: Union[_NotSetType, str],
+        app_id: Union[_NotSetType, int]=NotSet,
+        check_name: Union[_NotSetType, str]=NotSet,
```
EnricoMi, line 49: and here

dblanchette:
I went and look at all the files I changed to make sure there were no default values left in pyi. Crossing my fingers that everything is ok now 🤞

dblanchette:
I hate to ask again, but is it all good?

I'm looking forward to go back to using the official library instead of our fork.

Thanks for your work maintaining this library!

aabouzaid:
Any plans to merge this soon, please?

jonapich:
The last merged PR is from December 20th 2021 (3 months ago) and now there's over 60 pull requests waiting. It feels like there's a need for additional maintainers?

aaronn:
I can't seem to get my .pem file loaded into the `app_private_key=MY_PRIVATE_KEY` field with `open(file).read` nor `just the filename. Is there something I'm missing?

dblanchette:
@aaronn 
> I can't seem to get my .pem file loaded into the `app_private_key=MY_PRIVATE_KEY` field with `open(file).read()` nor just the filename. Is there something I'm missing?

I'm not really sure what's wrong. It works for us with `MY_PRIVATE_KEY` being the private key as a decoded string beginning with `-----BEGIN RSA PRIVATE KEY-----` and ending with `-----END RSA PRIVATE KEY-----` (included). The line feeds are also included and they are `\n` (not `\r\n` i.e. on Windows). Maybe you need to set an encoding when opening the file? Otherwise, I'd try with the rest API directly and make sure that works.

Hope that helps!

yskyba-kenshoo:
Hi, any updates here? Could you merge it?

alelapi:
I think there is no more need for this PR. 
GitHub App authentication can be implemented so:

```
integration = GithubIntegration(app_id, private_key)
installation_id = integration.get_installation(org, repo_name).id
access_token = integration.get_access_token(installation_id).token
github_client =  Github(login_or_token=self.access_token)
```

jnewland:
> I think there is no more need for this PR.

I'm using this PR to support [creating a GitHub App from a manifest](https://docs.github.com/en/rest/apps/apps#create-a-github-app-from-a-manifest), which expects the integrator to handle the case where an app has been installed on an account or organization but not yet installed on any repositories.

skeggse:
> I think there is no more need for this PR. GitHub App authentication can be implemented so:
> 
> ```
> integration = GithubIntegration(app_id, private_key)
> installation_id = integration.get_installation(org, repo_name).id
> access_token = integration.get_access_token(installation_id).token
> github_client =  Github(login_or_token=self.access_token)
> ```

That works fine for short-lived interactions, but if you want to interact with the Github API for hours or days, that access token will expire. It's better to have this implemented in PyGithub itself so it can handle automatic token refreshes.

alelapi:
> 

Only for my clarity's sake, could you provide an example of long interaction that cannot be replaced by using webhooks? 
GitHub forces the token's expiration time to 600 s for some reason I think

alelapi:
> > I think there is no more need for this PR. GitHub App authentication can be implemented so:
> > ```
> > integration = GithubIntegration(app_id, private_key)
> > installation_id = integration.get_installation(org, repo_name).id
> > access_token = integration.get_access_token(installation_id).token
> > github_client =  Github(login_or_token=self.access_token)
> > ```
> 
> That works fine for short-lived interactions, but if you want to interact with the Github API for hours or days, that access token will expire. It's better to have this implemented in PyGithub itself so it can handle automatic token refreshes.

Only for my clarity's sake, could you provide an example of long interaction that cannot be replaced by using webhooks? 
GitHub forces the token's expiration time to 600 s for some reason I think

cmac4603:
> Uses the first found installation for an app. This means that it only works in the context of private apps

So if you have an app installed in two organizations GitHubIntegration will no longer work? This would be a real issue for myself and potentially others I would think

skwashd:
I have a need for this. I am happy to assist with making changes needed to get it merged.

@dblanchette do you want some help with adding tests? I am also able to assist with fixing "Uses the first found installation for an app. This means that it only works in the context of private apps" so it works for multiple installations.

@s-t-e-v-e-n-k is there anything else you want to see here?

sfdye:
Looks like there are conflicts that need to be manually resolved on this PR

dblanchette:
> @dblanchette do you want some help with adding tests? I am also able to assist with fixing "Uses the first found installation for an app. This means that it only works in the context of private apps" so it works for multiple installations.

Of course, help would be appreciated. Feel free to open a PR in the coveord fork so that I can add it gets added to this one or build a new PR on top of this one.

> Looks like there are conflicts that need to be manually resolved on this PR

All resolved

ammarmallik:
> Of course, help would be appreciated. Feel free to open a PR in the coveord fork so that I can add it gets added to this one or build a new PR on top of this one.

@dblanchette here's the [PR](https://github.com/coveord/PyGithub/pull/28) that includes changes related to test cases and some new features related to fetching installation for user/repo/org. Also regarding multiple installations, currently Github doesn't have an API/functionality to generate the token that can cover multiple installations. Currrently, you can have one token per installation. As long as you select the right installation ID, you are good to go.

I've requested feedback from the community to check if there's any chance of having multiple installations feature provided by Github for Github Apps. I'll share their feedback here once I have it.

CC: @skwashd

dblanchette:
@ammarmallik I merged your changes https://github.com/coveord/PyGithub/pull/28 and they are now integratred with this PR.

I agree with requesting the installation id, even if it is one more call for our use case. Taking the first installation id could result in surprising behavior!

@sfdye Could you take a look please?

github/Requester.py
```diff
@@ -349,6 +365,47 @@ def __init__(
...
+
+    def _refresh_token(self) -> None:
+        """In the context of a GitHub app, refresh the access token"""
+        assert (
+            self.__app_id is not None
+            and self.__app_private_key is not None
+            and self.__app_installation_id is not None
+        )
+        self.__installation_authorization = self._get_installation_authorization()
+        self.__authorizationHeader = f"token {self.__installation_authorization.token}"
```
rcoup, line 407: This isn't thread-safe at the moment, and race conditions are quite likely to occur given the nature of the expiry. A few ideas:

1. an `RLock` around refreshing
2. put `_install_authorization` & `_authorization_header` in a thread-local, but that's more calls to the GH API even when there's no race.
3. do something like this using a tuple/namedtuple so if there is a refresh-race the instance attribute is assigned atomically. And change other uses of `__installation_authorization`/`__authorizationHeader` to reference the tuple.
```
install_authorization = self._get_installation_authorization()
authorization_header = f"token {self.__installation_authorization.token}"
# atomic assignment
self.__authorization = (install_authorization, authorization_header)
```

skwashd:
@sfdye can this be merged as is or does the issue flagged by @rcoup need to be resolved first?

rcoup (commented):
I'm happy with the credentials refresh, but I haven't reviewed the rest of the PR in depth. Beyond using it :-)

dblanchette:
@rcoup Would you be the one to review it?

We've been using this in production since I posted this PR in June 2021 and haven't got issues.

rcoup:
@dblanchette I'm not involved in PyGithub beyond being a user. We're using this in an internal application as well without issues.

I think @s-t-e-v-e-n-k & @sfdye are the active contributors.

s-t-e-v-e-n-k:
There are a large amount of unrelated changes in this PR, would you mind rebasing it?

dblanchette:
@s-t-e-v-e-n-k The unrelated changes are fixes around .pyi files not having the same values as the .py.

I removed them in my most recent commit so that the changes in the PR are only the ones related to GitHub App authentication. I can submit them in another PR after this one is merged.

github/GithubIntegration.py
```diff
@@ -0,0 +1,198 @@
...
+            )
+
+        body = {"permissions": permissions}
+        try:
+            headers, response = self.__requester.requestJsonAndCheck(
+                "POST",
+                f"/app/installations/{installation_id}/access_tokens",
+                input=body,
+            )
+        except UnknownObjectException:
```
s-t-e-v-e-n-k, line 131: Why are you catching and then raising these two exceptions here?

github/GithubIntegration.py
s-t-e-v-e-n-k: This is the existing call? It should probably also state it's deprecated in the docstring. Also, it should mention the class, not just the bare function.

github/GithubIntegration.pyi
```diff
@@ -0,0 +1,26 @@
...
+class GithubIntegration:
+    def __init__(
+        self,
+        integration_id: Union[int, str],
+        private_key: str,
+        base_url: str = ...,
+        jwt_expiry: int = ...,
+    ) -> None: ...
+    def _get_installed_app(self, url: str) -> Installation: ...
+    def _get_headers(self) -> dict: ...
```
s-t-e-v-e-n-k, line 16: dict of what?

github/MainClass.py
```diff
@@ -97,17 +83,26 @@ def __init__(
...
         pool_size=None,
     ):
         """
         :param login_or_token: string
         :param password: string
+        :param jwt: string
+        :param app_id: int or string
+        :param app_private_key: string
+        :param app_installation_id: int
+        :param app_token_scopes: dict
```
s-t-e-v-e-n-k, line 105: `app_token_permissions` in the signature, and `app_token_scopes` here.

github/MainClass.py
```diff
@@ -784,97 +794,3 @@ def get_app(self, slug=github.GithubObject.NotSet):
         else:
             headers, data = self.__requester.requestJsonAndCheck("GET", f"/apps/{slug}")
             return GithubApp.GithubApp(self.__requester, headers, data, completed=True)
-
-
-class GithubIntegration:
```
s-t-e-v-e-n-k, line 789: Now for my big question -- is this an API break?

github/MainClass.pyi
```diff
@@ -32,10 +30,12 @@ class Github:
         login_or_token: Optional[str] = ...,
         password: Optional[str] = ...,
         jwt: Optional[str] = ...,
+        app_id: Optional[Union[int, str]] = ...,
+        app_private_key: Optional[str] = ...,
+        app_installation_id: Optional[int] = ...,
+        app_token_permissions: Optional[dict] = ...,
```
s-t-e-v-e-n-k, line 36: dict containing what types?

github/MainClass.pyi
```diff
@@ -87,7 +86,9 @@ class Github:
     @overload
     def get_user(self, login: _NotSetType = ...) -> AuthenticatedUser: ...
     @overload
-    def get_user(self, login: str) -> NamedUser: ...
+    def get_user(
+        self, login: Union[str, _NotSetType] = ...
+    ) -> Union[NamedUser, AuthenticatedUser]: ...
```
s-t-e-v-e-n-k, line 91: This is an unrelated change, but also a welcome one.

github/Requester.pyi
```diff
@@ -47,14 +48,19 @@ class HTTPSRequestsConnectionClass:
...
+    __app_token_permissions: Optional[dict]
     def DEBUG_ON_RESPONSE(
         self, statusCode: int, responseHeader: Dict[str, str], data: str
     ) -> None: ...
     def NEW_DEBUG_FRAME(self, requestHeader: Dict[str, str]) -> None: ...
     def __check(
         self,
         status: int,
-        responseHeader: Dict[str, Any],
+        responseHeaders: Dict[str, Any],
```
s-t-e-v-e-n-k, line 63: There are still unrelated changes in this file, but I understand it's hard to untangle them.

github/SelfHostedActionsRunner.pyi
```diff
@@ -1,19 +1,19 @@
...
-    @property
-    def name(self) -> str: ...
-    @property
-    def os(self) -> str: ...
-    @property
-    def status(self) -> str: ...
-    @property
-    def busy(self) -> bool: ...
-    def labels(self) -> List[Dict[str, Union[str, int]]]: ...
+from typing import Any, Dict, List, Union
```
s-t-e-v-e-n-k, line 1: This looks unrelated?

s-t-e-v-e-n-k (changes requested):
Broadly, this is looking excellent, albeit a little terrifying. I still have some small concerns, which I've marked inline.

github/GithubIntegration.py
dblanchette: I honestly don't remember, but I see other code that does this, maybe it is to alert readers as to the exceptions that can be raised by this method?

This could also be achieved by using `:raises UnknownObjectException:` and `:raises GithubException:` in the docstring. Do you prefer I change it?

github/MainClass.pyi
dblanchette: I'll change those to `Dict[str, str]`. The GitHub reference just states `object`

github/SelfHostedActionsRunner.pyi
dblanchette: Missed that one, sorry. We added a bunch of stuff for self-hosted runners in our fork.

dblanchette: Actually, this is a change caused by wrong line endings. By loading the file, my IDE changed it from `\r\n` to `\n`. Since the latter is used all over the repo, I think this is a good change. You can see there is no actual change by checking `Hide whitespaces` in the cog wheel menu at the top of the PR.

github/GithubIntegration.py
s-t-e-v-e-n-k: We commonly catch the exceptions to change the return type -- for example `Team.get_repo_permission`, so I don't see any value for us catching the exceptions and then re-raising them. We may want to document in the docstrings which exceptions can be raised, but I'd prefer we punt that to the future.

github/MainClass.py
s-t-e-v-e-n-k: Perhaps with a deprecation message as well? Then we don't have to ship it forever.

github/MainClass.py
```diff
@@ -97,17 +83,26 @@ def __init__(
         login_or_token=None,
         password=None,
         jwt=None,
-        base_url=DEFAULT_BASE_URL,
-        timeout=DEFAULT_TIMEOUT,
+        app_id=None,
+        app_private_key=None,
+        app_installation_id=None,
+        app_token_permissions=None,
```
EnricoMi, line 89: Looks like these four go together in many places in this PR. What do you think about moving them into a class so we are carrying around only a single argument, e.g. `app_auth`.

Ideally, the now four existing ways for authentication (token, login and password, jwt, app auth) should be a single argument `auth` of a single base class (e.g. `GithubAuth`) with different implementations (`GitHubTokenAuth`, `GitHubLoginAuth`, `GitHubJwtAuth`, `GitHubAppAuth`). That would simplify a lot of signatures and stop growing signatures when new auths are added. But that is out-of-scope of this PR.

github/MainClass.py
dblanchette: I'm not sure how to add a message when a user imports from the wrong location, do you have something in mind?

github/MainClass.py
dblanchette: That is a good point and it's done :)

github/GithubIntegration.py
dblanchette: I removed the useless exception handling.

s-t-e-v-e-n-k:
Now fails tests with ` E       AttributeError: module 'github' has no attribute 'AppAuthentication'`

github/GithubIntegration.py
```diff
@@ -0,0 +1,188 @@
...
+            headers=headers,
+            attributes=response,
+            completed=True,
+        )
+
+    def create_jwt(self):
+        """
+        Creates a signed JWT, valid for 60 seconds by default. Could be set to a maximum of 600 seconds.
+        https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-a-github-app
+
```
EnricoMi, line 90: The

    valid for 60 seconds by default. Could be set to a maximum of 600 seconds.

does not make sense in this method any more, as the method does not have the `expiry` parameter. The comment should refer to `self.jwt_expiry.

This comment should go into the `__init__` to document parameter `jwt_expiry`.

github/GithubIntegration.py
```diff
@@ -0,0 +1,188 @@
...
+        :return string:
+        """
+        now = int(time.time())
```
EnricoMi, line 93: GitHub documents `iat` as

    To protect against clock drift, we recommend you set this 60 seconds in the past.

That should be another parameter to this class, which defaults to 60s.

github/Consts.py
EnricoMi: Comparing this is really difficult. Would have been better to do the refactoring (without changes) in a separate PR and base this PR on that. Then we would see only changes, not moving unchanged code around.

github/AppAuthentication.py
```diff
@@ -0,0 +1,35 @@
...
+
+class AppAuthentication:
+    def __init__(
+        self,
+        app_id,
+        private_key,
+        installation_id,
+        token_permissions=None,
+    ):
+        self.app_id = app_id
```
EnricoMi, line 32:
```suggestion
        assert isinstance(app_id, (int, str)), app_id
        assert isinstance(private_key, str)
        assert isinstance(installation_id, int), installation_id
        assert isinstance(token_permissions, int), token_permissions
        self.app_id = app_id
```

github/GithubIntegration.py
```diff
@@ -0,0 +1,188 @@
...
+        jwt_expiry=Consts.JWT_EXPIRY,
+    ):
+        """
+        :param integration_id: int
+        :param private_key: string
+        :param base_url: string
+        :param jwt_expiry: int
+        """
+        assert isinstance(integration_id, (int, str)), integration_id
+        assert isinstance(private_key, str), private_key
```
EnricoMi, line 33: Value of parameter `private_key` should not be leaked, even if it is not a string.
```suggestion
        assert isinstance(private_key, str)
```

github/Requester.py
```diff
@@ -349,6 +362,41 @@ def __init__(
...
+        if not self.__installation_authorization:
+            return
+        with self.__auth_lock:
+            if self._must_refresh_token():
+                logging.debug("Refreshing access token")
+                self._refresh_token()
+
+    def _refresh_token(self) -> None:
+        """In the context of a GitHub app, refresh the access token"""
+        assert self.__app_auth is not None
```
EnricoMi, line 396: It does not look like this method really requires `self.__app_auth is not None`, it is sufficient that `_get_installation_authorization` has that assertion. This method should not care, it is internal to `_get_installation_authorization`.

github/GithubIntegration.py
dblanchette: Replaced by a message not containing the private key

dblanchette:
> `Now fails tests with  E       AttributeError: module 'github' has no attribute 'AppAuthentication'`

That is fixed. I must say I find the imports in this project to be done in a peculiar fashion.

github/GithubIntegration.py
```diff
@@ -0,0 +1,188 @@
...
+        assert isinstance(base_url, str), base_url
+        assert isinstance(jwt_expiry, int), jwt_expiry
+        assert 15 <= jwt_expiry <= 600, jwt_expiry
```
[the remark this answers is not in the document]
dblanchette: I moved them to `Consts` where they are documented (link to the GitHub online documentation)

github/Consts.py
```diff
@@ -131,3 +131,15 @@
...
+DEFAULT_STATUS_URL = "https://status.github.com"
+# As of 2018-05-17, Github imposes a 10s limit for completion of API requests.
+# Thus, the timeout should be slightly > 10s to account for network/front-end
+# latency.
+DEFAULT_TIMEOUT = 15
+DEFAULT_PER_PAGE = 30
+
+# JWT expiry in seconds. Could be set for max 600 seconds (10 minutes).
+# https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-a-github-app
+JWT_EXPIRY = 60
```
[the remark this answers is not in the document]
dblanchette: Changed

github/GithubIntegration.py
dblanchette: Moved

github/GithubIntegration.py
dblanchette: I added a parameter. I don't think it will be used since users will probably not need to set that to a different value, but it does not hurt to have the option!

github/Requester.py
dblanchette: Indeed, removed

dblanchette:
> Comparing this is really difficult. Would have been better to do the refactoring (without changes) in a separate PR and base this PR on that. Then we would see only changes, not moving unchanged code around.

Agreed. I tend to go all out in refactoring as you've seen with the unrelated changes. I'll know to do that in the future.

Thank you both for the reviews!

github/AppAuthentication.py
```diff
@@ -0,0 +1,41 @@
...
+        private_key,
+        installation_id,
+        token_permissions=None,
+    ):
+        assert isinstance(app_id, (int, str)), app_id
+        assert isinstance(private_key, str)
+        assert isinstance(installation_id, int), installation_id
+        assert token_permissions is None or isinstance(
+            token_permissions, int
+        ), token_permissions
```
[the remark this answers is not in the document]
ammarmallik: Yes, they are optional and this is the correct type: `Optional[Dict[str, str]]`

dblanchette:
All good now? I'm excited to see this release :)

s-t-e-v-e-n-k (approved):

EnricoMi:
@dblanchette can you please take a look at #2420 /  #2425?

EnricoMi:
@dblanchette another issue likely being introduced by this PR: #2431

EnricoMi:
Another issue: #2432

EnricoMi:
This is quite some fallout, two more issues reported: #2436, #2441.

Issue #1768, opened 2020-11-30
Add 'twitter_username' 


kaustubhgupta:
Twitter username support was added this year. It would be great if you could add its access method in the AuthenticatedUser object as I checked that all other relevant access methods such as bio, avatar URL, blog link lie in this class. Also, as the twitter_username is returned in api.github.com/user/{username}, it makes sense to get this value here.

Can I contribute to this issue?

s-t-e-v-e-n-k:
Absolutely, we would welcome a PR about this issue.

kaustubhgupta:
I am not working on this issue anymore.

PR #1990, opened 2021-07-05
Feat: Add `twitter_username` property in `AuthenticatedUser`

Hanaasagi:
Ref: https://docs.github.com/en/rest/reference/users#get-the-authenticated-user

`AuthenticatedUser` should have a property `twitter_username`.


kaustubhgupta:
@Hanaasagi Link it to this issue #1768

Issue #1671, opened 2020-08-23
AuthenticatedUser.get_notifications() boolean query parameters always return true.


newswangerd:
### Description

Setting `all` and `participating` filters don't filters doesn't do anything.

`all=True` and `participating=True` should filter notifications, but instead `get_notifications` always returns the full list of notifications.

### Steps to reproduce

```
from github import Github
g = Github("secret")
user = g.get_user()

# all=False is broken
notifications = user.get_notifications(all=False)

# returns 371
print(notifications.totalCount)

notifications = user.get_notifications(all=True)

# returns 371
print(notifications.totalCount)
```

s-t-e-v-e-n-k:
Based on the GitHub API docs and the code, we're sending the correct parameters, so we're entirely at the mercy of whatever GitHub sends back.

newswangerd:
@s-t-e-v-e-n-k The github API seems to do the right thing. `https://api.github.com/notifications?all=false` only returns my unread notifications. I did notice that `https://api.github.com/notifications?all=False` (with the capital F) returns all my notifications.

It looks like the client sets the all param as a boolean: https://github.com/PyGithub/PyGithub/blob/07e29fe014b7e214e72b51129fbef55468a7309d/github/AuthenticatedUser.py#L978

Since python booleans use capital letters, I'm wondering if this getting set as `?all=False` when the params get converted to a string. If that's the case, there might be other params that are broken as well, such as `participating`

brighton1101:
I am running into this problem with `participating` and with `all`. I believe @newswangerd is correct in that the api does not assume `False == false` in the url params.

Am willing to put together a PR if you're interested in having this fixed.

s-t-e-v-e-n-k:
Sounds like we need to say `"true" if all else "false"`. I would welcome a PR! Make sure to include "Fixes #1671" in the commit message.

PR #2001, opened 2021-07-21
Fix: #1671 Convert Python Bool to API Parameter for Authenticated User Notifications

sshekdar-VMware:
Converts Python Bools `all` and `participating` to lower case string that are valid API parameters.  
`True` is converted to `"true"`

### DIFF
--- a/github/AuthenticatedUser.py
+++ b/github/AuthenticatedUser.py
@@ -973,9 +973,11 @@ class AuthenticatedUser(github.GithubObject.CompletableGithubObject):
 
         params = dict()
         if all is not github.GithubObject.NotSet:
-            params["all"] = all
+            # convert True, False to true, false for api parameters
+            params["all"] = str(all).lower() if isinstance(all, bool) else all
         if participating is not github.GithubObject.NotSet:
-            params["participating"] = participating
+            # convert True, False to true, false for api parameters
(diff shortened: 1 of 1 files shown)


github/AuthenticatedUser.py
```diff
@@ -973,9 +973,11 @@ def get_notifications(
 
         params = dict()
         if all is not github.GithubObject.NotSet:
-            params["all"] = all
+            # convert True, False to true, false for api parameters
+            params["all"] = str(all).lower() if isinstance(all, bool) else all
         if participating is not github.GithubObject.NotSet:
-            params["participating"] = participating
+            # convert True, False to true, false for api parameters
+            params["participating"] = str(participating).lower() if isinstance(participating, bool) else participating
```
s-t-e-v-e-n-k, line 980: At this point, you know `participating` is set and is a `bool`, so I think you could be simplified to:

`params["participating"] = "true" if participating else "false"`

s-t-e-v-e-n-k: And for `all` as well

sshekdar-VMware: Okay sounds good.

sshekdar-VMware:
@s-t-e-v-e-n-k

sshekdar-VMware:
@s-t-e-v-e-n-k could you please review and merge? Thank you.

PR #2006, opened 2021-07-27
Add Github Actions Secrets to organization

peresypkinamarina:
Add functionality to create and delete Secrets for Organizations.
Usage:
```
g.get_organization("test").create_secret("secret-name", "unencrypted-value", "visibility"=all/private/selected, "selected_repository_ids"=[])
g.get_organization("test").delete_secret("secret-name")
```


peresypkinamarina:
@s-t-e-v-e-n-k Could you please check it?

github/Organization.py
```diff
@@ -582,6 +582,42 @@ def create_repo(
             self._requester, headers, data, completed=True
         )
 
+    def create_secret(
+        self,
+        secret_name,
+        unencrypted_value,
+        visibility="all",
+        selected_repository_ids=[],
```
s-t-e-v-e-n-k, line 590: It feels more true to our interpretation of the API to call this `selected_repositories` and then allow users to pass either `Repository` objects or ids, and then transform them to a list of IDs when we make the PUT request.

s-t-e-v-e-n-k (changes requested):
This is looking excellent, I just have one issue before merging.

github/Organization.py
peresypkinamarina: @s-t-e-v-e-n-k Thank you for review.
I made changes to pass list of repos, similar approach as in Organization.create_team. Please check it

github/Organization.py
```diff
@@ -582,6 +582,47 @@ def create_repo(
...
+            ), selected_repositories
+
+        public_key = self.get_public_key()
+        payload = public_key.encrypt(unencrypted_value)
+        put_parameters = {
+            "key_id": public_key.key_id,
+            "encrypted_value": payload,
+            "visibility": visibility,
+        }
+        if selected_repositories:
```
s-t-e-v-e-n-k, line 616: This isn't guarded like the assert check, which means that if you don't pass visibility and do pass selected_repositories something that isn't a list of Repository objects, then we throw an exception.

Do we need to be more careful here and assert that it's an empy list (or NotSet) if visibility isn't selected?

github/Organization.pyi
```diff
@@ -66,6 +66,13 @@ class Organization(CompletableGithubObject):
         allow_merge_commit: Union[bool, _NotSetType] = ...,
         allow_rebase_merge: Union[bool, _NotSetType] = ...,
     ) -> Repository: ...
+    def create_secret(
+        self,
+        secret_name: str,
+        unencrypted_value: str,
+        visibility: str,
+        selected_repositories: List[Repository],
```
s-t-e-v-e-n-k, line 74: I think you need to mark the visibility parameter with it's default value. Also, what does type checking return if you call create_secret() with no selected_repositories?

s-t-e-v-e-n-k (changes requested):
Loving the direction, I just have a few more concerns.

github/Organization.py
peresypkinamarina: yeah, it this was not covered, thanks
Made changes to handle cases:
visibility==selected and selected_repositories is NotSet
visibility!=selected and selected_repositories is not NotSet

@s-t-e-v-e-n-k Please check it

github/Organization.pyi
peresypkinamarina: Marked and set NotSet as default value for selected_repositories

PR #2007, opened 2021-07-27
tox.ini: Ignoring long lines and setting max-line-length?

cclauss:
Better to set flake8 `max-line-length` to the actual value to discourage contributors from adding even longer lines.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,7 +36,7 @@ deps = -rrequirements.txt
 commands = sphinx-build doc build
 
 [flake8]
-max-line-length = 88
+max-line-length = 1109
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
1109, though?

cclauss:
`tests/License.py:41:89: E501 line too long (1109 > 88 characters)`

s-t-e-v-e-n-k:
In which case, we should probably fix those test cases to use textwrap -- https://docs.python.org/3/library/textwrap.html

cclauss:
There are 862 lines with > 88 characters.

Issue #2009, opened 2021-07-30
Reference links are dead


marksweb:
I'm attempting to gather stats on repo contributors (ultimately wanting to count merged PRs over a given time frame per user & rank them).

In doing so I've been looking at what information is available on `StatsContributor` but the reference link doesn't work anymore;
http://docs.github.com/en/rest/reference/repos/statistics#get-contributors-list-with-additions-deletions-and-commit-counts

The same is true for `StatsParticipation` - http://docs.github.com/en/rest/reference/repos/statistics#get-the-weekly-commit-count-for-the-repo-owner-and-everyone-else

I guess these reference docs have moved around or at least their titles have changed over time. I'll try to track down new links for these if I can.

PR #2011, opened 2021-07-31
Fixed links to github docs

marksweb:
Pretty big set of changes, but this updates all links to docs as best I can.

Fixed links to github docs. [#2009]


marksweb:
@zkoppert That's right, this is my first contribution.

I was using the docs for something - I forget what it was now - and had trouble with some links so thought I'd be helpful & update them all!! 😂

PR #2032, opened 2021-08-17
fixed bug issue and added test

### LINKED ISSUE #2030: Bug: Function create_project in class Organization isnt accessible
Add  the following line to organization.pyi

def create_project(self, param:str, body:github.GithubObject.NotSet ) -> None: ...

karthik-kadajji:
Hi,
I have fixed the bug issue in the create_project function for class Organization. The bug is resolved by modifying the organization.pyi file, due to which the function is now accessible

closes #2030 

Regards,
Karthik


github/Organization.pyi
```diff
@@ -194,6 +195,7 @@ class Organization(CompletableGithubObject):
     def remove_from_membership(self, member: NamedUser) -> None: ...
     def remove_from_public_members(self, public_member: NamedUser) -> None: ...
     def remove_outside_collaborator(self, collaborator: NamedUser) -> None: ...
+    def create_project(self, param: str, body: github.GithubObject.NotSet) -> None: ...
```
s-t-e-v-e-n-k, line 198: This should follow the existing pattern of using _NotSetType, so you can remove that import.

tests/Issue2030.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Organization(Framework.TestCase):
+    def testIssue2030CreateProject(self):
+        super().setUp()
+        project = "ultratendency"
+        self.user = self.g.get_user("karthik-kadajji-t")
+        self.org = self.g.get_organization("testkarthik")
+        self.org.create_project(project)
+        success = False
+        projects = self.org.get_projects()
```
s-t-e-v-e-n-k, line 12: This is an expensive operation, can't you look up the project by name?

s-t-e-v-e-n-k (changes requested):

tests/Issue2030.py
karthik-kadajji: No, We cannot get the project by the name. Since get_projects returns PaginatedList . But I could create a organization with no projects And then run the modified tests to fetch just the 1st element in the list .
project = self.org.get_projects()[0].name
This would assure that we could end the iteration early. Also since this is just for testing, I believe it wouldn't affect the overall performance of the library.
I will modify with the above changes and give a PR:)

karthik-kadajji: @s-t-e-v-e-n-k Would you think any more changes is need on the new PR?

PR #2036, opened 2021-08-22
link to GH docs fixed: rate_limit → rate-limit

NicoWeio:
http://docs.github.com/en/rest/reference/rate_limit does not work for me, while http://docs.github.com/en/rest/reference/rate-limit does.
So I used `grep -R "docs.github.com/en/rest/reference/rate_limit"` to find and update all occurrences of that URL. :)

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -195,7 +195,7 @@ class Github:
         """
         Rate limit status for different resources (core/search/graphql).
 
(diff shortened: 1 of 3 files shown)


NicoWeio:
While I don't think GitHub performed a _rebase_, you should now be able to merge this PR. :+1:

PR #2041, opened 2021-08-27
create_secret: also return true for status 204

fwieffering-blockfi:
204 is the status code for updating. Right now updated secrets
return `False` which doesn't distinguish between successful
and failed operations
https://docs.github.com/en/rest/reference/actions#create-or-update-a-repository-secret

> Response when creating a secret
>  Status: 201 Created
>Response when updating a secret
>  Status: 204 No Content


fwieffering-blockfi:
@s-t-e-v-e-n-k

paulzzy:
Any progress on this pull request? Would love to see this implemented!

PR #2043, opened 2021-08-30
The install commands error

mickeydarrenlau:
It should be pip install PyGithub[integrations] not pip install PyGithub['integrations']


mickeydarrenlau:
A error was made in the install command

mickeydarrenlau:
> It should be pip install PyGithub[integrations] not pip install PyGithub['integrations']

This is the error

PR #1872, opened 2021-03-11
Add support for the 'visibility' attribute on the Repository object

lightningboltemoji:
Adds support for [the 'visibility' attribute](https://developer.github.com/changes/2019-12-03-internal-visibility-changes/) on the Repository object. Closes https://github.com/PyGithub/PyGithub/issues/1446.

I updated the affected replay tests manually (not seeing a BeaverSoftware org?). I added the header to reflect the new request we'll send, and added `"visibility":"public"` next to `"private":false` in all response objects.

I also smoke tested this against our real enterprise account:

```
$ python3
Python 3.9.1 (default, Jan 30 2021, 15:51:05) 
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import github
>>> g = github.Github("<token>")
>>> repos = list(g.get_organization("my-enterprise").get_repos())
>>> repos[0].private
False
>>> repos[0].visibility
'public'
>>> repos[10].private
True
>>> repos[10].visibility
'internal'
>>> g.get_organization("my-enterprise").get_repo("my-repo").private
True
>>> g.get_organization("my-enterprise").get_repo("my-repo").visibility
'internal'
>>> g.get_organization("my-enterprise").create_repo(name="quick-test", visibility="internal")
Repository(full_name="my-enterprise/quick-test")
# verified this is internal
>>> g.get_organization("my-enterprise").create_repo(name="quick-test-2", private=True)
Repository(full_name="my-enterprise/quick-test-2")
# verified this is private
```

Thank you for the useful library and taking the time to consider this contribution!

PR #2051, opened 2021-09-07
Adding visibility field to repositories

WilliamBenEmbarek:
This PR when merged adds the visibility property to repositories which is currently missing.

This was added using the add_attribute.py script.


s-t-e-v-e-n-k:
Covered more adequately by #1872 but thank you for your submission.

PR #2052, opened 2021-09-09
make Requester thread-safe

IronLung7:
HTTPSRequestsConnectionClass and HTTPRequestsConnectionClass are not thread safe.
They can be simply use one `request` function to do the request and not saving any URL as a member parameter.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -116,23 +116,17 @@ class HTTPSRequestsConnectionClass:
         self.session.mount("https://", self.adapter)
 
     def request(self, verb, url, input, headers):
-        self.verb = verb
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
I wonder if we want to refactor that method to the base class, and then we only have to change it once

IronLung7:
@s-t-e-v-e-n-k  Sorry, which base class?

s-t-e-v-e-n-k:
I thought there was a base class, my apologies, but maybe there should be?

IronLung7:
@s-t-e-v-e-n-k  Could we merge it?

PR #2061, opened 2021-09-21
Add support to pass proxies during object init.

shanmugara:
This change allows to pass a proxy dict to Github class during object init. I needed a way to pass proxies since we are behind proxies, and this change works for us. I see other users were looking for a similar fix.

The new arg is `proxies` passed as a dict `{'http':'http://proxy.domain.som:80', 'https':'http://proxy.domain.com:80}`

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -104,6 +104,7 @@ class Github:
         verify=True,
         retry=None,
         pool_size=None,
+        proxies=None,
     ):
         """
         :param login_or_token: string
@@ -140,6 +141,7 @@ class Github:
             verify,
             retry,
             pool_size,
+            proxies
         )
 
     @property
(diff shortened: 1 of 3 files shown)


s-t-e-v-e-n-k:
If it's simple, you can use HTTP_PROXY in the environment. I will think about this change.

shanmugara:
> If it's simple, you can use HTTP_PROXY in the environment. I will think about this change.

Thank you @s-t-e-v-e-n-k . In our case we don't set proxy at system level, we use proxy in automation workflows as needed. And we use multiple proxies, depending on what resource is being accessed. So within a single automation call, we may use 2 or 3 different proxy.

s-t-e-v-e-n-k:
Maybe we need to add a test too. You could probably mock session and assert that it's had the right information set.

PR #2079, opened 2021-10-15
github/MainClass: Fix GitHubException instantiation

CyberShadow:
The mandatory "headers" argument was not being supplied.

Attempts to fix e.g.:

```
...
  File "/home/runner/actions-runner/_work/repo/repo/venv/lib64/python3.9/site-packages/github/MainClass.py", line 862, in get_access_token
    raise GithubException.GithubException(
TypeError: __init__() missing 1 required positional argument: 'headers'
```

Not sure how to test this; hope this helps nevertheless.


atodorov:
@s-t-e-v-e-n-k this seems like a serious problem. Can you merge this PR and possibly build a new version for PyPI ?

github/MainClass.py
```diff
@@ -853,14 +853,18 @@ def get_access_token(self, installation_id, user_id=None):
...
         elif response.status_code == 404:
             raise GithubException.UnknownObjectException(
-                status=response.status_code, data=response.text
+                status=response.status_code,
+                data=response.text,
+                headers=response.headers,
             )
         raise GithubException.GithubException(
-            status=response.status_code, data=response.text
+            status=response.status_code, data=response.text, headers=response.headers
```
treee111, line 867: why is this exception not formatted as the above two?
I would expect to have either 3 lines in all cases or only one line.

treee111 (changes requested):
small formatting thingy

github/MainClass.py
EnricoMi: It is probably not long enough to _require_ newlines and the linter broke the above into multiple lines but not this one.

So you are saying this should intentionally be split into multiple lines to have the same formatting as above, though it is not required.

treee111: sure, it is only visual and not functional!

If the formatting is done automatically, I am fine with it. If it is not I would prefer to have equal coding equally formated.

CyberShadow: I'm not a fan of it either, but that's how it is in this [black](https://github.com/psf/black), [black](https://github.com/PyGithub/PyGithub/blob/001970d4a828017f704f6744a5775b4207a6523c/.pre-commit-config.yaml#L2-L5) world. ;)

atodorov:
ping. Any news here? It looks like this PR is ready to be merged.

CyberShadow:
???

CyberShadow:
@s-t-e-v-e-n-k @xmo-odoo If contributions / pull requests are not welcome in this project could you please clarify this in the project documentation so we don't waste our time in the future, thank you.

xmo-odoo:
I've no idea, I just contributed a few PRs (and at least one of mine got closed by stalebot recently: #2120, #2122).

I think stevenk is just busy with stuff, that happens.

EnricoMi:
Looks like this has been fixed when #1986 moved to using the Requester, which implements those exceptions correctly.

EnricoMi:
Already fixed by #1986.

PR #2084, opened 2021-10-20
Correctly deal when PaginatedList's data is a dict

### LINKED ISSUE #2053: Bug in calculating totalCount of review requests
I want to get the count of all review requests (users and teams), but there is a problem when using the following code:

```
for pull in repo.get_pulls(state='all'):
    review_reqs = pull.get_review_requests()
    users = review_reqs[0]
    teams = review_reqs[1]
    print(users.totalCount)
    print(teams.totalCount)
```
It always prints `2` for the count of users and teams no matter what is the actual count of them.

For example, in the following case, you can see that both users and teams list is empty but again totalCount returns 2 for both of them:

image

s-t-e-v-e-n-k:
When calculating totalCount for a PaginatedList, if a Link header is not
in the returned data, we use len() to calculate the number of items.
PullRequest.get_review_requests() actually returns a dictionary, which
neatly defeats this naive method. Peer inside the dictionary in this
case, and add a test case.

Fixes #2053


Justintime50:
When is this fix getting released? This is a nasty bug that I have to code around in one of my apps. There appear to be dozens of commits on the master branch that have not yet been released. Anything we can do to speed this along?

PR #2098, opened 2021-11-03
Add support for missing branch protections

rkleinman-hpe:
Add linear history, allow force pushes, allow deletions, and required_conversation_resolution to the existing branch protections. See issue PyGithub#1716.


rkleinman-hpe:
I am not sure if I need to add new tests in a separate file, or include new tests in tests/Branch.py. Any pointers are greatly appreciated.

s-t-e-v-e-n-k:
Either is fine, whatever is easier.

treee111:
Hi @rkleinman-hpe I also would be happy to have those fields available.
Can I assist you with this PR?

jekwesco:
> Hi @rkleinman-hpe I also would be happy to have those fields available. Can I assist you with this PR?

I'd suggest doing a PR yourself since this one seems dead in the water. I'm tempted to do so because I also need these fields ;)

PR #2118, opened 2021-11-29
Improve signature of Repository.create_repo

### LINKED ISSUE #2116: Misleading `Repository.create_fork` docstring?
I'm currently exploring the behaviour of GH's edge cases, and figured PyGithub was a more convenient way to access it than raw `requests` calls, however in looking at `create_fork` I stumbled upon something odd: the docstring says

>     :param organization: string or "none" or "*"

however:

* `*` is mentioned nowhere in github's own documentation
* there seems to be no special codepath inside pygithub special-casing this, the string is just passed directly to the API
* when calling `repo.create_fork("*")` I get the same error as passing in any other non-existent organisation name:
  ```
  github.GithubException.GithubException: 422 {
      "message": "Validation Failed", 
      "errors": [{
          "resource": "Fork",
          "code": "invalid",
          "field": "organization"
      }],
      "documentation_url": "https://docs.github.com/rest/reference/repos#create-a-fork"
  }
  ```

Is this a remnant of an older revision or somesuch? I think this section of docstrings should probably be updated to:

```reStructuredText
:param str organization: name of the organisation to create the fork in, omit to fork in the current user's account
```
(after some testing, specifying the current user's login explicitely fails, and the error message explicitely notes that the error comes from it being a *user account* a opposed to an organisation)

/cc @timgates42

xmo-odoo:
The docstring stated that `"*"` and `"none"` were valid values for the
organization, but neither was actually supported (unless your
organization was called `"none"`).

Remove the misleading information from the docstring, and add support
for passing in an actual `Organization` as that's convenient for typing and clarity.

Fixes #2116


github/Repository.py
```diff
@@ -2325,15 +2325,20 @@ def get_forks(self):
...
-            organization, str
-        ), organization
         post_parameters = {}
-        if organization is not github.GithubObject.NotSet:
+        if isinstance(organization, Organization.Organization):
+            post_parameters["organization"] = organization.login
+        elif isinstance(organization, str):
             post_parameters["organization"] = organization
+        else:
+            assert organization is None \
```
s-t-e-v-e-n-k, line 2339: Why? Everywhere else in this code base we use NotSet. I know None is more Pythonic, but we don't need extra (untested!) code to handle it

github/Repository.py
```diff
@@ -2325,15 +2325,20 @@ def get_forks(self):
     def create_fork(self, organization=github.GithubObject.NotSet):
         """
         :calls: `POST /repos/{owner}/{repo}/forks <https://docs.github.com/en/rest/reference/repos#forks>`_
-        :param organization: string or "none" or "*"
+        :param organization: the name of the organization to fork into,
```
s-t-e-v-e-n-k, line 2328: This should be the type, not an explanation. 
```
:param orginization: :class:`github.Organization.Organization or string
```

s-t-e-v-e-n-k (changes requested):
It's a good start, I have some concerns inline.

github/Repository.py
xmo-odoo: I moved that into a `:type:` tag, as that was becoming quite long and an explanation as to what theb `str` case is seemed clearer.

s-t-e-v-e-n-k: Yes, but it also doesn't match every other method in this library.

github/Repository.py
xmo-odoo: Mostly because I figured it was not much and it would help caller code: currently if the caller can have an optional orgname (much like this method itself) it will have to do something like
```python
if name is None:
    repo.create_fork()
else:
    repo.create_fork(name)
```
or
```python
repo.create_fork(github.GithubObject.NotSet if name is None else name)
```
which seems even worse.

github/Repository.py
```diff
@@ -2325,15 +2325,16 @@ def get_forks(self):
...
-        :param organization: string or "none" or "*"
+        :param organization: :class:`github.Organization.Organization` or string
         :rtype: :class:`github.Repository.Repository`
         """
-        assert organization is github.GithubObject.NotSet or isinstance(
-            organization, str
-        ), organization
         post_parameters = {}
-        if organization is not github.GithubObject.NotSet:
+        if isinstance(organization, Organization.Organization):
```
s-t-e-v-e-n-k, line 2332: Surely this is NameError? We should also add a test calling create_fork with an object.

xmo-odoo:
Finally came around to writing a test, suite passes locally (at least when running `tox` straight).

Not sure how the replay files are supposed to get generated to I created one by hand copying the one from the regular call, and copying over the org data from the fork's `owner` key (as that's roughly similar, though not exactly the same payload the differences don't matter for this case) to simulate the organization fetch.

xmo-odoo:
@s-t-e-v-e-n-k are there additional things you need me to change?

xmo-odoo:
Ping, is this still of interest or should I just close it?

s-t-e-v-e-n-k (approved):

PR #2120, opened 2021-11-30
Fix type of commit for `create_file`

xmo-odoo:
The `commit` key of `create_file`'s result currently uses `github.Commit.Commit`, but that does not match the data model of the endpoint: while ill documented on docs.github.com, looking at the published schema (via https://github.com/github/rest-api-description or more readably https://github.com/octokit/types.ts which is automatically generated from the former) `file-commit.commit` looks significantly more like a `git-commit` than a `commit`:

* `commit`'s author and committer are git*hub* objects (`nullable-simple-user`), while `file-commit.commit`'s are git objects (triplets of name, email, and date)
* `commit` embeds the tree, and message in a sub-object, while `file-commit.commit` has them at the toplevel
* `file-commit.commit` has no `stats` or `files` properties, or `comments_url`

In all of these, `file-commit.commit` matches `git-commit` exactly.

The only real divergence between `git-commit` and `file-commit.commit` is that the schema indicates all of `file-commit.commit`'s properties are optional, but I think it's a case of the endpoint being under-specified and the properties *not* being marked required, rather than them being explicitly optional. See github/rest-api-description#650 for an issue on that subject with more information.


s-t-e-v-e-n-k (changes requested):
This looks amazing, and thank you for the brilliant detective work! I think part of this failure is that testCreateFile in tests/Repository.py does not check the return value of create_file() at all -- are you able to also correct that in this PR?

xmo-odoo:
@s-t-e-v-e-n-k updated with a bit of checking in the test but wasn't quite sure what to test, so I just checked if the raw data matched the GitCommit, and if it !matched Commit.

I wanted to take a look at the tree property, but that requires adding new github responses and as noted in #2118 I'm not quite sure what the "proper" way to add them is.

EnricoMi:
@xmo-odoo I have rebased your branch to the latest maser

PR #2122, opened 2021-12-02
Fix handling of `create_git_blob`'s result

xmo-odoo:
`create_git_blob` does not actually return the entire blob object content, per the OpenAPI description it only returns a `short-blob` with the `sha` and `url` properties whereas a "full" blob also has `encoding`, `content`, and `size` (and `node_id` but that's not usually relevant).

Fix by marking the `GitBlob` returned as non-complete, this way normal completion will fetch the additional data iff it is accessed.

However testing github's behaviour raises a few more questions here: `testCreateGitBlob` specifies the `latin1` encoding, per github's documentation that's not actually supported but apparently github simply treats anything other than `"base64"` as `"utf-8"`, which is really "text": the json string gets encoded to utf-8 and stored as-is in the blob. So that test's encoding is really misleading.

I figure the interface could be improved there, but it's not entirely clear to me whether `PyGithub` is to provide a shallow Python interface to Github or possibly a more high-level one e.g. one could imagine a design where `create_git_blob` would automatically select between encoding depending on the input type (`str` | `bytes`) thus making the explicit encoding unnecessary, the encoding could also be an Enum and default to text-equivalent (`"utf-8"`) with base64 being opt-in.

An other annoyance with the interface is it looks like github *always returns base64-encoded content*, again depending on the suitability of higher-level API `GitBlob` could have a utility property handling that automatically rather than requiring the user be aware of the issue.


xmo-odoo:
:(

PR #2127, opened 2021-12-15
Raise a `RateLimitExceededException` when a secondary rate limit is hit

william-richard:
I have been writing some new code that ended up hitting github's secondary rate limit,
and was confused why it was being returned as a generic `GithubException` instead of a `RateLimitExceededException`.

I dug into the code, figured out why, and put together this patch.

Let me know if there's anything I can do to improve it, especially with the replay data

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -424,6 +424,9 @@ class Requester:
             or output.get("message")
             .lower()
             .endswith("please wait a few minutes before you try again.")
+            or output.get("message")
+            .lower()
+            .startswith("you have exceeded a secondary rate limit")
         ):
             cls = GithubException.RateLimitExceededException
         elif status == 404 and output.get("message") == "Not Found":
--- a/tests/Exceptions.py
+++ b/tests/Exceptions.py
@@ -133,6 +133,17 @@ class SpecificExceptions(Framework.TestCase):
 
         self.assertRaises(github.RateLimitExceededException, exceed)
 
+    def testSecondaryRateLimitExceeded(self):
+        g = github.Github()
+
+        def exceed():
+            for i in range(100):
+                g.get_user("jacquev6")
+
+        self.assertRaisesRegexp(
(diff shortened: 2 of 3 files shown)


EnricoMi:
I have not seen your error before:
```
"You have exceeded a secondary rate limit and have been temporarily blocked from content creation. Please retry your request again later."
```

But I have seen:
```
"You have exceeded a secondary rate limit. Please wait a few minutes before you try again."
```
which is already caught properly by PyGithub due to the `.endswith("please wait a few minutes before you try again.")`.

Your message does not match as it misses that string ending. Maybe, besides adding the string starting as you suggest, it is also worth adding the string ending of your message as well to catch other errors that haven't been seen / reported yet and GitHub is likely to come up with: `.endswith("please retry your request again later.")`.

github/Requester.py
```diff
@@ -424,6 +424,9 @@ def __createException(self, status, headers, output):
             or output.get("message")
             .lower()
             .endswith("please wait a few minutes before you try again.")
+            or output.get("message")
```
EnricoMi, line 427: Maybe it is more intuitive to move that `startswith` up to the first `startswith` to collect `startswith` conditions and `endswith` conditions.

tests/Exceptions.py
```diff
@@ -133,6 +133,17 @@ def exceed():
 
         self.assertRaises(github.RateLimitExceededException, exceed)
 
+    def testSecondaryRateLimitExceeded(self):
```
EnricoMi, line 136: There are other messages that indicate secondary rate limit, worth testing them as well? https://github.com/PyGithub/PyGithub/pull/2127#issuecomment-1008842933

EnricoMi:
Included in #2387: https://github.com/PyGithub/PyGithub/pull/2387/files#diff-ba844e9d459901810597560cb9f4f5f9bca9882a58819dc14e10c38ab35df105R35

PR #2147, opened 2022-01-10
Remove enableRetry and setPoolSize from Framework

EnricoMi:
Using these methods in a test has side-effects on subsequent tests.

For instance, after calling `Framework.setPoolSize(20)` in test `PollSize.py`, all subsequent tests run with `Github.__requester.__pool_size = 20`. All tests before `PoolSize.py` run with `Github.__requester.__pool_size = None`.

The same holds for `Framework.enableRetry`.

Added a comment to avoid adding similar methods in the future (as I was about to do for `per_page`, which escaped and broke all subsequent tests).

### DIFF
--- a/tests/Framework.py
+++ b/tests/Framework.py
@@ -363,6 +363,8 @@ class TestCase(BasicTestCase):
 
 # methods like these should only be used for test session wide setups
 # not individual tests as values set here will escape tests and impact subsequent tests
+# see https://github.com/PyGithub/PyGithub/pull/2147
+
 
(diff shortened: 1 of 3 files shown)


EnricoMi:
@sfdye please consider this for reopening

EnricoMi:
Add setter methods like: https://github.com/EnricoMi/PyGithub/commit/ce1bf7ea417a56dde4c7ae2d546425fcf2c66904

EnricoMi:
Superseded by #3271.

PR #1941, opened 2021-05-05
Document difference between Commit and GitCommit

bagashvilit:
This pull request resolves Issue #1917. We added clarifying docstrings in Commit and GitCommit classes, to document the difference between the two.

PR #2152, opened 2022-01-10
Document the difference between commit and gitcommit

### LINKED ISSUE #1917: Clarify `github.Commit` vs `github.GitCommit`
I can't find any documentation, in the docs OR in comments, explaining the difference between `Commit` and `GitCommit`.

If possible, there should also be conversion functions, at least one way.

This is an issue because it seems that the return value of some functions arbitrarily use and require one of the two; for example, `Branch.commit` returns a `Commit`, but the commits passed as parent commits to `Repository.create_git_commit` expect a `GitCommit`.

treee111:
closes #1917

I squashed the changes from #1941 and rebased onto master.
#1941 should be closed after merging this PR.

### DIFF
--- a/github/Commit.py
+++ b/github/Commit.py
@@ -150,7 +150,7 @@ class Commit(CompletableGithubObject):
         position: Opt[int] = NotSet,
     ) -> CommitComment:
         """
-        :calls: `POST /repos/{owner}/{repo}/commits/{sha}/comments <https://docs.github.com/en/rest/reference/repos#comments>`_
+        :calls: `POST /repos/{owner}/{repo}/commits/{sha}/comments <https://docs.github.com/en/rest/commits/comments?apiVersion=2022-11-28#create-a-commit-comment>`_
         """
         assert isinstance(body, str), body
         assert is_optional(line, int), line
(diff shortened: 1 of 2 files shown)


makusu2 (commented):
Looks great, thank you!

stevenpitts:
Not stale, this PR is still open and waiting for review

EnricoMi:
Could you say a `Commit` is the Github commit (as in the Github concept of a commit) and a `GitCommit` is the Git commit (as in the Git concept of a commit)? Which is why `Commit` references the `GitCommit`?

treee111:
> Could you say a `Commit` is the Github commit (as in the Github concept of a commit) and a `GitCommit` is the Git commit (as in the Git concept of a commit)? Which is why `Commit` references the `GitCommit`?

yes, you could say like that.
or, in other words: `GitCommit` is the lower-level object whereas `Commit` extends it with some Github additions (comments, ...)

Do you think something has to be changed in the docs?

EnricoMi:
I would then say that the proposed changes do not reflect that, they might be even misleading. Remote server might be Git or Github, in the way it is phrased it's more like to be a Git server, when then would be wrong.

I'd suggest to refer to Git's and Github's concept of a commit, even with the "extends" bit you mentioned.

treee111:
updated the documentation and added the new URL's. Ready to review @EnricoMi

treee111:
I close this PR in favor of the unified, standardized comments of coding done in #3107.
Might also close the linked issue #1917

PR #2156, opened 2022-01-11
bypass pr review user team

yoyomeng2:
Currently PyGithub cannot set branch protection to allow bypass pull request review users or teams. This is done against the GitHub API with a PUT @ [/repos/{owner}/{repo}/branches/{branch}/protection](https://docs.github.com/en/rest/reference/branches#update-branch-protection), which `Branch.edit_protection()` already hits. I've added new parameters to that method and `Branch.edit_required_pull_request_reviews()` following existing patterns that result in including a `bypass_pull_request_reviews` object in the payload.

I am having trouble generating the ReplayData for a new test because of an auth issue, but have curated it manually to allow the test to pass. I also did some additional local testing with a project that uses PyGithub.

- support to set bypass pr review users or teams
- manually created replay data

### DIFF
--- a/github/Branch.py
+++ b/github/Branch.py
@@ -114,6 +114,8 @@ class Branch(github.GithubObject.NonCompletableGithubObject):
         strict=github.GithubObject.NotSet,
         contexts=github.GithubObject.NotSet,
         enforce_admins=github.GithubObject.NotSet,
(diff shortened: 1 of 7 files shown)


surgiie:
Im looking for this support :/

rmk135:
+1

melendezd:
I'm needing this functionality as well

PR #2172, opened 2022-02-11
Fix get events not returning user events

ncareau:
The `get_events()` function from the `AuthenticatedUser` class doesn't return events of only the user, it returns all events. 

I have modified the function to return user events only, with a flag on getting Private and Public events (`False`), or just Public events (`True`, Default). 

The original functionality is covered by `Github.get_events()`.


ncareau:
Still an issue.

hellkite500:
This definitely feels like a bug in the mapping of the python objects to the API, given that there is a `/users` event end point and a generic, public `/events` endpoint.  I was about to open an issue about this exact issue and found this PR.  If an issue would be be useful I'll create one to link to this PR.

PR #2190, opened 2022-02-28
Get all secrets from org

gerardsegarra:
Get secrets from organization


gerardsegarra:
We weren't planning on adding anything else. Do you have any suggestion?

gerardsegarra:
@qua-jones is this mergeable?

shlokabhalgat:
Hello folks, will we be having this one rolling out? I was working on a project that requires this, so was hoping we could add this one. :)

github/Organization.py
```diff
@@ -673,7 +688,7 @@ def create_secret(
         status, headers, data = self._requester.requestJson(
             "PUT", f"{self.url}/actions/secrets/{secret_name}", input=put_parameters
         )
-        return status == 201
+        return status <= 299
```
EnricoMi, line 691: @gerardsegarra can you motivate this change, please?

EnricoMi: Looks like `201` is used when creating a secret and `204` when updating a secret (calling `create_secret` on an existing secret): https://github.com/PyGithub/PyGithub/pull/2218#issue-1201786009

@jonashartwig is that your observation?

https://github.com/PyGithub/PyGithub/pull/2218/files#diff-0ccc1045a7c37896a11edf884024adbf4b46fe9c8c8b05ed2e5e64b01dea5225R1490

@nthings what do you think? Should `create_secret` fail when the secret exists already, though return status is 2xx?

I think, updating a secret should be done only through `get_secret().edit()`.

jonashartwig: yes, it is.

PR #2211, opened 2022-04-05
Added required_linear_history

aparsuri-bird:
The Branch protection edit rules did not expose `required_linear_history`. Added a bool value to set this param as part of `edit_protection` for a branch.


msteiger:
Similarly, the bypass could be added like this:
```python
if user_bypass is not github.GithubObject.NotSet:
        post_parameters["required_pull_request_reviews"]["bypass_pull_request_allowances"] = {
            "users": user_bypass,
            "teams": team_bypass
        }
```
I think that this is a valuable PR and should be merged. Maybe @s-t-e-v-e-n-k could have a quick look?

surgiie:
Looking for `bypass_pull_request_allowances` support as well. Are there plans for this?

EnricoMi:
This has been added in #2512

PR #2218, opened 2022-04-12
Calling `Repository.create_secret` may update existing secret

jonashartwig:
According to the github documentation is create and update secret the same endpoint. returning false when updating a secret (status code differently) is misleading. A quick fix would be to allow 201 and 204. Otherwise maybe another method could be there for update and check only on 204.


Bloomby (approved):
This is exactly what I came here looking to do.  Agree that this should be the way it's handled.

jonashartwig:
LGTM :)

jonashartwig:
@sfdye any update on this?

EnricoMi:
The name `create_secret` does not reflect the behaviour of this endpoint. I suggest to rename the existing method (with our change) to `create_or_update_secret` and add a `create_secret` method that calls into `create_or_update_secret` for backward compatibility.

Further, organization and environment secrets have the same behaviour, so this should be fixed there similarly.

Finally, tests for the new methods should be added.

EnricoMi:
Looks like implementation has changed. A `Secret` object is returned by `create_secret` in master. Maybe that class should get an `update` method to reflect the update, and the `create_secret` should document that existing secret with same name is overwritten.

EnricoMi (approved):
LGTM!

EnricoMi:
Looks like this is obsolete as `create_secret` has been changed in the meanwhile to return the latest `Secret` object.

Issue #2067, opened 2021-09-30
Support for Environments + Environment secrets


anovis:
Would be nice to have support for environment secrets or environments in general  

https://docs.github.com/en/rest/reference/repos#environments

https://docs.github.com/en/rest/reference/actions#get-an-environment-secret

Adib234:
Hi, I would like to tackle this feature request. The first link appears to be invalid, I've tried searching the environments section and got nothing. My approach to this would be to create a method in the Repository class for the management of environment secrets, I originally planned on having an Environment class but there seems to be not that much information about environments other than handling environment secrets in the Github API docs.

tomkennes:
HI there, thumbs up for this feature! Any update @Adib234 ? Or is there anything I can do to help out?

Adib234:
Hi, I'm not working on this feature anymore since life got in the way, it's all yours!

alson:
I needed environments, so I have [added support for environments](https://github.com/PyGithub/PyGithub/pull/2223), but not environment secrets. It should be easy to add support for environment secrets on top of this. Also, if necessary help with getting this PR ready and merged by someone more experienced in this project would be appreciated.

sandesvitor:
This feature is great @alson , it would be awesome if it got approved! Me and my team are in need for environment creation to be used as Stackstorm Github Exchange

lukaszraczylo:
@alson would you kindly try rebasing your code? Can't wait for it to be merged

alson:
@lukaszraczylo Of course, I just pushed a version rebased on top of the latest PyGithub master.

alson:
If the branch you want to limit deployment to is covered by branch protection rules, then setting `protected_branches=True` like [in this test](https://github.com/PyGithub/PyGithub/blob/main/tests/Environment.py#L137C17-L137C40) might work. See [documentation](https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment) under "properties of deployment_branch_policy". If not, then you might have to implement [deployment branch policies](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28).

The way we solved this is by just specifying branches in the Github Actions workflow and let that determine to what environment a particular branch is deployed. We are not so defensive that we want to protect against developers modifying the workflow to deploy their development branch to production.

ssbarnea:
[...]

PR #2223, opened 2022-04-20
Add support for environments

alson:
[Environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) are a feature in Github that allows setting up deployment rules, like manual approval, per repository. Because they can only be configured per repository, this makes automating their management particularly important for many-repository micro-service application. This PR adds support for getting, creating, updating and deleting environments as documented [here](https://docs.github.com/en/rest/deployments/environments).

This partially implements the features requested by [issue 2067](https://github.com/PyGithub/PyGithub/issues/2067) (environments, but not environment secrets).


alson:
The linter CI run failed, but the failure is that git is prompting for credentials to clone some packages from Gitlab. So that does not look related to my changes in any way, since I didn't touch the build or requirements files. Locally I get the same error with PyGithub:master.

willejs:
I would love to see this merged somehow!

Paul-AUB:
Same, I'm very interested in this feature !

EnricoMi (changes requested):
Looks good.

We are migrating away from `.pyi` files. Can you please merge all typing information from **new** `.pyi` files (so not `Repository.pyi`) into the corresponding `.py` files? This will help not building up tech debt.

EnricoMi (approved):
LGTM!

PR #2231, opened 2022-05-04
Fix stub for creator param in Repository.get_issues()

colinodell:
[The `creator` parameter seems to accept a string](https://github.com/PyGithub/PyGithub/blob/001970d/github/Repository.py#L2495), but the `Repository.pyi` stub does not show this as an allowed type, which results in mypy errors when a string is used.  This PR adds that missing type.

### DIFF
--- a/github/Repository.pyi
+++ b/github/Repository.pyi
@@ -379,7 +379,7 @@ class Repository(CompletableGithubObject):
         sort: Union[str, _NotSetType] = ...,
         direction: Union[str, _NotSetType] = ...,
         since: Union[_NotSetType, datetime] = ...,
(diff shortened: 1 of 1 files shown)


treee111:
same as in #2473:
if we don't want to add the .pyi files back this PR should be closed in my opinion to clean up a little bit.
@trim21 @EnricoMi

trim21:
> same as in #2473: if we don't want to add the .pyi files back this PR should be closed in my opinion to clean up a little bit. @trim21 @EnricoMi

yes, but i can't close issue

EnricoMi:
Then this should rather be fixed in `Repository.py`.

@colinodell do you mind updating your PR accordingly?

PR #1985, opened 2021-06-29
Add Artifact class and WorkflowRun.get_artifacts

yixinguo2:
Add ability to get artifacts information and download build artifacts from workflow runs.
- Added Artifact class
- Added WorkflowRun.get_artifacts
- Added WorkflowRun.test_get_artifacts
- Added tests/Artifacts.py

PR #2245, opened 2022-05-30
Add Workflow.get_artifacts

yixinguo2:
This is recreating a PR that was closed due to being stale. https://github.com/PyGithub/PyGithub/pull/1985

- create`Workflow.get_artifacts`


yixinguo2:
@s-t-e-v-e-n-k Could you take a look at your convenience? I believe I have addressed your comments in the original PR https://github.com/PyGithub/PyGithub/pull/1985

EnricoMi:
PyGithub does not return `dict`s, it returns instances of classes like `Workflow` or `WorkflowRun`, right? So `Workflow.get_artifact(run)` should return an `Artifact`.

I think this work should be based on #2319.

PR #2248, opened 2022-06-01
Add support for pulling the latest release

kiwiz:
Github allows you to hit `.../releases/latest` to fetch the latest release for a repo. [Example](https://api.github.com/repos/PyGithub/PyGithub/releases/latest). This currently doesn't work in PyGithub because passing in a string will try to hit `.../releases/tags/latest` instead.


EnricoMi (changes requested):
Please replace:

    :param id: int (release id), str (tag name)

with

    :param id: int (release id), str (tag name or "latest")

Please add a test where latest release is being pulled. If there are no `Repository.get_release[s]` tests so far, please add those as well in `tests/Repository.py`. You can fetch releases from `self.repo`.

kiwiz:
Updated. I removed some IDs from the fixtures, so hopefully they still work.

tests/ReplayData/LazyRepository.testGetReleases.txt
```diff
@@ -0,0 +1,22 @@
...
+
```
EnricoMi: This request can be removed if we swap two lines in the test:
```suggestion
```

tests/Repository.py
```diff
@@ -2275,3 +2275,24 @@ def testGetVulnerabilityAlertWhenTurnedOff(self):
     def testRequester(self):
         lazy_repo = self.getLazyRepository()
         assert lazy_repo.requester is lazy_repo._requester
+
+    def testGetReleases(self):
+        lazy_repo = self.getLazyRepository()
+        releases = lazy_repo.get_releases()
+        self.assertEqual(releases.totalCount, 2)
+        self.assertListEqual([release.tag_name for release in releases], ["v1.55", "v1.50"])
```
EnricoMi: If we swap these lines, we should be able to get away with one request (
```suggestion
        self.assertListEqual([release.tag_name for release in releases], ["v1.55", "v1.50"])
        self.assertEqual(releases.totalCount, 2)
```

EnricoMi:
Why is `Repository.get_latest_release` not sufficient?

EnricoMi (changes requested):

tests/ReplayData/LazyRepository.testGetLatestRelease.txt
```diff
@@ -0,0 +1,11 @@
+https
+GET
+api.github.com
+None
+/repos/gh_user/PyGithub/releases/latest
```
EnricoMi, line 5: I think all `gh_user/PyGithub` occurrences in these txt files should be replaced with `PyGithub/PyGithub` for the tests to work.

github/Repository.py
```diff
@@ -3582,7 +3582,7 @@ def get_releases(self) -> PaginatedList[GitRelease]:
     def get_release(self, id: int | str) -> GitRelease:
         """
         :calls: `GET /repos/{owner}/{repo}/releases/{id} <https://docs.github.com/en/rest/reference/repos#get-a-release>`_
-        :param id: int (release id), str (tag name)
+        :param id: int (release id), str (tag name or "latest")
```
EnricoMi, line 3585: I am worried that users would  not be able to get the tag `latest` any more.

PR #2274, opened 2022-07-13
[Bug fix] - Remove references to `unicode` type, no longer supported in Python3

goliaro:
The `unicode` type is no longer supported by Python3, and has been absorbed into the `str` type. Using the `unicode` type in Python3 will cause errors, so I removed all references to this type


marcorosa:
You don't just fix a unicode type reference here, but also introduce 2 (missing) workflow attributes, so you'd better specify this in the description.
By the way, the PR looks good to me and I think the maintainers can approve it.

marcorosa:
Just noticed there is #2273 open that is same as this PR, I think you can leave open just one of them not to have the same review twice

PR #2275, opened 2022-07-15
Add the ability to get codeowner errors

binford2k:


github/Repository.py
```diff
@@ -2266,6 +2267,21 @@ def get_dir_contents(self, path, ref=github.GithubObject.NotSet):
         """
         return self.get_contents(path, ref=ref)
 
+    def get_codeowner_errors(self):
+        """
+        :calls: `GET /repos/{owner}/{repo}/topics <https://docs.github.com/en/rest/repos/repos#list-codeowners-errors>`_
```
EnricoMi, line 2272: URL path here is wrong

github/Repository.py
```diff
@@ -2266,6 +2267,21 @@ def get_dir_contents(self, path, ref=github.GithubObject.NotSet):
...
+        :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.CodeownerError.CodeownerError`
+        """
+        headers, data = self._requester.requestJsonAndCheck(
+            "GET",
+            f"{self.url}/codeowners/errors",
+        )
+        return [
+            github.CodeownerError.CodeownerError(self._requester, headers, element, completed=False)
+            for element in data["errors"]
+            if element is not None and "errors" in data.keys()
```
EnricoMi, line 2282: This should be equivalent:
```suggestion
            if element is not None and "errors" in data
```

EnricoMi, line 2282: Looks like the schema of `element` does not contain a `errors` property.
```suggestion
            if element is not None
```

github/Repository.py
```diff
@@ -2266,6 +2267,21 @@ def get_dir_contents(self, path, ref=github.GithubObject.NotSet):
...
+        :calls: `GET /repos/{owner}/{repo}/topics <https://docs.github.com/en/rest/repos/repos#list-codeowners-errors>`_
+        :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.CodeownerError.CodeownerError`
+        """
+        headers, data = self._requester.requestJsonAndCheck(
+            "GET",
+            f"{self.url}/codeowners/errors",
+        )
+        return [
+            github.CodeownerError.CodeownerError(self._requester, headers, element, completed=False)
+            for element in data["errors"]
```
EnricoMi, line 2870: To make this robust if `"errors"` is not in `data`:
```suggestion
            for element in data.get("errors", [])
```

EnricoMi (changes requested):
Please add tests.

github/Repository.py
```diff
@@ -2855,6 +2856,21 @@ def get_dir_contents(self, path: str, ref: Opt[str] = NotSet) -> list[ContentFil
         """
         return self.get_contents(path, ref=ref)  # type: ignore
 
+    def get_codeowner_errors(self):
+        """
+        :calls: `GET /repos/{owner}/{repo}/topics <https://docs.github.com/en/rest/repos/repos#list-codeowners-errors>`_
+        :rtype: :class:`github.PaginatedList.PaginatedList` of :class:`github.CodeownerError.CodeownerError`
```
EnricoMi, line 2862: This does not return a `PaginatedList` but a list.

PR #2286, opened 2022-08-04
fix: black/click version issue

yoyomeng2:
Fixes `tox -e lint` running into issue where black cannot import from click.

Issue:
```
black....................................................................Failed
- hook id: black
- exit code: 1

<...>
ImportError: cannot import name '_unicodefun' from 'click' (/Users/xxx/.cache/pre-commit/repotvj361yp/py_env-python3/lib/python3.8/site-packages/click/__init__.py)
```


yoyomeng2:
Alternatively, to avoid file reformatting, pre-commit could limit click<=8.1.

sfdye:
Fixed in #2330

Issue #1753, opened 2020-11-12
Bad Credentials, token works fine with `curl`


finnkauski:
Hey folks, 

Used a personal access token as authentication, yet it does not work through python. No issues when using same token with curl, however here is the issue when using it in python.

```python
>>> gh = Github(token)
>>> gh.get_user("finnkauski")
GET https://api.github.com/users/finnkauski {'Authorization': 'token (oauth token removed)', 'User-Agent': 'PyGithub/Python'} None ==> 401 {'date': 'Thu, 12 Nov 2020 12:49:57 GMT', 'content-type': 'application/json; charset=utf-8', 'content-length': '80', 'server': 'GitHub.com', 'status': '401 Unauthorized', 'x-github-media-type': 'github.v3; format=json', 'x-ratelimit-limit': '60', 'x-ratelimit-remaining': '45', 'x-ratelimit-reset': '1605187930', 'x-ratelimit-used': '15', 'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset', 'access-control-allow-origin': '*', 'strict-transport-security': 'max-age=31536000; includeSubdomains; preload', 'x-frame-options': 'deny', 'x-content-type-options': 'nosniff', 'x-xss-protection': '1; mode=block', 'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin', 'content-security-policy': "default-src 'none'", 'vary': 'Accept-Encoding, Accept, X-Requested-With', 'x-github-request-id': 'E402:5A85:2910526:2E6F2D3:5FAD2F75'} {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/arturas/projects/gh_wrangler/.venv/lib/python3.8/site-packages/github/MainClass.py", line 271, in get_user
    headers, data = self.__requester.requestJsonAndCheck(
  File "/Users/arturas/projects/gh_wrangler/.venv/lib/python3.8/site-packages/github/Requester.py", line 317, in requestJsonAndCheck
    return self.__check(
[...]

PR #2290, opened 2022-08-12
PR for ISSUE #1753, if an auth header is provided it will stick.

ShiftedMr:
Adding in a Manual Auth Handler so that the requests Library respects the manually created auth headers that are used in the Requesters Class.

#1753 

Using the [Custom Auth Class paradigm](https://requests.readthedocs.io/en/latest/user/authentication/?highlight=Auth#new-forms-of-authentication) to bypass the chances in which the Auth header rewriting.


ShiftedMr:
closing as it wasn't making progress in months

PR #2330, opened 2022-10-13
Fix CI

sfdye:
- [x] `Jinja2>=3.1` has been failing out unit tests 
see https://github.com/sphinx-doc/sphinx/issues/10291#issuecomment-1078046986
- [x] update pre-commit hooks


.git-blame-ignore-revs
```diff
@@ -9,3 +9,6 @@
 
 # Add pyupgrade to pre-commit
 e113e37de1ec687c68337d777f3629251b35ab28
+
+Run pre-commit autoupdate and run all hooks
```
EnricoMi, line 13: This should be prefixed by `#` as a comment, right?

sfdye: Oops yes

sfdye: Fixed in https://github.com/PyGithub/PyGithub/pull/2342

PR #2353, opened 2022-11-09
fix: update GithubIntegration.get_access_token

ebarti:
The method is in need for some updating. The current body parameter, `user_id` is not listed anywhere within the Github specification.

What is [listed](https://docs.github.com/en/enterprise-cloud@latest/rest/apps/apps#create-an-installation-access-token-for-an-app), however, are three body parameters to pass in order to scope the access token to the minimum privileges it requires.


ebarti:
Depreated

PR #2356, opened 2022-11-13
Add enable() method for workflows

tituspijean:
### The Problem

Workflows disabled due to the 60-day-long period of inactivity (`state: disabled_inactivity`) need to be re-enabled before using `create_dispatch()`.

### The Solution

Implement the `enable()` that hits the `/enable` endpoint of API with a PUT request.
I have not added tests due to lack of knowledge about them, sorry.

Cheers from [YunoHost-Apps](https://github.com/YunoHost-Apps) project, that has a tad-too-many repositories to be managed one-by-one. 🙂

### DIFF
--- a/github/Workflow.py
+++ b/github/Workflow.py
@@ -149,6 +149,16 @@ class Workflow(CompletableGithubObject):
         )
         return status == 204
 
+    def enable(self) -> bool:
+        """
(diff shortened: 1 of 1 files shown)


github/Workflow.py
```diff
@@ -131,6 +131,16 @@ def create_dispatch(self, ref, inputs=github.GithubObject.NotSet):
         )
         return status == 204
 
+    def enable(self):
```
trim21, line 134: can you also add inline type？

tituspijean: Done! 🙂

EnricoMi:
Closed in favour of #3088.

PR #2382, opened 2022-12-13
CI: Update action actions/setup-python

cclauss:
https://docs.python.org/3.11/whatsnew/3.11.html
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite.


trim21:
this PR is stalled…

EnricoMi:
I think we have move to released Python3.11 already.

trim21:
> @trim21 You mean it is not reviewed?

Then I comment, the change of this PR is already made in master branch

trim21 (approved):

Borda (commented):
let's rename it to `update action actions/setup-python` as this does not touch python version at all

cclauss:
> this does not touch Python version at all

Yes.  A lot has changed in the past 7 months.

EnricoMi (approved):
LGTM!

PR #2393, opened 2022-12-30
Fix broken urls in docstrings

Muscaw:
I noticed that the URLs for accept_invitation and get_repos are broken on ReadTheDocs.

By adding the underscore at the end, it fixes the link when rendered with Sphinx


EnricoMi:
While we are there, can you also fix the missing `_` in
- Branch.py
- Repository.py
- Team.py

I found those occurrences by searching for regexp ">\`$":
```
grep -E -e ">\`$" -R github
```

Muscaw:
@EnricoMi Good catch. Thanks for the command.

EnricoMi (approved):
LGTM! @s-t-e-v-e-n-k @sfdye

PR #2397, opened 2023-01-10
Add `name`, `display_title` and `path` attributes to `WorkflowRun`

nuang-ee:
Introduce new attributes of `WorkflowRun` - see "Response Schema" of [Get a workflow run](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run)

Added attributes:
- name
- display_title
- path

---
By the way, I tried to record new `ReplayData`, since existing ones were outdated thus not containing those attributes,
But since I don't have sufficient permissions on this repository, so some tests are failing.

Could I ask a favor to maintainers (or whom that has permissions) to record & update the `ReplayData` instead of me?
Or can I just change the assertion value, assuming that tests can be run by users with insufficient permission like me?


tests/WorkflowRun.py
```diff
@@ -29,83 +29,105 @@ class WorkflowRun(Framework.TestCase):
...
+        self.assertEqual(self.workflow_run.name, "CI")
+        self.assertEqual(self.workflow_run.head_branch, "feat/workflow-run")
         self.assertEqual(
-            self.workflow_run.head_sha, "f91c729d786efcc93db47dd755313a26172c105e"
+            self.workflow_run.head_sha, "c6e5cac67a58a4eb11f1f28567a77a6e2cc8ee98"
         )
-        self.assertEqual(self.workflow_run.run_number, 162)
+        self.assertEqual(self.workflow_run.path, ".github/workflows/ci.yml")
+        self.assertEqual(self.workflow_run.display_title, "TEST PR")
+        self.assertEqual(self.workflow_run.run_number, 930)
```
EnricoMi, line 47: While we are here, can we also assert existing `run_attempt` field?
```suggestion
        self.assertEqual(self.workflow_run.run_attempt, ...)
        self.assertEqual(self.workflow_run.run_number, 930)
```

tests/WorkflowRun.py
```diff
@@ -29,83 +29,105 @@ class WorkflowRun(Framework.TestCase):
...
         self.assertEqual(self.workflow_run.event, "pull_request")
         self.assertEqual(self.workflow_run.status, "completed")
```
EnricoMi, line 49:
```suggestion
        self.assertEqual(self.workflow_run.event, "pull_request")
        self.assertEqual(self.workflow_run.run_started_at, ...)
        self.assertEqual(self.workflow_run.status, "completed")
```

EnricoMi:
> Could I ask a favor to maintainers (or whom that has permissions) to record & update the ReplayData instead of me?
> Or can I just change the assertion value, assuming that tests can be run by users with insufficient permission like me?

You could easily target your personal public repository, that should be as good as any other repo.

nuang-ee:
Sorry for late response, I'll work on your suggestions on my free time. Thanks!

PR #2402, opened 2023-01-23
python 3 updates and flake8 tidyups

marksmayo:
- added trailing commas as per python after 3.5, 3.6+
- sorted imports with isort
- replaced some strings with newer f-strings
 - re-applied black formatting
 - confirmed all tests pass
 - ran pre-commit hooks


EnricoMi:
What are the commands to reproduce / re-do this?

.hypothesis/unicode_data/14.0.0/charmap.json.gz
trim21, line 1: what's this？

PR #2409, opened 2023-02-01
Allow to add list of class attributes

EnricoMi:
Allows adding list of classes attribute:

    python scripts/add_attribute.py Commit parents list class github.Commit.Commit


EnricoMi:
Superseded by #2729.

PR #2425, opened 2023-02-16
Pass requester base URL to integration (#2420)

### LINKED ISSUE #2420: #1986 seems to break setting GitHub base URL
#1986 seems to break setting base URL.

What I did:
```
auth = AppAuthentication(int(os.getenv("GITHUB_APP_ID")),
                         t,
                         int(os.getenv("GITHUB_APP_INSTALLATION_ID")))
g = Github(base_url="https://test.somegit.com/api/v3", app_auth=auth)
```

When I turn on debugger, the base URL for requester seems to be set to api.github.com. Lmk if I am doing something wrong here, thanks!

bluek1te:
Base URL was not passed into GithubIntegration leaving it to use the one defined in const. Fixes #2420.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -312,6 +312,7 @@ class Requester:
 
         self.__installation_authorization = None
         self.__app_auth = app_auth
+        self.__base_url = base_url
 
         self.__auth_lock = RLock()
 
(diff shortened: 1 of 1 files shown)


EnricoMi:
Excellent, thanks for picking this up!

Btw., is there a way to test this?

s-t-e-v-e-n-k:
Oh wow, I've never seen that particular Github actions failure. @dblanchette what do you think, since you're rather familiar with that code?

bluek1te:
Looks like checks passed after I rebased the code

EnricoMi (approved):
LGTM!

PR #2434, opened 2023-02-23
Move CI to Python 3.11 release and 3.12 dev

EnricoMi:
Python 3.11 has been released, 3.12 is the new dev.

@s-t-e-v-e-n-k @sfdye please add `test (Python 3.11)` to the required checks when merging this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     name: test (Python ${{ matrix.python-version }})
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


sfdye:
Should we also update https://github.com/PyGithub/PyGithub/blob/master/setup.py#L101 with python 3.12?

EnricoMi:
> Should we also update https://github.com/PyGithub/PyGithub/blob/master/setup.py#L101 with python 3.12?

No, I think not, since 3.12 is not `Required` and released.

PR #2435, opened 2023-02-23
Fixing all stubs

EnricoMi:
Adds `stubtest` to tox and CI, finds 98 `.pyi` related errors.


trim21:
i have a little question about this, now we are not supporting python2.7 anymore, why use stubs instead of python3 typing?

EnricoMi:
See #2481.

trim21:
I guess we won't need this in a short time

PR #2437, opened 2023-02-25
Move RLock initialization

### LINKED ISSUE #2430: 1.58 is not compatible with AWS Lambda
1.58 introduces a dependency on multiprocessing.RLock which it not compatible with AWS Lambda. multiprocessing.SemLock (used by RLock in Requester.py) will raise OSError [Errno 38] Function not implemented. This is caused by /dev/shm not being available in AWS Lamba.

I don't have a workaround and downgraded to 1.57 which works fine.

s-t-e-v-e-n-k:
Since Requester.__init__ is called everywhere ever and may be used in restricted environments, we can not depend on semaphore locks being available. Set it to None, and initialize it when we need it, keeping it around.

Fixes #2430

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -314,7 +314,8 @@ class Requester:
         self.__app_auth = app_auth
         self.__base_url = base_url
 
-        self.__auth_lock = RLock()
+        if self.__app_auth is not None:
+            self.__auth_lock = RLock()
 
         if password is not None:
             login = login_or_token


s-t-e-v-e-n-k:
@dblanchette What do you think?

EnricoMi:
I think this does not fix the issue, it only defers it to the first call to a `request*` method. We still end up with `RLock()` raising an error.

Plus it introduces a race condition when two threads enter the method at the same time and create two different locks.

s-t-e-v-e-n-k:
Our hands are a little tied here. We only need the lock for app-based token auth, so maybe we create it only in that case.

github/Requester.py
```diff
@@ -314,7 +314,8 @@ def __init__(
         self.__app_auth = app_auth
         self.__base_url = base_url
 
-        self.__auth_lock = RLock()
+        if self.__app_auth is not None:
+            self.__auth_lock = RLock()
```
EnricoMi, line 318: What about:

```suggestion
            try:
                self.__auth_lock = RLock()
            except OSError:
                from contextlib import contextmanager

                @contextmanager
                def noop_lock():
                    yield

                self.__auth_lock = noop_lock()
```

s-t-e-v-e-n-k: We have two choices and they're both terrible -- attempting to lock if using app auth and failing on Lambda, or having no locking at all on Lambda. They are both better than the current situation of just failing no matter waht.

PR #2439, opened 2023-02-27
Add expiration argument back to GithubIntegration.create_jwt

### LINKED ISSUE #2432: v1.58.0 TypeError: create_jwt() got an unexpected keyword argument 'expiration'
The `expiration` keyword argument was removed in v1.58.0. The interface defined in GithubIntegration.pyi is no longer accurate.

EnricoMi:
Fixes #2432.


tests/GithubIntegration.py
```diff
@@ -61,6 +61,27 @@ def testCreateJWT(self):
         )
         sys.modules["time"].time = self.origin_time
 
+    def testCreateJWTWithExpiration(self):
+        self.origin_time = sys.modules["time"].time
+        sys.modules["time"].time = lambda: 1550055331.7435968
```
s-t-e-v-e-n-k, line 66: This feels awkward. Why not use a mock here?

s-t-e-v-e-n-k (changes requested):
Looking good, just one niggle.

tests/GithubIntegration.py
EnricoMi: I don't like it either, just copied from existing `testCreateJWT` above.

s-t-e-v-e-n-k: Then we should fix both places, and I'd rather circle back and fix one, rather than both.

s-t-e-v-e-n-k: I don't have time to cut a release right now, but I'll get this merged.

s-t-e-v-e-n-k (approved):

PR #2444, opened 2023-03-03
Parameter name change in create_review_comment.

lukaspetersson:
Intuitively, commit_id refers to the sha of the commit, but it is the commit object itself that is required here.


s-t-e-v-e-n-k (changes requested):
You also need to update the type information in PullRequest.pyi

s-t-e-v-e-n-k:
Furthermore, the CI failure is real: `github/PullRequest.py:422:58: F821 undefined name 'commit_id'`

PR #2455, opened 2023-03-15
Propogate ssl verify (just like base_url)

goldstar611:
It's common to have a custom root CA for self hosted github enterprise installations. Propogate the `verify` parameter just like `base_url` parameter.


EnricoMi:
Just like `base_url` and `verify`, the `Requester` inside `GithubIntegration` should also populate 
- timeout
- user_agent
- per_page
- retry
- pool_size

Duplicating the arguments passed to `Requester.__init__` in `GithubIntegration.__init__` just feels like this needs some thorough refactoring.

goldstar611:
I'm up for moving and propagating those parameters in this PR but agree with you that there should be a better way. If python provided a way to override those constants that would be ideal.

Since I knew that the `verify` changes were lined up for the next release I thought a fix for those who use self signed certs for their github enterprise installation could also make it in together.

Let me know what you want to see here and I'll work on it.

goldstar611:
whoops, I renamed the branch to correct a typo (propagate) and the PR closed here. But do let me know if `verify` was enough on its own or if all of those variables should be propagated.

EnricoMi:
I think there should be one entry point that takes `verify`, `base_url`, ..., rather than duplicating those. Further, there are some existing issues around expired JWT tokens. Please see description and first approach in #2461.

PR #2459, opened 2023-03-19
Add `name` filter to `Repository.get_artifacts()`

trim21:
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-artifacts-for-a-repository


EnricoMi (approved):
LGTM!

EnricoMi:
@trim21 can you add a test for this please?

trim21:
> @trim21 can you add a test for this please?

done

tests/Artifact.py
```diff
@@ -19,6 +19,14 @@ def testGetArtifactsFromWorkflow(self):
             repr(artifact), 'Artifact(name="vscode-codeql-extension", id=10495898)'
         )
 
+    def testGetArtifactsFromWorkflowWithName(self):
+        artifact = self.repo.get_artifacts(name="vscode-codeql-extension")[0]
```
EnricoMi, line 23: This does not test there is exactly one artifact, it only tests the first returned artifact is as expected.

tests/Artifact.py
```diff
@@ -19,6 +19,14 @@ def testGetArtifactsFromWorkflow(self):
             repr(artifact), 'Artifact(name="vscode-codeql-extension", id=10495898)'
         )
 
+    def testGetArtifactsFromWorkflowWithName(self):
```
EnricoMi, line 22: Your test name suggests this gets artifacts for a workflow. Given the name, we would expect only a single artifact. That is the reason for my confusion. Test should be renamed:

```suggestion
    def testGetArtifactsFromRepoWithName(self):
```

EnricoMi: Would you consider to add the name filter to `Repository.get_workflow_run(...).get_artifacts` as well?

tests/Artifact.py
EnricoMi: We could always move to a different repo.

EnricoMi: Assert that we expect multiple artifacts here:
```suggestion
        artifacts = self.repo.get_artifacts(name="vscode-codeql-extension")
        self.assertEqual(artifacts.totalCount, 5420)
        artifact = artifacts[0]
```

EnricoMi (changes requested):

tests/Artifact.py
trim21: Maybe later in another PR, not in this one.

trim21: after pyi are merged back to py

EnricoMi (approved):
LGTM!

PR #2461, opened 2023-03-19
Reworking GithubIntegration

EnricoMi:
The `GithubIntegration` (Github App installations) can be used as follows:

```python
gi = GithubIntegration(app_id, private_key, ...)
ri = gi.get_repo_installation(owner, repo)
gh = gi.get_github_for_installation(ri.id, permissions)
```

Then, `gh` is a `github.Github` instance, authenticated as Github App installation `ri`.

The arguments of `Github(...)` and `GithubIntegration(...)` have to stay in sync, except for those related to authentication. A test is added to assert this.

Allows `Requester(jwt=...)` to be given a method instead of a static token string. That method is called whenever a request is made. This ensures the JWT token (which has an expiry) is always up-to-date.

A `Requester(app_auth=...)` asks `app_auth._get_access_token_func` to return a function that provides an active token. Method `_get_access_token_func` reuses a requester (the calling requester) with JWT authentication. The caching and refreshing is moved into `AppAuthentication`.

All requesters use the same values for `verify`, `retry`, `base_url`, ... as given to `GithubIntegration(...)`.

Btw., current `AppAuthentication` is better called `AppInstallationAuthentication`, as it authenticates an Github App _installation_ (app id, private key, installation id). The (app id, private key) tuple is used to authentcate as an Github App (e.g. to fetch installations), so that is better called `AppAuthentication`.

With this rework, I don't see a need for users to call `create_jwt` or `get_access_token`, they will expire anyway.


trim21:
can we just have inline type annotation for new file?

EnricoMi:
Fixed by #2528 and #2556.

Issue #2462, opened 2023-03-20
Merge stubs back to source 


trim21:
Now we don't support python2 anymore, stubs is not necessary anymore, we can just use python3 typing syntax.

I'd like to contribute to that, are you open to prs that merge stubs back to source?

EnricoMi:
Waiting for maintainers thoughts on this in #2481.

Issue #1943, opened 2021-05-06
Github.oauth_scopes property is None initially


dregad:
I do not understand why _oauth_scopes_ is not set when the object is created. It only gets populated after some method or property is called.

```python
from github import Github
token = 'xxx'  # Some Github token with at least 1 scope
gh = Github(token);
print(gh.oauth_scopes)  # None
gh.rate_limiting
print(gh.oauth_scopes)  # ['public_repo']
```

Python 3.8.5, PyGithub 1.55

s-t-e-v-e-n-k:
Because we don't make a request until you call a property or a method.

dregad:
Hi Steve, thanks for the fast reply. That makes sense. 

What confused me, was that when debugging, PyCharm is showing me the protected _requester.oauth_scopes having a value:

![image](https://user-images.githubusercontent.com/449891/117408672-8e7fa680-af10-11eb-947a-47bc421561b8.png)

I guess it is PyCharm's debugger itself that triggers a request when getting the *rate_limiting* property. DBP does not show this behavior until *gh.rate_limiting* is queried.

```
> /tmp/test.py(4)<module>()->None
-> breakpoint()
(Pdb) p gh.oauth_scopes
None
(Pdb) p gh._Github__requester.oauth_scopes
None
(Pdb) p gh.rate_limiting
(4985, 5000)
(Pdb) p gh.oauth_scopes
['admin:org', 'public_repo']
```

mkenigs:
I ran into a bug because of this, because the docstring says it's a list of string: https://github.com/PyGithub/PyGithub/blob/84912a675cc09a57e947766674c462ddb0b45dbf/github/MainClass.py#L234

I think it either needs to be an empty list or the docstring should be updated.

mkenigs:
It's also pretty confusing what will cause `oauth_scopes` to become something other than `None`, particularly when comparing between a token that has readonly access to a repo:
```
> cat reproducer.py
from github import Github

gh_with_write = Github("redacted")
gh_with_write.get_user()
print(gh_with_write.oauth_scopes)
gh_with_write.rate_limiting
print(gh_with_write.oauth_scopes)

gh_readonly = Github("redacted")
gh_readonly.get_user()
print(gh_readonly.oauth_scopes)
gh_readonly.rate_limiting
print(gh_readonly.oauth_scopes)
```

```
> python3 reproducer.py
None
['gist', 'read:org', 'repo', 'workflow']
None
None
```

[...]

PR #2463, opened 2023-03-20
merge GithubObject.pyi/Requester.pyi stubs back to source

trim21:
ref https://github.com/PyGithub/PyGithub/issues/2462


github/Authorization.py
```diff
@@ -136,29 +153,23 @@ def edit(
...
-        assert remove_scopes is github.GithubObject.NotSet or all(
-            isinstance(element, str) for element in remove_scopes
-        ), remove_scopes
-        assert note is github.GithubObject.NotSet or isinstance(note, str), note
-        assert note_url is github.GithubObject.NotSet or isinstance(
-            note_url, str
-        ), note_url
         post_parameters = dict()
-        if scopes is not github.GithubObject.NotSet:
+        if not isinstance(scopes, github.GithubObject._NotSetType):
```
trim21, line 157: mypy can only narrow type for `isinstance`

github/GithubObject.py
```diff
@@ -45,23 +62,29 @@ def __repr__(self):
...
-class _ValuedAttribute:
-    def __init__(self, value):
+class _ValuedAttribute(Generic[T]):
+    value: T
+
+    def __init__(self, value: T):
         self.value = value
 
 
 class _BadAttribute:
```
trim21, line 77: this can be a subclass of `_ValuedAttribute`, so we don't need to type `Union[_BadAttribute, _ValuedAttribute]`

trim21: your
 suggestion looks good， I'll implement it

github/Authorization.py
```diff
@@ -27,106 +27,98 @@
...
+from datetime import datetime
+from typing import TYPE_CHECKING, List, Optional
 
 import github.AuthorizationApplication
 import github.GithubObject
 
+from .GithubObject import Attribute, NotSet, Opt, _NotSetType
+
+if TYPE_CHECKING:
+    from .AuthorizationApplication import AuthorizationApplication
```
EnricoMi, line 39: Can we keep absolute package names?
```suggestion
from github.GithubObject import Attribute, NotSet, Opt, _NotSetType

if TYPE_CHECKING:
    from github.AuthorizationApplication import AuthorizationApplication
```

github/Authorization.py
```diff
@@ -27,106 +27,98 @@
 # along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
 #                                                                              #
 ################################################################################
+from datetime import datetime
+from typing import TYPE_CHECKING, List, Optional
 
 import github.AuthorizationApplication
 import github.GithubObject
 
+from .GithubObject import Attribute, NotSet, Opt, _NotSetType
```
EnricoMi, line 36: We have to import `NotSet` and `_NotSetType`. Only using the former would be grand:
instead of
```
assert isinstance(scopes, _NotSetType)
```
we could use
```
assert scopes is NotSet
```
What do you think?

github/Authorization.py
```diff
@@ -27,106 +27,98 @@
...
+    def note_url(self) -> Optional[str]:
         self._completeIfNotSet(self._note_url)
         return self._note_url.value
 
     @property
-    def scopes(self):
-        """
-        :type: list of string
-        """
+    def scopes(self) -> List[str]:
```
EnricoMi, line 89: Actually, `scopes` is a string (comma-separated list of scopes): https://github.com/PyGithub/PyGithub/issues/1943#issuecomment-1582817370
```suggestion
    def scopes(self) -> str:
```
Do you want to fix this here, or in a separate PR?

github/Authorization.py
```diff
@@ -136,45 +128,43 @@ def edit(
...
-        if remove_scopes is not github.GithubObject.NotSet:
-            post_parameters["remove_scopes"] = remove_scopes
-        if note is not github.GithubObject.NotSet:
-            post_parameters["note"] = note
-        if note_url is not github.GithubObject.NotSet:
-            post_parameters["note_url"] = note_url
+        assert isinstance(note, (_NotSetType, str)), note
+        assert isinstance(note_url, (_NotSetType, str)), note_url
+
+        post_parameters = NotSet.remove_unset_items(
```
EnricoMi, line 143: I like the idea of removing all that `if note is not github.GithubObject.NotSet:` boilerplate code. What if `requestJsonAndCheck` removes all `NotSet` values from `input`, so caller does not have to create a clean `post_parameters`?

github/Authorization.py
trim21: we can't, mypy can't narrow type on `scopes is NotSet`

github/Authorization.py
trim21: I don't think so, some caller will need to transform the input. for example, in coms api user can pass `Iterator[T]` for `list[T]`, caller will have to check input anyway. Doesn't make much value to check to in `requestJsonAndCheck`

github/Authorization.py
trim21: Do you mean the real type of `.scopes` is `str` instead of `list[str]`?

trim21:
can you consider https://github.com/PyGithub/PyGithub/pull/2522 , so the ci will failed with better message?

github/Authorization.py
EnricoMi: Yes, it is a plain string `str`.

github/Authorization.py
```diff
@@ -27,106 +27,97 @@
...
     This class represents Authorizations. The reference can be found here https://docs.github.com/en/enterprise-server@3.0/rest/reference/oauth-authorizations
     """
 
-    def __repr__(self):
+    _app: Attribute["AuthorizationApplication"]
+    _created_at: Attribute[datetime]
+    _id: Attribute[int]
+    _note: Attribute[Optional[str]]
+    _note_url: Attribute[Optional[str]]
+    _scopes: Attribute[List[str]]
```
trim21, line 51:
```suggestion
    _scopes: Attribute[str]
```

github/GithubObject.py
```diff
@@ -101,89 +144,93 @@ def __init__(self, requester, headers, attributes, completed):
...
-            return _ValuedAttribute(value)
+            return _ValuedAttribute(value)  # type: ignore
         else:
-            return _BadAttribute(value, [type])
+            return _BadAttribute(value, [type])  # type: ignore
 
     @staticmethod
-    def __makeTransformedAttribute(value, type, transform):
+    def __makeTransformedAttribute(
+        value: Any, type: Type[T], transform: Callable[..., Any]
```
JLLeitschuh, line 194: Shouldn't `value` be of type `T` and callable be `Callable[[T], Any]`?

github/GithubObject.py
```diff
@@ -285,44 +347,57 @@ def format_params(params):
...
         return not self == other
 
-    def _completeIfNotSet(self, value):
-        if value is NotSet:
+    def _completeIfNotSet(self, value: Attribute) -> None:
+        if isinstance(value, _NotSetType):
             self._completeIfNeeded()
 
-    def _completeIfNeeded(self):
+    def _completeIfNeeded(self) -> None:
```
JLLeitschuh, line 389: Indentation ?

JLLeitschuh (commented):
This is a massive bit of work! I'm very impressed! Just a few minor things and I'm more than happy to merge this! 🎉

github/GithubObject.py
EnricoMi: This method returns `_BadAttribute` if `value` is not of type `T`, so `value` can be of any type.

github/GithubObject.py
trim21: what?

github/Authorization.py
```diff
@@ -136,45 +127,43 @@ def edit(
         :param note_url: string
         :rtype: None
         """
-        assert scopes is github.GithubObject.NotSet or all(
+        assert isinstance(scopes, _NotSetType) or all(
```
EnricoMi, line 130: I really don't like how this degenerates from

    assert scopes is NotSet

to

    assert isinstance(scopes, _NotSetType)

@JLLeitschuh any suggestions?

github/GithubObject.py
```diff
@@ -238,7 +293,14 @@ def _makeListOfClassesAttribute(self, klass, value):
         else:
             return _BadAttribute(value, [dict])
 
-    def _makeDictOfStringsToClassesAttribute(self, klass, value):
+    def _makeDictOfStringsToClassesAttribute(
+        self,
+        klass: "Type[GistFile]",
```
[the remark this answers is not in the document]
trim21: Yes, this method is only used in GistFile

github/AccessToken.py
```diff
@@ -21,16 +21,23 @@
...
 class AccessToken(github.GithubObject.NonCompletableGithubObject):
     """
     This class represents access tokens.
     """
 
-    def __repr__(self):
+    _token: Attribute[str]
+    _type: Attribute[str]
+    _scope: Attribute[str]
+
```
EnricoMi, line 39: I think, this should go here, but then typing becomes odd:

```
    _created: Optional[datetime]
    _expires_in = Attribute[int]
    _refresh_token = Attribute[str]
    _refresh_expires_in = Attribute[int]
```

Any ideas?

github/GithubObject.py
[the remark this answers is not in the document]
EnricoMi: But it is generic and could be used by with any klass?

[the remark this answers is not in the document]
trim21: I have no idea...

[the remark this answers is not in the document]
trim21: looks like it can `GithubObject`

github/GithubObject.py
JLLeitschuh: Never mind, it looked wrong on my phone, but it's fine.

github/GithubObject.py
```diff
@@ -41,40 +41,75 @@
...
 
+class _ValuedAttribute(Attribute):
+    def __init__(self, value: T):
+        self._value = value
 
-class _ValuedAttribute:
-    def __init__(self, value):
-        self.value = value
+    @property
+    def value(self):
```
JLLeitschuh, line 98:
```suggestion
    def value(self) -> T:
```

github/Authorization.py
JLLeitschuh: https://github.com/python/typing/issues/236

trim21: Oh yes, it impact performance.

But I don't think it's very necessary

```py
import timeit


class T:
    ...


class UnSetType(T):
    ...


unset = UnSetType()

print(timeit.timeit(lambda: unset is unset)) # 0.03375210000012885
print(timeit.timeit(lambda: isinstance(unset, UnSetType))) # 0.03873609999936889
```

and also `UnSet` is `Attribute[None]` now.

github/GithubObject.py
trim21: this is not not necessary

github/Authorization.py
trim21: using enum is possible but it will touch almost all files

JLLeitschuh: For now, let's go with what we have, and then would you be willing to open a follow-up pull request to attempt the enum changeover?

EnricoMi: I mean this aesthetically.

    assert scopes is NotSet

Is readable,

    assert isinstance(scopes, _NotSetType)

is ugly.

Plus, we need to import a private class `_NotSetType`. A lot of code has to be touched. The former was present before.

trim21: But python typing system doesn't support this.

If I use emun as single value here, this can't pass mypy:

```py
o: Attribute[Optional[int]] = NotSet
```

trim21: Also `Protocol` doesn't support generic, so make `Attribute` a `Protocol` won't work.

I don't know if there is any way to avoid all these problem.

trim21: There is a good side of this, now you can write `assert isinstance(v, (_NotSetType, str))` instead of `assert v is NotSet or instance(v, str)`

JLLeitschuh: I agree with @trim21 here. There may not be a way to make the type checker happy moving forward. Im inclined to let us poke at trying to make this better in a followup PR

JLLeitschuh (approved):
LGTM!

PR #2464, opened 2023-03-20
Merge GithubException.pyi stubs back to source

trim21:
ref https://github.com/PyGithub/PyGithub/issues/2462


github/GithubException.py
```diff
@@ -38,35 +39,35 @@ class GithubException(Exception):
...
     """
 
-    def __init__(self, status, data, headers):
+    def __init__(self, status: int, data: Any, headers: Optional[Dict[str, str]]):
         super().__init__()
         self.__status = status
         self.__data = data
         self.__headers = headers
-        self.args = [status, data, headers]
+        self.args = (status, data, headers)
```
trim21, line 47: type of `Exception.args` is `Tuple[Any, ...]` so mypy will raise a error for this

trim21, line 47: type of `Exception.args` is `Tuple[Any, ...]` so mypy will raise a error for this

EnricoMi:
Would you prefer to also improve the typing or only merge the `.pyi` into the `.py` file?

trim21:
> Would you prefer to also improve the typing or only merge the `.pyi` into the `.py` file as is?

also improve the typing, we may find type error in method function body when merging types back

trim21:
do you perfer `from typing import ...` or `import typing as t` ?

EnricoMi:
> do you perfer `from typing import ...` or `import typing as t` ?

The `.pyi` files use the former throughout this project, so I would go for that.

github/GithubException.py
```diff
@@ -38,35 +39,35 @@ class GithubException(Exception):
...
-    def data(self):
+    def data(self) -> Dict[str, Union[str, List[str], List[Dict[str, str]]]]:
         """
         The (decoded) data returned by the Github API
         """
         return self.__data
 
     @property
-    def headers(self):
+    def headers(self) -> Union[None, Dict[str, str]]:
```
EnricoMi, line 64: I think `Optional[T]` is better than `Union[None, T]`.

github/GithubException.py
```diff
@@ -99,27 +100,44 @@ class BadAttributeException(Exception):
     Exception raised when Github returns an attribute with the wrong type.
     """
 
-    def __init__(self, actualValue, expectedType, transformationException):
+    def __init__(
+        self,
+        actualValue: Any,
+        expectedType: Union[
```
EnricoMi, line 106: Is this type really correct / complete?

github/GithubException.py
```diff
@@ -38,35 +39,35 @@ class GithubException(Exception):
...
-    def status(self):
+    def status(self) -> int:
         """
         The status returned by the Github API
         """
         return self.__status
 
     @property
-    def data(self):
+    def data(self) -> Dict[str, Union[str, List[str], List[Dict[str, str]]]]:
```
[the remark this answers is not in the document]
trim21: I'm not 100% confident about these types, these are copied from pyi, I'm not sure how it's used,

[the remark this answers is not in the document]
trim21: we can fix this later, when they are used in source, I think

EnricoMi (approved):

PR #2466, opened 2023-03-20
Add query `check_suite_id` integer to `Workflow.get_runs`

trim21:
https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-required-workflow


EnricoMi (approved):
LGTM!

PR #2471, opened 2023-03-23
Add support for get_app endpoint with app authentication

chantra:
add support for get_app endpoint for app authenticated applications
    
The /app endpoint requires the use of a JWT to be accessed: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-the-authenticated-app
    
This change creates a new GithubIntegration and calls the `/app`
endpoint through it so the `requester` used uses JWT and not a token.
    
* Updated exiting GithubApp testGetAuthenticatedApp to use authentication
* added test to validate assertion is raised when not called by an authenticated app.
    
Prior to this, an authenticated application would fail to get access to
its attribute with the following error: https://gist.github.com/chantra/f84e6981e2ab732a5e7acb2978872109

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -381,6 +381,15 @@ class Requester:
             permissions=self.__app_auth.token_permissions,
         )
 
+    def get_app(self):
+        assert self.__app_auth is not None
+        integration = GithubIntegration.GithubIntegration(
+            self.__app_auth.app_id,
+            self.__app_auth.private_key,
+            base_url=self.__base_url,
+        )
+        return integration._get_app()
+
     def _refresh_token_if_needed(self) -> None:
         """Get a new access token from the GitHub app installation if the one we have is about to expire"""
         if not self.__installation_authorization:
--- a/github/GithubIntegration.py
+++ b/github/GithubIntegration.py
@@ -4,6 +4,7 @@ import deprecated
 import jwt
 
 from github import Consts
+from github.GithubApp import GithubApp
 from github.GithubException import GithubException
 from github.Installation import Installation
 from github.InstallationAuthorization import InstallationAuthorization
@@ -84,6 +85,24 @@ class GithubIntegration:
         )
 
         return Installation(
+            requester=self.__requester,
+            headers=headers,
+            attributes=response,
+            completed=True,
+        )
+
+    def _get_app(self):
+        """
+        Get app.
+
+        :param url: str
+        :rtype: :class:`github.GithubApp.GithubApp`
+        """
+        headers, response = self.__requester.requestJsonAndCheck(
+            "GET", "/app", headers=self._get_headers()
+        )
+
+        return GithubApp(
             requester=self.__requester,
             headers=headers,
             attributes=response,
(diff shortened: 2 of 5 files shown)


github/Requester.py
```diff
@@ -381,6 +381,15 @@ def _get_installation_authorization(self):
             permissions=self.__app_auth.token_permissions,
         )
 
+    def get_app(self):
```
EnricoMi, line 384: I think this method should go into `GithubIntegration`.

EnricoMi: I'd rather say GithubIntegration requires some rework to get the authentication, token refresh and Requester init args sorted out: https://github.com/PyGithub/PyGithub/pull/2461

chantra:
> In github/Requester.py
> <https://github.com/PyGithub/PyGithub/pull/2471#discussion_r1148567224>:
>
> > @@ -381,6 +381,15 @@ def _get_installation_authorization(self):
>              permissions=self.__app_auth.token_permissions,
>          )
>
> +    def get_app(self):
>
> I'd rather say GithubIntegration requires some rework to get the
> authentication, token refresh and Requester init args sorted out: #2461
> <https://github.com/PyGithub/PyGithub/pull/2461>
>
> thanks for sorting out the building blocks. I will rebase and adjust the
PR accordingly.

chantra:
Ok, attempted https://github.com/chantra/PyGithub/commit/1154f22b78facef394f1c1407e44408634afaf2a

There is an issue with being able to call github.get_app() through as I don't see how we can get the GithubIntegration back to call GithubIntegration.get_app().

Also, I believe, if calling using the Github's requester, then the token will be sent, not the jwt, hence the need to be able to get back to the GithubIntegration form Github.

Another solution is to make `get_app` non-valid call if the slug is not provided, forcing to use the GithubIntegration object only.

chantra:
@EnricoMi do you have any feedback on my previous comment?

chantra:
@EnricoMi  just checking if you have any feedback on my previous comment? thanks

EnricoMi:
I think calling the `/app` endpoint should be moved into `GithubIntegration.get_app` and `Main.get_app` should make `slug` mandatory. This separates the two different endpoints. The former requires a personal access token or app installation token, the latter an application jwt.

Then, together with rework #2461, the following should work:

```
gh = Github(token)
# calls /apps/{slug} with personal access token
gh.get_app(slug)

gi = GithubIntegration(app_id, private_key, ...)
# calls /app with jwt
gi.get_app()

gh = gi.get_github_for_installation(id, permissions)
# calls /apps/{slug} with installation access token
gh.get_app(slug)
```

chantra:
Thanks @EnricoMi . I would be happy to do those changes once your branch is merged. Please tag me when it lands so I can check it out and update this PR.

EnricoMi:
@chantra auth rework is merged, go ahead and move `/app` call into `GithubIntegration`. Maybe a fresh new PR might be easier.

chantra:
Thanks @EnricoMi 

Let me close this PR and start fresh.

PR #2473, opened 2023-03-24
fix(types): :label: align push restriction change stubs with impl

lbillinghamwrk:
In #1290 we changed the verbs in the methods for `def .*_{user,team}_push_restrictions` from `edit` to `replace`  It looks like we missed that change to the method implimentation in the typestubs. This PR re-synchronizes them.

implimentations at https://github.com/PyGithub/PyGithub/blame/master/github/Branch.py#L439 and https://github.com/PyGithub/PyGithub/blame/master/github/Branch.py#L472 

(as far as I can tell, the pyi stub files didn't exist at time of #1290)


trim21:
@EnricoMi I think we should close this issue

lbillinghamwrk:
> @EnricoMi I think we should close this issue

because of the conflicts or some other reason?
I can fix conflicts in the fork if y'all want

trim21:
> > @EnricoMi I think we should close this issue
> 
> because of the conflicts or some other reason? I can fix conflicts in the fork if y'all want

file `github/Branch.pyi` has been removed

treee111:
> file `github/Branch.pyi` has been removed

well if we don't want to add the `.pyi` files back this PR should be closed in my opinion to clean up a little bit.
@trim21 @EnricoMi

EnricoMi:
I think [Branch.py](https://github.com/PyGithub/PyGithub/blob/main/github/Branch.py) is up-to-date.

PR #2475, opened 2023-03-27
added method to remove team from repository

yd-wu:


github/Team.py
```diff
@@ -305,6 +305,24 @@ def update_team_repository(self, repo, permission):
         )
         return status == 204
 
+    def remove_team_repository(self, repo):
```
trim21, line 308: Please add imline pep484 type

github/Team.py
```diff
@@ -305,6 +305,24 @@ def update_team_repository(self, repo, permission):
...
+        """
+        :calls: `DELETE /organizations/{org_id}/team/{team_id}/repos/{owner}/{repo} <https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#remove-a-repository-from-a-team>`_
+        :param repo: string or :class:`github.Repository.Repository`
+        :rtype: bool
```
EnricoMi, line 312:
```suggestion
```
@trim21 these lines can go away given the inline types, right?

EnricoMi (changes requested):
Adding a test for this would be grand!

PR #2483, opened 2023-03-30
Add support for new RepositoryAdvisories API :tada:

JLLeitschuh:
Related:
 - https://github.blog/changelog/2023-03-30-repository-security-advisories-rest-api/
 - https://docs.github.com/en/rest/security-advisories/repository-advisories

Signed-off-by: Jonathan Leitschuh <Jonathan.Leitschuh@gmail.com>


JLLeitschuh:
CI passes. Linting passses. AFAIK, everything has solid unit test coverage.

Seeking code reviews!

github/Repository.py
```diff
@@ -1448,6 +1452,89 @@ def __create_pull(self, **kwds):
...
+        self,
+        summary: str,
+        description: str,
+        severity_or_cvss_vector_string: str,
+        cve_id: typing.Optional[str] = None,
+        vulnerabilities: typing.Optional[
+            typing.Iterable[
+                github.RepositoryAdvisoryVulnerability.AdvisoryVulnerability
+            ]
+        ] = None,
```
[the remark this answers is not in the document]
JLLeitschuh: Turns out this isn't a bug. It's implemented as-designed

rthorpeii (commented):
The structure of the API objects and calls looks good to me!

github/Repository.py
```diff
@@ -1448,6 +1487,162 @@ def __create_pull(self, **kwds):
             self._requester, headers, data, completed=True
         )
 
+    def create_repository_advisory(
```
EnricoMi, line 1490: since this is on `Repository`, this could read:
```suggestion
    def create_advisory(
```
Or are there other methods that imply this naming convention?

github/Repository.py
```diff
@@ -1448,6 +1487,162 @@ def __create_pull(self, **kwds):
...
+            description=description,
+            severity_or_cvss_vector_string=severity_or_cvss_vector_string,
+            cve_id=cve_id,
+            vulnerabilities=vulnerabilities,
+            cwe_ids=cwe_ids,
+            credits=credits,
+            private_vulnerability_reporting=False,
+        )
+
+    def report_security_vulnerability(
```
EnricoMi, line 1528: I think this should read:
```suggestion
    def report_security_advisory(
```
or
```suggestion
    def create_security_advisory(
```

github/Repository.py
```diff
@@ -2190,6 +2385,35 @@ def create_file(
             ),
         }
 
+    def get_repository_advisories(
```
EnricoMi, line 2388: Here, `repository` could also be removed, but on the other hand, it returns a `RepositoryAdvisory.

EnricoMi (commented):
LGTM, minor comments

github/Repository.py
JLLeitschuh: This API is used by reporters, not by maintainers, to report vulnerabilities.

I was trying to figure out a way to differentiate the `create` which is something only a repository admin can take, from the `report` which is an action any GitHub user can take.

github/Repository.py
```diff
@@ -1448,6 +1487,162 @@ def __create_pull(self, **kwds):
...
+                github.RepositoryAdvisoryVulnerability.AdvisoryVulnerability
+            ]
+        ] = None,
+        cwe_ids: typing.Optional[typing.Iterable[str]] = None,
+        credits: typing.Optional[
+            typing.Iterable[github.RepositoryAdvisoryCredit.RepositoryAdvisoryCredit]
+        ] = None,
+    ) -> github.RepositoryAdvisory.RepositoryAdvisory:
+        """
+        :calls: `POST /repos/{owner}/{repo}/security-advisories/reports <https://docs.github.com/en/rest/security-advisories/repository-advisories#create-a-repository-security-advisory>`_
```
EnricoMi, line 1545: I think the link should point to

    https://docs.github.com/en/rest/security-advisories/repository-advisories#privately-report-a-security-vulnerability

github/Repository.py
EnricoMi: right, the link to the docs confused me, I think it is linking to creating an advisory, where it should link to creating a vulnerability

EnricoMi:
@JLLeitschuh happy to approve, please rebase and correct the doc links

JLLeitschuh:
Will do, just been bogged down in other work

github/Repository.py
JLLeitschuh: So does the one above. Which naming convention would you like to see used here?

github/Repository.py
JLLeitschuh: Done!

github/Repository.py
JLLeitschuh: Unless you have a strong preference, I think we should opt to use the naming convention in the GitHub API docs here.

EnricoMi (approved):
LGTM!

PR #2502, opened 2023-04-20
Add type stub for get_project_column

Railslide:
`get_project_column` was missing from the type stubs, so both mypy and pyright complained when using the method.


EnricoMi (approved):
LGTM!

PR #2477, opened 2023-03-28
Add allow_auto_merge support to Repository

BradChengIRESS:
Closes #2476 

Adds `allow_auto_merge` attribute to a Repository: https://docs.github.com/en/rest/repos/repos#update-a-repository

> `allow_auto_merge` boolean
> Either `true` to allow auto-merge on pull requests, or `false` to disallow auto-merge.

This attribute can be modified via edit (on a private repository):

```
>>> g.get_repo("<test repo>").allow_auto_merge
True
>>> g.get_repo("<test repo>").edit(allow_auto_merge=False)
>>> g.get_repo("<test repo>").allow_auto_merge
False
>>> g.get_repo("<test repo>").edit(allow_auto_merge=True)
>>> g.get_repo("<test repo>").allow_auto_merge
True
```

PR #2516, opened 2023-05-05
Add support for allow_auto_merge option for Repository.edit

### LINKED ISSUE #2456: Repository.edit(allow_auto_merge=false)
Github API now supports https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#update-a-repository `allow_auto_merge` option. Please implement the same for PyGithub.

yd-wu:
Fixes #2456.


EnricoMi:
Nice! While you are there, can you re-record the replay data for Repository.setUp and test for all implemented attributes, not just the new `allow_auto_merge`? That would be grand!

EnricoMi:
Looks like we have two pull requests for the same feature: #2477

EnricoMi:
CC @JLLeitschuh

tests/Framework.py
```diff
@@ -338,7 +338,9 @@ def __closeReplayFileIfNeeded(self):
             if (
                 not self.recordMode
             ):  # pragma no branch (Branch useful only when recording new tests, not used during automated tests)
-                self.assertEqual(readLine(self.__file), "")
+                self.assertTrue(
+                    readLine(self.__file) == "200" or readLine(self.__file) == ""
+                )
```
EnricoMi, line 343: What is the use case for this change? This impacts almost all tests.

EnricoMi:
Are you happy with #2477 to go ahead?

EnricoMi:
This has been implemented in #2477.

PR #2522, opened 2023-05-08
[CI] support mypy/pytest github pr annotations

trim21:
image

Now we want to add inlint type annotation, this would be very useful

---

also add pytest-github-actions-annotate-failures for pytest

https://github.com/pytest-dev/pytest-github-actions-annotate-failures


tox.ini
```diff
@@ -7,7 +7,7 @@ envlist =
 [gh-actions]
 python =
     3.7: py37
-    3.8: py38, docs, lint
```
EnricoMi, line 10: docs and lint are not run at the user side, only in CI, why is that?

trim21:
```text
(.venv) ❯❯ ~\..\PyGithub git:(pr/2463) tox
ROOT: tox-gh-actions won't override envlist because tox is not running in GitHub Actions
lint: commands[0]> pre-commit run --all-files --show-diff-on-failure
```

trim21: /ping

EnricoMi: Never mind, you are right.

tox.ini
```diff
@@ -7,7 +7,7 @@ envlist =
 [gh-actions]
 python =
```
EnricoMi, line 8: Looks like this list is useless, when running tox on github actions with python versions that are not listed her still runs them:
- 3.12-dev: https://github.com/PyGithub/PyGithub/actions/runs/5259774567/jobs/9505791611#step:5:26
- 3.11: https://github.com/PyGithub/PyGithub/actions/runs/5259774567/jobs/9505791522#step:5:26

So what is the point in maintaining this list?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,44 @@
+name: CI
```
EnricoMi, line 1:
```suggestion
name: Lint
```

tox.ini
trim21: oh， this make sure on ci job only run tests once

EnricoMi: right, I guess one entry is sufficient then

trim21: Wait, I'll test if it will run multiple tests in one ci job.

trim21: This is necessary.

comment added, also notice developer to keep python version in ci and tox.ini synced

EnricoMi (approved):
LGTM! Thanks for tightening this!

EnricoMi:
Thanks a lot!

Issue #2447, opened 2023-03-08
Authenticating with a GitHub App not working with custom URL


AlexHff:
Hello there,

It seems like the `base_url` passed to the default constructor is not used to fetch the access token for app authentication.

Here's an example code:
```
app_id = "XXX"
app_private_key = "XXX"
installation_id = "XXX"
g = github.Github(
    base_url="https://custom-github-url.com"
    app_auth=github.AppAuthentication(
        app_id=app_id,
        private_key=app_private_key,
        installation_id=installation_id,
    ),
)
```

Whatever `base_url` is passed, the following error occurs:
`requests.exceptions.ConnectionError: HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url: XXX`

As can be seen in the stack trace, the host is the one defined [here](https://github.com/PyGithub/PyGithub/blob/5e27c10a3140c3b9bbf71a0b71c96e71e1e3496c/github/Consts.py#L135), which is `DEFAULT_BASE_URL = "https://api.github.com"`.

I suspect the error comes from [here](https://github.com/PyGithub/PyGithub/blob/5e27c10a3140c3b9bbf71a0b71c96e71e1e3496c/github/Requester.py#L377). The param `base_url` is not passed to `GithubIntegration` which therefore uses the default url.

EnricoMi:
This is been reported in #2420 and fixed in #2425. Will be released in [v0.58.1](https://github.com/PyGithub/PyGithub/milestone/39).

goldstar611:
I tested with master today and while `base_url` is propagated, `verified` is not. `verified` is a bool or string and the string is needed when the enterprise installation uses a SSL/TLS certificate that is not issued by a root ca in pythons included ca bundle.

We're getting there though!

EnricoMi:
The fix has been released.

EnricoMi:
> but then while refreshing the access token, a check for expiring access token is done which then refreshes the access token, which checks the token ... until recursion limit is met.

@goldstar611 I could not reproduce this issue, and looking into the code, I could not see how this would recurse. The refresh request creates its own new requester instance. Can you provide a stack trace? That would help to investigate the issue.

goldstar611:
I tested both the `Github` class with a personal access token as well as the `GithubIntegration` class with a private key and did not see this issue again.

[...]

PR #2524, opened 2023-05-09
Raise error on unsupported redirects, log supported redirects

EnricoMi:
Redirects that do not change the url path but other pieces like host name or protocol schema end up in an endless loop because only the path part of the new location is redirected to: https://github.com/PyGithub/PyGithub/issues/2447#issuecomment-1539048322

This identifies situations where hostname, protocol or port number change in a redirect. Instead of redirecting to the new location, an error is raised because the base url should be corrected. Otherwise, each request will be redirected, duplicating the traffic and calls.

Further, any redirection is logged to info level. There have been many issues where knowing a redirect happened (without enabling debug level) would have helped investigating and understanding.


github/Requester.py
```diff
@@ -617,7 +617,30 @@ def __requestRaw(self, cnx, verb, url, requestHeaders, input):
...
+                    f"please correct your Github server URL via base_url: Github(base_url=...)"
+                )
+            if o.path == url:
+                port = ":" + str(self.__port) if self.__port is not None else ""
+                requested_location = f"{self.__scheme}://{self.__hostname}{port}{url}"
+                raise RuntimeError(
+                    f"Requested {requested_location} but server redirected to {location}, "
+                    f"you may need to correct your Github server URL "
+                    f"via base_url: Github(base_url=...)"
+                )
```
JLLeitschuh, line 639: This looks like it may be fixing a potential security vulnerability where credentials could be leaked to the server being redirected to? Do you agree?

This would be similar to this vulnerability in curl: https://curl.se/docs/CVE-2022-27774.html

goldstar611: [I discovered an infinite loop](https://github.com/PyGithub/PyGithub/issues/2447#issuecomment-1539048322) this PR fixes with a self hosted Github Enterprise Server (GHES) instance, it was not due to a vulnerability. Our appliance is reachable under several host names but the default GHES option is to send a 302/301 redirect if the `Host` header doesn't match the expected value. This can be seen if you attempt to connect to `www.github.com` with the simple reproducer:

```python
from github import Github

# Github Enterprise with custom hostname
g = Github(base_url="https://www.github.com/api/v3",
#                            ^^^^ notice the www.
           login_or_token="doesn't matter what it is",)
print(g.get_user().name)
```

I welcome the RuntimeErrors seen here because this happens periodically and is hard to remember that the infinite loop is caused by a mismatched host name value.

EnricoMi: Yes, this fix is not trying to fix a vulnerability. Such redirects caused an infinite loop and a non-descriptive error. So they now fail on the first response and explain the situation.

PyGithub redirects any 301 response and sends the headers to the new location, but it only considered the path of the new location. So it reused the original hostname and port and protocol, so did not leak anything. Now it only redirects if hostname and port and protocol are identical, so there is no functional change (except for better error messages in the error case).

JLLeitschuh: Sounds good to me!

JLLeitschuh:
Feel free to merge

EnricoMi:
Thanks!

Issue #2504, opened 2023-04-21
Installation.get_repos() appears to be broken - always throws BadCredentialsException


eap:
Is there anyone using this method? The only other reference to my issue is two comments in this stale pull request #1791 which seems to agree with my findings. Alternatively, we are both guilty of user error.

To recreate the issue:
1) Create a github app with read access to repositories.
2) Install it in your user account granting access to all repositories, or at least one private repo used for this example.
3) Create and download an application private key ".pem" file.
4) Note the application ID.

Initialize the integration and installation. Ensure that the installation credentials work for mundane requests (get repo).
```
$ python3
Python 3.11.2 (main, Feb 16 2023, 03:07:35) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>> from github import *
>> with open("mykeyfile.pem", "r") as f:
...  private_key = f.read().strip()

>> app_id = "123456" // use your application id.
>> integration = github.GithubIntegration(app_id, private_key)
>> installation = integration.get_repo_installation('myuser', 'my-private-repo')
>> token = integration.get_access_token(installation.id).token
>> g = Github(token)
>> g.get_repo('myuser/my-private-repo')
Repository(full_name="myuser/my-private-repo")
```

The failure

```
>> print('Using the same installation, I get this failure')
>> installation.get_repos()
<github.PaginatedList.PaginatedList object at 0x11248d450>
>>> hitting_the_api_now = [r for r in installation.get_repos()]
File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <listcomp>
  File "/users/Ubuntu/test_error/.venv/lib/python3.11/site-packages/github/PaginatedList.py", line 56, in __iter__
    newElements = self._grow()
                  ^^^^^^^^^^^^
  File "/users/Ubuntu/test_error/.venv/lib/python3.11/site-packages/github/PaginatedList.py", line 67, in _grow
    newElements = self._fetchNextPage()
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/users/Ubuntu/test_error/.venv/lib/python3.11/site-packages/github/PaginatedList.py", line 201, in _fetchNextPage
    headers, data = self.__requester.requestJsonAndCheck(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/Ubuntu/test_error/.venv/lib/python3.11/site-packages/github/Requester.py", line 398, in requestJsonAndCheck
[...]

PR #2528, opened 2023-05-10
Add authentication classes, move auth logic there

### LINKED ISSUE #2431: 401 unauthorized after upgrading to 1.58.0
We use `GithubIntegration.get_access_token` to authenticate and after upgrading we now get this response:

```
github.GithubException.GithubException: 401 {"message": "'Expiration time' claim ('exp') must be a numeric value representing the future time at which the assertion expires", "documentation_url": "https://docs.github.com/rest"}
```

Reverting to 1.57 solves the issue.

EnricoMi:
Reduces complexity from Requester and moves lots of auth-specific logic into `Auth` implementations.

Fixes #2431.

Allows for easy fix of #2471 and #2504: https://github.com/EnricoMi/PyGithub/pull/4.

Fixes bad code like

https://github.com/PyGithub/PyGithub/blob/b94a83cbf328bbcdabeaaaef572d5ca360e1dbba/github/ApplicationOAuth.py#L94

https://github.com/PyGithub/PyGithub/pull/2528/files#diff-e1bc4144bb47d4a87072d8149e91f8c66463d81346d958b6f5d3d7dcd9f33375

This is backward compatible to existing auth arguments.


github/Auth.py
```diff
@@ -0,0 +1,328 @@
...
+        private_key: str,
+        jwt_expiry: int = Consts.DEFAULT_JWT_EXPIRY,
+        jwt_issued_at: int = Consts.DEFAULT_JWT_ISSUED_AT,
+        jwt_algorithm: str = Consts.DEFAULT_JWT_ALGORITHM,
+    ):
+        assert isinstance(app_id, (int, str)), app_id
+        assert isinstance(private_key, str)
+
+        self._app_id = app_id
+        self._private_key = private_key
```
JLLeitschuh, line 143: Probably good to fail-fast here and verify that these aren't empty or `None` at this point

JLLeitschuh (changes requested):

github/Auth.py
EnricoMi: I agree, but asserting for not None is not common in this project, and I am too lazy to start this.

Given the arguments are not optional, `None` values would be surprising from a user point of view. Also, typing suggests `None` is not allowed.

tests/Authentication.py
```diff
@@ -37,39 +37,60 @@ def testNoAuthentication(self):
...
         self.assertEqual(g.get_user("jacquev6").name, "Vincent Jacques")
 
     def testJWTAuthentication(self):
-        g = github.Github(jwt=self.jwt)
-        self.assertEqual(g.get_user("jacquev6").name, "Vincent Jacques")
-
-    def testUserAgent(self):
-        g = github.Github(user_agent="PyGithubTester")
+        g = github.Github(jwt=self.jwt.token)
         self.assertEqual(g.get_user("jacquev6").name, "Vincent Jacques")
```
JLLeitschuh, line 49: Any assertions you want to put in here around verifying that the warning is logged?

github/MainClass.py
```diff
@@ -118,12 +126,40 @@ def __init__(
...
+                category=DeprecationWarning,
+            )
+            auth = Auth.AppAuthToken(jwt)
+        elif app_auth is not None:
+            warnings.warn(
+                "Argument app_auth is deprecated, please use "
+                "auth=github.Auth.AppInstallationAuth(...) instead",
+                category=DeprecationWarning,
+            )
+            auth = app_auth
```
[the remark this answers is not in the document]
EnricoMi: Well, I think that should be removed in the next major release (v2) as this is quite fundamental / widespread use.

We have seen lots of small breaking changes in minor releases, recently, and given there is a lot of refactoring still need, I would refrain from following semantic versioning.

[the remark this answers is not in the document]
EnricoMi: > What if no auth is provided, is that anonymous auth at that point? Should an `Auth.Auth` type for anonymous access be added too, just so that there isn't a 'None' case?

Good point. I think `auth=None` is expressive enough. I don't see benefit for an `NoAuth` or `Anonymous` auth class.

tests/Authentication.py
EnricoMi: done

github/Auth.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+
+
+class Token(Auth):
+    """
+    This class is used to authenticate Requester with a single constant token.
+    """
+
+    def __init__(self, token: str):
+        self._token = token
```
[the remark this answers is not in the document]
EnricoMi: done

github/Auth.py
EnricoMi: done

github/Auth.py
```diff
@@ -0,0 +1,342 @@
...
+from github.Requester import Requester, WithRequester
+
+# For App authentication, time remaining before token expiration to request a new one
+ACCESS_TOKEN_REFRESH_THRESHOLD_SECONDS = 20
+TOKEN_REFRESH_THRESHOLD_TIMEDELTA = timedelta(
+    seconds=ACCESS_TOKEN_REFRESH_THRESHOLD_SECONDS
+)
+
+
+class Auth:
```
JLLeitschuh, line 43: I think to be correct, this needs to also implement `ABC`
```suggestion
class Auth(abc.ABC):
```

github/Auth.py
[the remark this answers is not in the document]
EnricoMi: Btw., some auth classes like the AppInstallationAuth has to refresh its token, so they are not all immutable.

github/Auth.py
EnricoMi: done

EnricoMi:
Thanks!

Issue #2521, opened 2023-05-08
Updating Settings of Repositories-Limited options


AleksanderMino:
Hello, I am writing a script to update the settings of repositories with common required settings. I face some limitations when I use the available methods for the settings.
For example, I am using the edit method from the Repository class which calls the endpoint [PATCH /repos/{owner}/{repo} but only limited settings can be updated by using this method while through the endpoint all of the settings can be changed.
I wanted to ask if the rest of the options will be added to the code.
![edit_method](https://user-images.githubusercontent.com/115500459/236939992-7e0452e2-cfdf-46b6-b5a8-21c894f82949.png)
@nkakouros

amnuts:
Have run into the same issue for the same reason today...

EnricoMi:
This should have been fixed by #2529. Please reopen if not.

PR #2529, opened 2023-05-12
Fixes issue #2521

amnuts:
The Repository interface was behind the function itself and would not allow a couple of the options to be used.  This PR brings them inline.


JLLeitschuh (approved):

Issue #2523, opened 2023-05-08
proposal: migrate project to python-poetry


trim21:
https://python-poetry.org/

this help to manage deps

PR #2530, opened 2023-05-13
build: use python poetry

trim21:
https://github.com/PyGithub/PyGithub/issues/2523


Borda:
I am not much fun of `poetry`, but I like the config unification in `pyproject.toml`

trim21:
setuptools support pep518, it's still possible to use pyproject.toml without using poetry, but imt not very interested in submiting a pr for this

PR #2531, opened 2023-05-15
Add github.Rate.used field

yurinnick:
Add missing `used` field to `github.Rate` class, as it should be according to Github documetation: [Rate limit](https://docs.github.com/en/rest/rate-limit?apiVersion=2022-11-28)


yurinnick:
I decided to keep `Bearer jwt_removed` and configure tests to work with it, instead of replacing auth header with `Basic login_and_password_removed`. Not sure it this is what you folks want to see. Let me know if I should use basic auth in tests to keep it consistent.

EnricoMi:
> I decided to keep `Bearer jwt_removed` and configure tests to work with it, instead of replacing auth header with `Basic login_and_password_removed`. Not sure it this is what you folks want to see. Let me know if I should use basic auth in tests to keep it consistent.

I think that was OK for recording the replay data, but since the test itself does not require jwt auth, it should be reset to standard. Simply remove the `jwtAuthMode = True` and revert line 6 of the modified replay data files. Done.

Happy to approve.

github/Rate.py
```diff
@@ -63,10 +63,18 @@ def reset(self):
         """
         return self._reset.value
 
+    @property
+    def used(self):
```
trim21, line 67:
```suggestion
    def used(self) -> int:
```

yurinnick: I believe that typing is handled by `Rate.pyi`

trim21: we are going to merge `Rate.pyi` back to `Rate.py`, so it would be helpful to also add inline type.

yurinnick:
- Added contributor header
- Added typing annotation to `Rate.pyi`
- Revert test auth headers back to `BasicAuth`

EnricoMi (approved):
LGTM!

PR #2542, opened 2023-06-05
Add `rerequest` endpoint to `CheckRun` model. 

suzyng83209:
This exists on the `CheckSuite` model but not on the `CheckRun` model even though it is similarly supported by Github API.
https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#rerequest-a-check-run


github/CheckRun.py
```diff
@@ -169,6 +169,17 @@ def url(self):
         self._completeIfNotSet(self._url)
         return self._url.value
 
+    def rerequest(self):
```
trim21, line 172:
```suggestion
    def rerequest(self) -> bool:
```

github/CheckRun.py
```diff
@@ -169,6 +169,17 @@ def url(self):
...
+        """
+        :calls: `POST /repos/{owner}/{repo}/check-runs/{check_run_id}/rerequest https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#rerequest-a-check-run`_
```
EnricoMi, line 174:
```suggestion
        :calls: `POST /repos/{owner}/{repo}/check-runs/{check_run_id}/rerequest <https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#rerequest-a-check-run>`_
```

EnricoMi (approved):
LGTM! Comments are minor (`pyi` file contains typing, other links in the file have the same wrong syntax). A test would be grand, though.

github/CheckRun.py
EnricoMi: Yes, the `pyi` file has been removed in main branch, so this typing has to go into the `py` file.

EnricoMi (changes requested):

PR #2543, opened 2023-06-06
Make MainClass.get_app return completed GithubApp

### LINKED ISSUE #2487: GithubApp.slug is None due to no `_completeIfNotSet` call
I noticed that when I do the following code, it will print out `None`. I saw that the slug attribute doesn't use `_completeIfNotSet` before returning the value.:

```python
g = Github(jwt="jwt_token_here")
app = g.get_app()
print(f"slug: {app.slug}")
```

https://github.com/PyGithub/PyGithub/blob/master/github/GithubApp.py#L107-L112

EnricoMi:
There is no need not to return a completed GithubApp at this point. A changing `url` smells. Fixes #2487.


EnricoMi:
I have rethought this: https://github.com/EnricoMi/PyGithub/commit/76158670a0f0ca6cbc68376e23d27fa81a0abda7

When a slug is given, an app is fully identified. With this incomplete information, a GithubApp can be returned lazily.

Without a slug, the current app is fetched through `/app` and a complete object is returned. An incomplete GithubApp without the slug (identifier) is of no use here.

It was the opposite earlier. I think, it makes more sense now. Especially, the `app.url` does not change any more between before and after completion.

PR #2544, opened 2023-06-07
Add sort order and direction for getting comments

EnricoMi:
Some methods getting comments miss sort query parameters.


JLLeitschuh (approved):
LGTM!

PR #2546, opened 2023-06-08
Implement `AppUserAuth` for Github App user tokens

EnricoMi:
Allows to refresh Github App user token. Integrates `ApplicationOAuth` into `github.Auth`.


EnricoMi:
CC @rigaspapas

github/AccessToken.pyi
JLLeitschuh, line 1: Given this file is so small, you could likely delete it and just put the types on the implementation

JLLeitschuh (approved):
Not sure if I can review this on substance, but looks good to me!

github/AccessToken.pyi
EnricoMi: Good point! I started this, but this unveiled lots of typing warnings, discussed and to be fixed here: https://github.com/PyGithub/PyGithub/pull/2463#discussion_r1227714353

PR #2547, opened 2023-06-08
Fix auth issues with `Installation.get_repos`

### LINKED ISSUE #2504: Installation.get_repos() appears to be broken - always throws BadCredentialsException

EnricoMi:
Fixes #2504.


github/Installation.py
```diff
@@ -46,6 +47,16 @@ class Installation(github.GithubObject.NonCompletableGithubObject):
     This class represents Installations. The reference can be found here https://docs.github.com/en/rest/reference/apps#installations
     """
 
+    def __init__(self, requester, headers, attributes, completed):
+        super().__init__(requester, headers, attributes, completed)
+
+        auth = self._requester.auth if self._requester is not None else None
+        # Usually, an Installation is created from a Requester with App authentication
+        if isinstance(auth, AppAuth):
```
JLLeitschuh, line 55: What if it's not? Or there is no Auth? Can you have anonymous access to an installation?

JLLeitschuh (approved):
Going to be completely honest, I don't know anything about installations, so I'm going to have to trust you on this one

github/Installation.py
EnricoMi: The `auth` should neither be `None` nor not of type `AppAuth`, but just in case it is, I don't want to change existing behaviour of user code, which should see an error response by the API.

Maybe there is a valid code path that creates an Installation without auth, because it never uses the requester.

PR #2549, opened 2023-06-09
Add support for get_app() with App authentication

chantra:
An authenticated app can call `/app` to get details about itself.

Authentication must be done using jwt for authentication or it will fail with an error like https://gist.github.com/chantra/209819389b2dec6cc6e03a13d301f5e5


chantra:
@EnricoMi , I am not super satisfied of the current solution. Let me know how you would see this better integrated.

It seems to me that wether through GithubIntegration, or Github main object, we need to detect that the authentication will be done using JWT. In MainClass.py, I use the token_type as a proxy for this.
I don't think GithubIntegration is exclusively JWT auth based, so the same should probably apply.
In any cases, the current solution is wonky. Happy to iterate on it.

github/MainClass.py
```diff
@@ -808,6 +808,7 @@ def get_app(self, slug=github.GithubObject.NotSet):
         if slug is github.GithubObject.NotSet:
```
EnricoMi, line 808: I'd make `slug` mandatory and remove the `=github.GithubObject.NotSet` and `if slug is github.GithubObject.NotSet:`. This method should only serve the current `else` branch.

tests/GithubApp.py
```diff
@@ -98,8 +98,14 @@ def testGetPublicApp(self):
         self.assertEqual(app.updated_at, datetime(2019, 12, 10, 19, 4, 12))
         self.assertEqual(app.url, "/apps/github-actions")
 
+
+class GithubAppAuth(Framework.TestCase):
+    def setUp(self):
+        self.jwtAuthMode = True
+        super().setUp()
+
     def testGetAuthenticatedApp(self):
```
EnricoMi, line 107: test an be removed

EnricoMi (changes requested):

github/MainClass.py
```diff
@@ -808,6 +808,7 @@ def get_app(self, slug=github.GithubObject.NotSet):
...
             # with no slug given, calling /app returns the authenticated app,
             # including the actual /apps/{slug}
+            assert self.__requester._Requester__auth.token_type == 'Bearer', "Only AppAuth authentication support get_app without a slug."
             headers, data = self.__requester.requestJsonAndCheck("GET", "/app")
             return GithubApp.GithubApp(self.__requester, headers, data, completed=True)
```
EnricoMi, line 813: Let's not break user code that use this method without providing `slug`(it should have worked with `Github(jwt=GithubIntegration(...).create_jwt()`, which is deprecated).

Let's keep slug optional, but raise a deprecated warning:
```suggestion
            warnings.warn(
                "Argument slug is mandatory, calling this method without the slug argument is deprecated, please use "
                "github.GithubIntegration(auth=github.Auth.AppAuth(...)).get_app() instead",
                category=DeprecationWarning,
            )
            return GithubIntegration(auth=self.auth).get_app()
```

tests/GithubApp.py
EnricoMi: let's keep this test but also assert the warning, see this test for an example:
https://github.com/PyGithub/PyGithub/blob/db616376a51789c96afe1e6039f7aecc9b08f56e/tests/Authentication.py#L54-L61

github/MainClass.py
EnricoMi: ignore this, see below...

github/GithubIntegration.py
```diff
@@ -213,3 +214,16 @@ def get_app_installation(self, installation_id):
         :rtype: :class:`github.Installation.Installation`
         """
         return self._get_installed_app(url=f"/app/installations/{installation_id}")
+
+    def get_app(self):
+        """
+        :calls: `GET /app <https://docs.github.com/en/rest/reference/apps#get-the-authenticated-app>`_
+        :rtype: :class:`github.GithubApp.GithubApp`
+        """
+
```
EnricoMi, line 223: You could assert in this line

https://github.com/PyGithub/PyGithub/blob/db616376a51789c96afe1e6039f7aecc9b08f56e/github/GithubIntegration.py#L75

that

    assert isinstance(auth, AppAuth), f"GithubIntegration requires github.Auth.AppAuth authentication, not ${type(auth)}"

EnricoMi (changes requested):

chantra:
Thanks @EnricoMi 

Applied your recommendations in the second commit. If that works out, feel free to squash it into the first one.

chantra:
running manually, this is failing due to utcnow deprecation:
```
tests/GithubApp.py .F                                                                                                                                                                                                                                                                                            [100%]

======================================================================================================================================================= FAILURES =======================================================================================================================================================
________________________________________________________________________________________________________________________________________ GithubAppAuth.testGetAuthenticatedApp _________________________________________________________________________________________________________________________________________

self = <tests.GithubApp.GithubAppAuth testMethod=testGetAuthenticatedApp>

    def testGetAuthenticatedApp(self):
        with self.assertWarns(DeprecationWarning) as warning:
            app = self.g.get_app()

>           self.assertWarning(
                warning,
                "Argument slug is mandatory, calling this method without the slug argument is deprecated, please use github.GithubIntegration(auth=github.Auth.AppAuth(...)).get_app() instead",
            )

tests/GithubApp.py:111:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/Framework.py:334: in assertWarning
    self.assertWarnings(warning, expected)
tests/Framework.py:337: in assertWarnings
    self.assertEqual(len(warning.warnings), len(expecteds))
E   AssertionError: 2 != 1
=================================================================================================================================================== warnings summary ===================================================================================================================================================
github/Repository.py:3476
  /Users/chantra/devel/PyGithub/github/Repository.py:3476: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    def mark_notifications_as_read(self, last_read_at=datetime.datetime.utcnow()):

github/AuthenticatedUser.py:1232
  /Users/chantra/devel/PyGithub/github/AuthenticatedUser.py:1232: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    def mark_notifications_as_read(self, last_read_at=datetime.datetime.utcnow()):

tests/GithubApp.py::GithubApp::testGetPublicApp
  /Users/chantra/devel/PyGithub/.py312/lib/python3.12/site-packages/httpretty/core.py:1077: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    now = datetime.utcnow()

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================================================================================================================== short test summary info ================================================================================================================================================
FAILED tests/GithubApp.py::GithubAppAuth::testGetAuthenticatedApp - AssertionError: 2 != 1
```

While cI can fix the deprecation for pygithub, httpretty will still generate a warning.

chantra:
I am introducing `assertWarningIn` to be able to check for a specific warning in the list of warnings.

EnricoMi:
> running manually, this is failing due to utcnow deprecation:

Then use the recommended `datetime.now(datetime.UTC)`.

tests/Framework.py
```diff
@@ -328,6 +330,27 @@ def tearDown(self):
         self.__closeReplayFileIfNeeded()
         github.Requester.Requester.resetConnectionClasses()
 
+    def assertWarning(self, warning, expected):
```
EnricoMi, line 333: Good point moving this into Framework!

tests/GithubApp.py
```diff
@@ -98,9 +98,21 @@ def testGetPublicApp(self):
         self.assertEqual(app.updated_at, datetime(2019, 12, 10, 19, 4, 12))
         self.assertEqual(app.url, "/apps/github-actions")
 
+
+class GithubAppAuth(Framework.TestCase):
+    def setUp(self):
+        self.appAuthMode = True
```
EnricoMi, line 104: This sets auth mode for all subsequent tests, this should only be used by `tests.conftest.pytest_configure` via command line arguments like `--record`.

Please use the same way of authentication as used in `tests/GithubIntegration.py`:
```python
        auth = github.Auth.AppAuth(APP_ID, PRIVATE_KEY)
        g = github.Github(auth=auth)
        app = g.get_app()
```

You can import `APP_ID` and `PRIVATE_KEY` from `tests/GithubIntegration.py`.

So this `GithubAppAuth` class and `def setUp` can be removed.

tests/Framework.py
```diff
@@ -410,6 +437,10 @@ def activateJWTAuthMode():  # pragma no cover (Function useful only when recordi
     BasicTestCase.jwtAuthMode = True
 
 
+def activateAppAuthMode():  # pragma no cover (Function useful only when recording new tests, not used during automated tests)
```
EnricoMi, line 440: this needs to be added to `tests.conftest.pytest_configure` to be useful

github/GithubIntegration.py
```diff
@@ -71,7 +72,10 @@ def __init__(
                 jwt_algorithm=jwt_algorithm,
             )
 
-        assert auth is not None
+        assert isinstance(
+            auth, AppAuth
+        ), f"GithubIntegration requires github.Auth.AppAuth authentication, not {type(auth)}"
+
```
EnricoMi, line 78: can we have a test for this as well?

EnricoMi (changes requested):
Almost there!

github/MainClass.py
```diff
@@ -808,8 +808,12 @@ def get_app(self, slug=github.GithubObject.NotSet):
...
             # with no slug given, calling /app returns the authenticated app,
             # including the actual /apps/{slug}
-            headers, data = self.__requester.requestJsonAndCheck("GET", "/app")
-            return GithubApp.GithubApp(self.__requester, headers, data, completed=True)
+            warnings.warn(
+                "Argument slug is mandatory, calling this method without the slug argument is deprecated, please use "
+                "github.GithubIntegration(auth=github.Auth.AppAuth(...)).get_app() instead",
+                category=DeprecationWarning,
+            )
+            return GithubIntegration(auth=self.__requester._Requester__auth).get_app()
```
EnricoMi, line 816: Please rebase and access `Requester.auth`:
```suggestion
            return GithubIntegration(auth=self.__requester.auth).get_app()
```

EnricoMi (changes requested):

chantra:
> > running manually, this is failing due to utcnow deprecation:
> 
> Then use the recommended `datetime.now(datetime.UTC)`.

Per my commit message in https://github.com/PyGithub/PyGithub/pull/2549/commits/c3a2d35506db477b46df97878fcba765b95821fa , datetime.UTC was introduced in 3.11, datetime.timezone.utc works across the board of python 3.X tested in this project.

tests/GithubApp.py
chantra: ok, I removed this code from the PR then as it is not required. This feature would then be orthogonal to this PR.

tests/GithubApp.py
```diff
@@ -99,8 +102,17 @@ def testGetPublicApp(self):
...
+        auth = github.Auth.AppAuth(APP_ID, PRIVATE_KEY)
+        g = github.Github(auth=auth)
+
+        with self.assertWarns(DeprecationWarning) as warning:
+            app = g.get_app()
+
+            self.assertWarningIn(
+                warning,
+                "Argument slug is mandatory, calling this method without the slug argument is deprecated, please use github.GithubIntegration(auth=github.Auth.AppAuth(...)).get_app() instead",
+            )
```
EnricoMi, line 114:
```suggestion
        self.assertWarningIn(
            warning,
            "Argument slug is mandatory, calling this method without the slug argument is deprecated, "
            "please use github.GithubIntegration(auth=github.Auth.AppAuth(...)).get_app() instead",
        )
```

tests/GithubIntegration.py
```diff
@@ -67,6 +56,15 @@ def testDeprecatedAppAuth(self):
...
         )
 
+    def testRequiredAppAuth(self):
+        # GithubIntegration requires AppAuth authentication.
+        for auth in [self.oauth_token, self.jwt, self.login]:
+            with self.assertRaisesRegex(
+                AssertionError,
+                "GithubIntegration requires github.Auth.AppAuth authentication, not .*",
+            ):
+                github.GithubIntegration(auth=self.oauth_token)
```
EnricoMi, line 66:
```suggestion
            with self.assertRaises(
                AssertionError,
                f"GithubIntegration requires github.Auth.AppAuth authentication, not {type(auth)}",
            ):
                github.GithubIntegration(auth=auth)
```

tests/Framework.py
```diff
@@ -298,13 +299,6 @@ def setUp(self):
                 if GithubCredentials.jwt
                 else None
             )
-            self.app_auth = (
-                github.Auth.AppAuth(
-                    GithubCredentials.app_id, GithubCredentials.app_private_key
-                )
-                if GithubCredentials.app_id and GithubCredentials.app_private_key
-                else None
-            )
```
EnricoMi, line 307: No need to remove this existing code.

tests/GithubApp.py
```diff
@@ -99,8 +102,17 @@ def testGetPublicApp(self):
...
     def testGetAuthenticatedApp(self):
-        # For this to work correctly in record mode, this test must be run with --auth_with_jwt
-        app = self.g.get_app()
+        auth = github.Auth.AppAuth(APP_ID, PRIVATE_KEY)
+        g = github.Github(auth=auth)
+
+        with self.assertWarns(DeprecationWarning) as warning:
+            app = g.get_app()
+
+            self.assertWarningIn(
```
EnricoMi, line 111:
```suggestion
            self.assertWarning(
```

EnricoMi (changes requested):

tests/Framework.py
chantra: I had added this, do you want me to re-add it even if this is not used here?

tests/GithubApp.py
chantra: The problem is that httpretty/core generates another warning, see https://github.com/PyGithub/PyGithub/pull/2549#issuecomment-1587869573

tests/GithubIntegration.py
chantra: good catch on the auth vs self.oauth_token :)

tests/Framework.py
EnricoMi: This diff tells me, this was there before and your pull-request removes it.

tests/GithubApp.py
EnricoMi: Which Python version are you using?

tests/GithubApp.py
```diff
@@ -99,8 +102,18 @@ def testGetPublicApp(self):
...
```
EnricoMi, line 111: To be honest, I don't like `assertWarningIn`, because it may hide unexpected warnings that we want to fail on.

Having `httpretty` warn here is annoying, but we can disable that and everything works as expected:

```suggestion
        with self.assertWarns(DeprecationWarning) as warning:
            # we ignore warnings from httpretty dependency
            import warnings
            warnings.filterwarnings("ignore", module="httpretty")

            app = g.get_app()

            self.assertWarning(
```

tests/Framework.py
```diff
@@ -328,6 +322,27 @@ def tearDown(self):
...
+                return
+
+        assert False, f"Warning *{expected}* not found in *{warning.warnings}*"
+
+    def assertWarnings(self, warning, *expecteds):
+        self.assertEqual(len(warning.warnings), len(expecteds))
+        for message, expected in zip(warning.warnings, expecteds):
+            self.assertIsInstance(message, warnings.WarningMessage)
+            self.assertIsInstance(message.message, DeprecationWarning)
+            self.assertEqual(message.message.args, (expected,))
```
EnricoMi, line 344: An error like

     E   AssertionError: 2 != 1

is not really helpful, lets improve this:
```suggestion
    def assertWarnings(self, warning, *expecteds):
        actual = [(type(message), type(message.message), message.message.args) for message in warning.warnings]
        expected = [(warnings.WarningMessage, DeprecationWarning, (expected,)) for expected in expecteds]
        self.assertSequenceEqual(actual, expected)
```

EnricoMi (changes requested):

tests/GithubApp.py
chantra: oh great, I did not know we could filter the warnings. Will change. Thanks

chantra:
rebased, re-ported utcnow change and got rid of assertWarningIn.

tests/Framework.py
EnricoMi: This is the last bit left. I'd say there is no need to remove this code by this pull request.

chantra: hum, I was sure I had added that when I was creating a new test class.... I may have it mixed up with something else. Will re-introduce.

chantra:
ok, if I did not mess up, the CI should be green in a few and all feedback addressed :D

EnricoMi (approved):
LGTM

EnricoMi:
Thank you for your patience!

chantra:
Thanks to you too!

PR #2554, opened 2023-06-16
Merge `WorkflowRun` type stub back to source

trim21:


EnricoMi (changes requested):
I like defining the attributes with their getter functions, but we should stick to one layout while migrating type stubs back: #2553

trim21:
I found that instance attribute is not necessary

github/WorkflowRun.py
```diff
@@ -216,22 +209,16 @@ def get_artifacts(self) -> "PaginatedList[github.Artifact.Artifact]":
...
     @property
     def rerun_url(self) -> str:
         self._completeIfNotSet(self._rerun_url)
         return self._rerun_url.value
 
-    _workflow_url: Attribute[str]
-
     @property
     def workflow_url(self) -> str:
         """
```
EnricoMi, line 224: You removed these docs from other methods, should remove here then as well.

EnricoMi (approved):
LGTM, minor change requested.

github/WorkflowRun.py
```diff
@@ -142,79 +156,52 @@ def workflow_id(self):
         return self._workflow_id.value
```
EnricoMi, line 156: Sorry, but this property misses type annotation and lacks doc string removal.

EnricoMi (changes requested):

EnricoMi:
Please request review when done, so this flags up as ready for me. Hard to keep track which PR is ready and which are in progress.

trim21:
> Please request review when done, so this flags up as ready for me. Hard to keep track which PR is ready and which are in progress.

Sorry, I thought this PR is done when I trigger review request. Just didn't notice I missed one property.

github/WorkflowRun.py
```diff
@@ -21,200 +21,184 @@
...
-    def pull_requests(self):
-        """
-        :type: list of :class:`github.PullRequest.PullRequest`
-        """
+    def pull_requests(self) -> list[PullRequest]:
         self._completeIfNotSet(self._pull_requests)
         return self._pull_requests.value
 
     @property
     def created_at(self):
```
EnricoMi, line 171: missing typing

github/WorkflowRun.py
```diff
@@ -21,200 +21,184 @@
...
-        """
-        :type: datetime.datetime
-        """
         self._completeIfNotSet(self._created_at)
         return self._created_at.value
 
     @property
     def updated_at(self):
```
EnricoMi, line 176: missing typing

EnricoMi (changes requested):
Sorry, two more missing, please double-check as well.

trim21:
it's all typed this time!

EnricoMi (approved):
Looks like it is.

PR #2555, opened 2023-06-16
Merge PaginatedList.pyi back to source

trim21:


github/PaginatedList.py
```diff
@@ -34,48 +34,61 @@
...
-    def _isBiggerThan(self, index):
+    def _isBiggerThan(self, index: int) -> bool:
         return len(self.__elements) > index or self._couldGrow()
 
     def __fetchToIndex(self, index):
         while len(self.__elements) <= index and self._couldGrow():
             self._grow()
 
-    def _grow(self):
+    def _grow(self) -> Any:
```
EnricoMi, line 79: Shouldn' this be
```suggestion
    def _grow(self) -> Iterable[T]:
```

github/PaginatedList.py
```diff
@@ -34,48 +34,61 @@
...
         newElements = self._fetchNextPage()
         self.__elements += newElements
         return newElements
 
     class _Slice:
-        def __init__(self, theList, theSlice):
+        def __init__(self, theList: "PaginatedListBase[T]", theSlice: slice) -> None:
```
EnricoMi, line 85:
```suggestion
        def __init__(self, theList: "PaginatedListBase[T]", theSlice: slice):
```

github/PaginatedList.py
```diff
@@ -237,7 +249,7 @@ def __parseLinkHeader(self, headers):
                 links[rel] = url
         return links
 
-    def get_page(self, page):
+    def get_page(self, page: int) -> List[Any]:
```
EnricoMi, line 252:
```suggestion
    def get_page(self, page: int) -> List[T]:
```

github/PaginatedList.py
```diff
@@ -34,48 +34,61 @@
...
-    def _isBiggerThan(self, index):
+    def _isBiggerThan(self, index: int) -> bool:
         return len(self.__elements) > index or self._couldGrow()
 
     def __fetchToIndex(self, index):
         while len(self.__elements) <= index and self._couldGrow():
             self._grow()
 
-    def _grow(self):
+    def _grow(self) -> Iterable[T]:
```
trim21, line 79:
```suggestion
    def _grow(self) -> Iterator[T]:
```

github/Requester.py
```diff
@@ -426,7 +426,7 @@ def requestJsonAndCheck(
         parameters: Optional[Dict[str, Any]] = None,
         headers: Optional[Dict[str, str]] = None,
         input: Optional[Any] = None,
-    ) -> Tuple[Dict[str, Any], Optional[Dict[str, Any]]]:
```
EnricoMi, line 429: Why removing the `Dict` here?

github/Requester.py
```diff
@@ -469,7 +469,7 @@ def __check(
         status: int,
         responseHeaders: Dict[str, Any],
         output: str,
-    ) -> Tuple[Dict[str, Any], Dict[str, Any]]:
```
EnricoMi, line 472: Why removing the `Dict` here?

EnricoMi (changes requested):

github/Requester.py
trim21: It parse request response body, not necessarily dict

github/Requester.py
trim21: Same above

EnricoMi: `requestJsonAndCheck` returns what `__check` returns, which is what `__structuredFromJson` returns, which calls into `json.loads(data)`, so should be `Optional[Dict[str, Any]]`.

trim21: Doesn't `json.loads(data)` actually return `Any`?

trim21: https://github.com/PyGithub/PyGithub/blob/b6258f4be9bee07beb0935a749289b152b20f7f0/github/PaginatedList.py#L201-L204

trim21: I'm 100% sure github API not only returns objects

https://api.github.com/repos/minio/minio-js/releases

trim21: > Right, `Any` still seems a bit broad to me.

But we can't typed it as `list|dict|None`, if you are using a union type, mypy expect all element of this union match expected types.

For example, `list|dict|None` can't be the argument of  `def f(o: list): ...`

EnricoMi: Awful legacy, well then be it `Any`.

trim21: Not only legacy problem, call dict method on `None|dict` also raise mypy type checking error.

EnricoMi (approved):
LGTM!

EnricoMi:
@trim21 ready to merge?

trim21:
yes

trim21:
I don't have write permission 🙂

EnricoMi:
Thanks!

PR #2556, opened 2023-06-19
Rework GithubIntegration

EnricoMi:
Both `github.Github` and `github.GithubIntegration` are main entry points of this library. They create a `Requester` instance that handles all requests to the Github REST API. Requester has numerous arguments like `base_url` or `timeout`.

Firstly, `Github` and `GithubIntegration` should support the same (full) set of `Requester` arguments.

Secondly, creating a `Github` instance for a Github App Installation coming from `GithubIntegration` should use the same `Requester` arguments (except for `auth`). It should suffice to define `base_url` and other parameters once.

Fixes issues like #2455.


github/GithubIntegration.py
```diff
@@ -26,6 +29,12 @@ def __init__(
         integration_id=None,
         private_key=None,
         base_url=Consts.DEFAULT_BASE_URL,
+        timeout=Consts.DEFAULT_TIMEOUT,
+        user_agent=Consts.DEFAULT_USER_AGENT,
+        per_page=Consts.DEFAULT_PER_PAGE,
+        verify=True,
+        retry=None,
+        pool_size=None,
```
JLLeitschuh, line 37: This is an API breaking change if end users are using positional arguments

github/GithubIntegration.pyi
JLLeitschuh, line 1: Since this is a, relatively, small class, could we delete this `pyi` file and move all the types over to the `py` file since you're already touching this file?

github/GithubIntegration.py
```diff
@@ -35,8 +44,15 @@ def __init__(
         :param integration_id: int deprecated, use auth=github.Auth.AppAuth(...) instead
         :param private_key: string deprecated, use auth=github.Auth.AppAuth(...) instead
         :param base_url: string
+        :param timeout: integer
+        :param user_agent: string
+        :param per_page: int
+        :param verify: boolean or string
```
JLLeitschuh, line 50: Should this document what values of `verify` there are? I don't know why this would ever be a `str`, or what values it would have

JLLeitschuh: Ah, I tracked it down. this eventually flows to `requests.session` which takes the `verify` argument and uses it. I understand.

tests/Requester.py
```diff
@@ -39,6 +39,90 @@ def tearDown(self):
...
+            verify=False,
+            retry=3,
+            pool_size=5,
+        )
+        kwargs = requester.kwargs
+
+        # assert kwargs consists of ALL constructor arguments
+        self.assertEqual(
+            kwargs.keys(), github.Requester.Requester.__init__.__annotations__.keys()
+        )
```
JLLeitschuh, line 63: Good test!

JLLeitschuh (approved):
LGTM. One question about potentially API breaking changes though.

github/GithubIntegration.py
EnricoMi: That is right. The rational here is the `jwt_*` arguments have been added quite recently and should be rarely used. So this breaks for little user code only, and is easy to fix. This will be flagged in the change log.

I favour similarity to `github.Github` signature here over not breaking this API for everyone.

github/GithubIntegration.pyi
EnricoMi: I'd prefer to deal with that separately.

PR #2563, opened 2023-06-21
Release 1.59.0

EnricoMi:


github/GithubIntegration.py
```diff
@@ -22,13 +22,15 @@ class GithubIntegration:
...
+    # v2: move * before first argument so all arguments must be named,
     #     allows to reorder / add new arguments / remove deprecated arguments without breaking user code
+    #     added here to force named parameters because new parameters have been added
     def __init__(
         self,
         integration_id=None,
         private_key=None,
         base_url=Consts.DEFAULT_BASE_URL,
+        *,
         timeout=Consts.DEFAULT_TIMEOUT,
```
EnricoMi, line 34: Without this, existing code like `GithubIntegration(id, key, url, expiry)` would silently use `expiry` as `timeout`. This breaks code that uses `jwt_*` arguments as positional arguments.

JLLeitschuh: I actually don't know what this syntax means. Could you point to a doc that explains it?

JLLeitschuh (approved):
In general looks fine, but you're using python syntax that I'm unfamiliar with

PR #2565, opened 2023-06-22
Make datetime objects timezone-aware

### LINKED ISSUE #512: Datetime returned is naive, should be UTC

EnricoMi:
Continues #1831.

> Timestamps returned by the GitHub API include timezone information, yet PyGithub throws this information away. This can lead to wrong results in client code, as naïve timestamps are generally assumed to use the system time zone, yet the majority of timestamps returned by the API are in UTC. This PR fixes the library to retain timezone information when parsing timestamps and always produce timezone-aware datetime objects.

Additionally, this imports `date, datetime, timezone` directly and reduces code like `datetime(…, tzinfo=datetime.timezone.utc)` to `datetime(…, tzinfo=timezone.utc))`.

This also replaces some `datetime.timezone(datetime.timedelta(0))` with `timezone.utc`.

Fixes #512.


trim21:
this will cause breaking change

EnricoMi:
Yes, this breaks user code (where user code compares timestamps). This will be released with a breaking warning and as a pre-release, so users can test. I would refrain from bumping version to v2 just for this one feature.

JLLeitschuh:
> I would refrain from bumping version to v2 just for this one feature

Out of curiosity, why? It's just a number?  I don't understand why maintainers are generally afraid of reving the major version number

tests/BadAttributes.py
```diff
@@ -52,11 +57,11 @@ def testBadAttributeTransformation(self):
         self.assertEqual(raisedexp.exception.actual_value, "foobar")
         self.assertEqual(raisedexp.exception.expected_type, str)
         self.assertEqual(
-            raisedexp.exception.transformation_exception.__class__, ValueError
+            raisedexp.exception.transformation_exception.__class__, ParserError
         )
```
JLLeitschuh, line 61: Doesn't the testing api have `assertIsInstance`?

JLLeitschuh (approved):
LGTM! I wouldn't be afraid of reving the major version number though

EnricoMi:
@trim21 please join the Slack channel and ping me there: https://join.slack.com/t/pygithub-project/shared_invite/zt-duj89xtx-uKFZtgAg209o6Vweqm8xeQ

trim21:
> @trim21 please join the Slack channel and ping me there: https://join.slack.com/t/pygithub-project/shared_invite/zt-duj89xtx-uKFZtgAg209o6Vweqm8xeQ

what？

tests/BadAttributes.py
EnricoMi: It does, this looks like poor legacy code. I didn't spot this, I just adjusted the expectation :face_with_spiral_eyes:.

PR #2577, opened 2023-06-27
Merge `AuthenticatedUser` type stubs back to source

trim21:

PR #2582, opened 2023-06-29
Add create_repo_from_template to AuthenticatedUser.pyi

fus-kus:


EnricoMi:
The following methods ae also missing from `.pyi`, would you be happy to add those as well?

    create_project
    get_installations

EnricoMi:
Hold on, looks like #2577 is going to fix that all.

trim21:
should close this

Issue #2587, opened 2023-06-30
tests is broken on windows


trim21:
caused by https://github.com/PyGithub/PyGithub/pull/2145

```text
  Traceback (most recent call last):
    File "C:\Users\Trim21\scoop\apps\python310\3.10.11\lib\threading.py", line 1016, in _bootstrap_inner
      self.run()
    File "C:\Users\Trim21\scoop\apps\python310\3.10.11\lib\threading.py", line 953, in run
      self._target(*self._args, **self._kwargs)
    File "C:\Users\Trim21\proj\PyGithub\.venv\lib\site-packages\httpretty\core.py", line 1097, in fill_filekind
      status, headers, self.body = self.callable_body(self.request, self.info.full_url(), headers)
    File "C:\Users\Trim21\proj\PyGithub\tests\Framework.py", line 220, in __request_callback
      self.__readNextRequest(
    File "C:\Users\Trim21\proj\PyGithub\tests\Framework.py", line 193, in __readNextRequest
      assert self.__protocol == readLine(self.__file)
  AssertionError: assert 'https' == 'Authenticate...estEmails.txt'
    - AuthenticatedUser.testEmails.txt
    + https

    warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
```

trim21:
why it's not broken in linux...

trim21:
realy nice, `tests/ReplayData/RequesterThrottled.testShouldDeferRequests.txt` is a symbolic link...

EnricoMi:
*argh*, I hate Windows...

PR #2588, opened 2023-06-30
Remove symlinks

trim21:
symlinks doesn't always work on windows, add pre-commit hook to replace them with real file content.

And please never use symlinks...

fix https://github.com/PyGithub/PyGithub/issues/2587


EnricoMi (approved):
LGTM!

PR #2590, opened 2023-06-30
Test on Windows and macOS, don't fail fast

EnricoMi:
The `Test success` check aggregates all test jobs outcome and allows to require a single check. Then we can modify job names and configure final CI success through workflow file and not repo settings.

This requires removing `CI/test (Python x.y)` as required checks and adding `test success` instead.

This would have discovered #2587 before merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,21 @@ jobs:
         run: tox
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
+
+  test_success:
+    # this aggregates success state of all jobs listed in `needs`
+    # this is the only required check to pass CI
+    name: "Test success"
+    runs-on: ubuntu-latest
+    needs: [test]
+    steps:
+      - name: "Noop"
+        run: true
+        shell: bash
+
   draft:
     runs-on: ubuntu-latest
-    needs: test
+    needs: test_success
     if: github.ref == 'refs/heads/main'
     steps:
       - uses: release-drafter/release-drafter@v5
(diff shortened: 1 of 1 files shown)


EnricoMi:
@trim21 what do you think?

trim21:
lgtm

trim21:
we already have tests on linux, so MacOS may be not necessary...

trim21:
also, `fail-fast` maybe not a bad idea,

JLLeitschuh (approved):
LGTM!

EnricoMi:
> we already have tests on linux, so MacOS may be not necessary...

Sure, not expected to be necessary, but that is what I thought of Windows as well.

Just to be sure there is not some Mac-specific issue, only tests a single py version.

.github/workflows/ci.yml
```diff
@@ -5,12 +5,22 @@ on:
...
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
+        os: ["ubuntu-latest"]
+        os-label: ["Ubuntu"]
+        include:
+          - python-version: "3.8"
+            os: "windows-latest"
+            os-label: "Windows"
+          - python-version: "3.8"
+            os: "macos-latest"
+            os-label: "macOS"
```
Borda, line 23:
```suggestion
          - {python-version: "3.8", os: "windows-latest", os-label: "Windows"}
          - {python-version: "3.8", os: "macos-latest", os-label: "macOS"}
```
could be easier to read

Borda (approved):

Issue #2589, opened 2023-06-30
Mypy errors on some exception in 1.59.0


sodul:
After upgrading to 1.59.0 we got a mypy error on our code catching `UnknownObjectException`:

```
    except github.UnknownObjectException as err:
        message = err.data['message']
```

Now mypy is raising:
```
error: Invalid index type "str" for "str | dict[str, str | list[str] | list[dict[str, str]]]"; expected type "SupportsIndex | slice"  [index]
```

This seems to be caused by the improved type annotation of the `GithubException` base class introduced in #2463:

```
    @property
    def data(
        self,
    ) -> Union[str, Dict[str, Union[str, List[str], List[Dict[str, str]]]]]:
```
Also declared in the `__init__()` signature.

AFAIK `UnknownObjectException` is only raised by `Requester.py` always returning `data` as `output: Dict[str, Any]` so `UnknownObjectException` could be updated to clarify this more accurately so that mypy will not raise typing errors that cannot occur at runtime.

EnricoMi:
@trim21 can you take a look, please?

trim21:
> AFAIK UnknownObjectException is only raised by Requester.py always returning data as output: Dict[str, Any] so UnknownObjectException could be updated to clarify this more accurately so that mypy will not raise typing errors that cannot occur at runtime.

Yes, I think you are right about this.

Borda:
In this moce, may I ask what was the motivation for moving typing to pyi files when the min supported python is 3.7 and it effectively doubles the number of files in the package (and personally, it makes it harder to read as all nested in a single folder)

trim21:
> In this moce, may I ask what was the motivation for moving typing to pyi files when the min supported python is 3.7 and it effectively doubles the number of files in the package (and personally, it makes it harder to read as all nested in a single folder)

because pyi are added before pygithub drop previous python version, they are essential at that time.

Borda:
docs suggest it was added recently in 1.59 release: https://github.com/PyGithub/PyGithub/blob/main/doc/changes.rst#version-1590-june-22-2023

trim21:
> docs suggest it was added recently in 1.59 release: https://github.com/PyGithub/PyGithub/blob/main/doc/changes.rst#version-1590-june-22-2023

https://github.com/PyGithub/PyGithub/tree/v1.57/github

trim21:
pyi are removed, inline types are added in 1.59 release

Borda:
[...]

PR #2591, opened 2023-06-30
Add `message` property to `GithubException`

trim21:
close https://github.com/PyGithub/PyGithub/issues/2589

### DIFF
--- a/github/GithubException.py
+++ b/github/GithubException.py
@@ -28,12 +28,10 @@
 #                                                                              #
 ################################################################################
 import json
-from typing import Any, Dict, Generic, List, Optional, Tuple, Type, TypeVar, Union
-
-T = TypeVar("T", bound=Any)
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 
-class GithubException(Exception, Generic[T]):
+class GithubException(Exception):
     """
     Error handling in PyGithub is done with exceptions. This class is the base of all exceptions raised by PyGithub (but :class:`github.GithubException.BadAttributeException`).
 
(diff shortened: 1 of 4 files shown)


EnricoMi:
Ome `GithubException` has a non `dict` values for `data`, it's used as a message string:
https://github.com/PyGithub/PyGithub/blob/8d3b905786102d4b1890523932a8212cad56e88a/github/GithubObject.py#L380-L382

trim21:
> One `GithubException` has a non `dict` values for `data`, it's used as a message string:
> 
> https://github.com/PyGithub/PyGithub/blob/8d3b905786102d4b1890523932a8212cad56e88a/github/GithubObject.py#L380-L382
> 
> Maybe, this one should be derived from `Exception`, because the `IncompletableObject` exception does not come from Github, but it is the user's fault. While we are there, it could be renamed to `IncompletableObjectException`.

why mypy doesn't find this...

trim21:
I have no better idea about how to fix this...

EnricoMi:
> I have no better idea about how to fix this...

What about an optional exception message (additional argument)?

trim21:
> > I have no better idea about how to fix this...
> 
> What about an optional exception message (additional argument)?

what about current patch?

EnricoMi:
I wouldn't mix the error message into the `data` field, I'd rather have an optional string `msg` field.

All exceptions should have a message field to provide some extra meaning additional to the exception type. The `__repr__` could nicely use this.

EnricoMi (approved):
LGTM!

PR #2593, opened 2023-07-03
CI: Build and check package before release

Borda:
Validate package on master and before making release


trim21:
IMO we should put this in ci and run on all commits and PRs

Borda:
> IMO we should put this in ci job and run on all commits and PRs, not in release job.

OK I can enable it there...

.github/workflows/pypi-release.yml
```diff
@@ -1,8 +1,9 @@
 name: Publish to PyPI
 on:
   push:
-    tags:
-      - "*"
+    branches: [main]  # this will just build and check package
```
Borda, line 4:
```suggestion
    branches: [main]  # this will just build and check package
    pull_request: [main]  # this will just build and check package
```

.github/workflows/pypi-release.yml
```diff
@@ -12,14 +15,17 @@ jobs:
...
-        run: |
-          python setup.py sdist bdist_wheel
-          twine upload dist/*
+        run: pip install -U build twine
+      - name: Build package
+        run: python -m build
+      - name: Check package
+        run: twine check dist/*
+
+      - name: Publish distribution 📦 to PyPI
```
EnricoMi, line 26: All this action does is calling `twine upload dist/*` and checking inputs. I don't think it's worth adding more complexity here.

.github/workflows/pypi-release.yml
```diff
@@ -12,14 +15,17 @@ jobs:
...
+        if: startsWith(github.event.ref, 'refs/tags')
```
EnricoMi, line 27:
```suggestion
        if: startsWith(github.event.ref, 'refs/tags/')
```

EnricoMi:
Reading `Publish to PyPI / publish (pull_request) Successful in 1m` on a branch is scary, can we

1. Rename the workflow to `Build (and publish)`?
2. Split the workflow into jobs `build` and `publish`, where the latter `needs` the former and only runs on tags

.github/workflows/pypi-release.yml
Borda: yes, but then you would need to apply this action on PR and it would fail as you can't push existing version so with respect to reuse the same workflow we need to have it as separate step

EnricoMi:
Publish job tested here (except for upload): https://github.com/PyGithub/PyGithub/actions/runs/5453742972/jobs/9922966236?pr=2593

EnricoMi:
Build check:
grafik

Publish check:
grafik

@Borda what do you think?

EnricoMi (approved):
LGTM!

Borda:
> @Borda what do you think?

then we can simplify and use `pypa/gh-action-pypi-publish` mind check again?

trim21:
can't we just put this in ci.yml？ then no need for extra conditions， keep publish.yml as is

Borda:
> can't we just put this in ci.yml？ then no need for extra conditions， keep publish.yml as is

I don't think you can share artifacts across workflows

trim21:
> > can't we just put this in ci.yml？ then no need for extra conditions， keep publish.yml as is
> 
> I don't think you can share artifacts across workflows

build new files in publish.yml

Borda:
> build new files in publish.yml， like it used to be

Not sure I follow where you are going... if you copy job `build` to CI, then you can't have it twice and need to remember to update both, which could be easy to forget... if you move the build, you need to build package here again and you do not have testing that it is still correct especially if you update something in build in CI and not here

EnricoMi:
Yes, build could be entirely moved to CI, which duplicates things a bit. That would require to keep both in sync, which is not nice, but I wouldn't expect much changes here anyway.

I'd be happy either way.

@JLLeitschuh what do you think?

.github/workflows/pypi-release.yml
```diff
@@ -1,25 +1,45 @@
...
+    if: startsWith(github.event.ref, 'refs/tags/')
+    needs: [build]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: package
+          path: dist
+      - name: Publish 📦 to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.8.7
```
EnricoMi, line 42: Yes, this simplifies our code, but pulls in another action that does not add any features. It merely calls `twine upload dist/*`. I have reviewed that action and I do not like the code base. I don't want to rely on that action and pass the pypi password to it for no benefit. I am fine with setting up python and installing twine in our workflow.

EnricoMi (changes requested):

.github/workflows/pypi-release.yml
Borda: It is official pypi action...

Borda:
```suggestion
        run: |
          pip install -q twine
          twine upload dist/*
```

JLLeitschuh:
You can specify re-usable work flows that you can apply to multiple work flows.

Are we concerned that we are executing a workflow namdd "publish" when we aren't publishing? If so, just move everything for building and publishing into "CI", that way the naming isn't confusing

Borda:
> You can specify re-usable work flows that you can apply to multiple work flows.

done

EnricoMi (changes requested):
Can we not rename the `pypi-release.yml` file? Besides that, happy to approve.

Borda:
> Can we not rename the `pypi-release.yml` file? Besides that, happy to approve.

@EnricoMi reverted rename

EnricoMi (approved):
LGTM!

PR #2595, opened 2023-07-03
precommit: move `flake8` as last

Borda:
`flake8` is just a read-only check so does not perform any changes compared to pyupgrade so it makes better sense to move it as last as, eventually, pyuprade could make a change after flake validation, and based on the actual flake8 configuration it may not be compatible so next user will see failing lint even any of his changes...


trim21 (approved):
👏

EnricoMi (approved):
LGTM!

PR #2596, opened 2023-07-03
pkg: move package to src/

Borda:
I would like suggest moving the package to `src/` folder, which is Python's standard to prevent eventual name collision and confusion with GitHub internal folder `.github/` :chipmunk:


trim21:
I don't think it's worth doing...

EnricoMi:
Not sure either...

Borda:
> I don't think it's worth doing...

I beilve it is good separateion but feel free to close it 🦩

PR #2598, opened 2023-07-04
unify requirements

Borda:
let's move all development requirements to a common folder and remove sphinx requirements from the package base dependency tree :rabbit:


Borda:
@sfdye @EnricoMi could you pls have look / review :flamingo:

EnricoMi:
I like this. @JLLeitschuh @trim21 what do you think?

trim21:
I think the main `requirements.txt` can also be removed, it's duplicated with `setup.py`

Borda:
> Or read this file in setup.py instead of write them twice.

Rather read in setup.py as dependabot easier pick these from file

trim21:
`pkg_resources` is deprecated

setup.py
```diff
@@ -40,9 +39,20 @@
...
 import setuptools
+from pkg_resources import parse_requirements
+
+_PATH_ROOT = os.path.dirname(__file__)
+_PATH_REQUIRES = os.path.join(_PATH_ROOT, "requirements")
+
+
+def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> list:
+    reqs = parse_requirements(open(os.path.join(path_dir, file_name)).readlines())
+    return list(map(str, reqs))
```
trim21, line 54:
```suggestion
def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> list[str]:
    with open(os.path.join(path_dir, file_name)) as f:
        return [x for x in f if x and not x.startswith('#')]
```

setup.py
```diff
@@ -40,9 +39,20 @@
 #                                                                              #
 ################################################################################
 
+import os
 import textwrap
 
 import setuptools
+from pkg_resources import parse_requirements
```
Borda, line 46:
```suggestion
```

setup.py
trim21: Oh, forget to quote `list[str]` so it's broken in old python version

setup.py
```diff
@@ -40,10 +39,20 @@
...
 
 import setuptools
 
+_PATH_ROOT = os.path.dirname(__file__)
+_PATH_REQUIRES = os.path.join(_PATH_ROOT, "requirements")
+
+
+def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> list[str]:
+    with open(os.path.join(path_dir, file_name)) as f:
+        return [x for x in f if x and not x.startswith('#')]
```
Borda, line 53:
```suggestion
    with open(os.path.join(path_dir, file_name)) as fo:
        return [r.strip() for r in fo.readlines() if x and not x.startswith('#')]
```

setup.py
```diff
@@ -40,10 +39,20 @@
...
 
 import setuptools
 
+_PATH_ROOT = os.path.dirname(__file__)
+_PATH_REQUIRES = os.path.join(_PATH_ROOT, "requirements")
+
+
+def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> list[str]:
+    with open(os.path.join(path_dir, file_name)) as fo:
+        return [r.strip() for r in fo.readlines() if x and not x.startswith('#')]
```
trim21, line 53:
```suggestion
def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> "list[str]":
    with open(os.path.join(path_dir, file_name)) as fo:
        return [r.strip() for r in fo.readlines() if r.strip() and not r.strip().startswith("#")]
```

ci is still broken

setup.py
```diff
@@ -40,10 +39,20 @@
...
 
 import setuptools
 
+_PATH_ROOT = os.path.dirname(__file__)
+_PATH_REQUIRES = os.path.join(_PATH_ROOT, "requirements")
+
+
+def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> "list[str]":
+    with open(os.path.join(path_dir, file_name)) as fo:
+        return [r.strip() for r in fo.readlines() if r.strip() and not r.strip().startswith("#")]
```
EnricoMi, line 53: I think this makes the list comprehension more readable (while `strip()` is being called only once):
```suggestion
def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> "list[str]":
    with open(os.path.join(path_dir, file_name)) as fo:
        return [requirement
                for line in fo.readlines()
                for requirement in [line.strip()]
                if requirement and not requirement.startswith("#")]
```

setup.py
```diff
@@ -40,10 +39,20 @@
...
 import textwrap
 
 import setuptools
 
+_PATH_ROOT = os.path.dirname(__file__)
+_PATH_REQUIRES = os.path.join(_PATH_ROOT, "requirements")
+
+
+def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> "list[str]":
+    with open(os.path.join(path_dir, file_name)) as fo:
```
EnricoMi, line 52: pre-commit code spell does not like `fo`, maybe `f` or `r` works better

setup.py
trim21: maybe `[r for r in (x.strip() for x in f.readlines()) if r and not r.startswith("#")]` for more readable

Borda:
```suggestion
def _load_requirements(path_dir: str = _PATH_ROOT, file_name: str = "requirements.txt") -> "list[str]":
    with open(os.path.join(path_dir, file_name)) as fo:
        reqs = [ln.strip() for ln in fo.readlines()]
    return [r for r in reqs if r and not r.startswith("#")]
```

EnricoMi (approved):
LGTM!

trim21 (approved):
LGTM

PR #2594, opened 2023-07-03
CI: drop precommit job, use bot, move docs into CI workflow

Borda:
Moving docs build to CI checks as this is not much lint but real check
Also, I may suggest installing recommit [bot](https://precommit.ci/) instead of GH action which has the advantage of eventually automatically fixing open PRs

cc: @EnricoMi

PR #2599, opened 2023-07-04
Set line length to 120 characters

Borda:
cc: @EnricoMi


EnricoMi:
@JLLeitschuh @trim21 what do you think?

trim21:
I suggest add a config file for black instead of hard-coded argument in pre-commit

Borda:
> I suggest add a config file for black instead of hard-coded argument in pre-commit

Added `pyproject.toml` which we can also use for #2617

pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.black]
+# https://github.com/psf/black
+line-length = 120
+exclude = "(.eggs|.git|.hg|.mypy_cache|.venv|_build|buck-out|build|dist)"
```
trim21, line 4: I think the default exclude already cover this, isn't it?

Borda:
```suggestion
```

trim21:
maybe also add `target-version = ['py37']`

Borda:
> target-version = ['py37']

does not have any effect, but sdded

trim21:
> > target-version = ['py37']
> 
> does not have any effect, but sdded

looks like it does 

https://github.com/PyGithub/PyGithub/pull/2599/commits/0e49ad8b6c5b1a39f4c22e60723c3f35dbdda9cf#diff-d2a693bccbc5953aa099d38d5c73594e3cec275e229b18c820f80b43a64a51c0

trim21 (approved):
LGTM

pyproject.toml
```diff
@@ -1,4 +1,3 @@
 [tool.black]
 # https://github.com/psf/black
 line-length = 120
-target-version = ['py37']
```
trim21: why removed?

Borda:
> > setting 3.7 in #2594
> 
> no, it's not the same thing. tools like black and ruff doesn't infer target-version from the python version it's running.

reverted

trim21:
lgtm

trim21:
one more thing: 

https://github.com/PyGithub/PyGithub/blob/61f9ae6c4b943124dca93c14de54506a7661d68e/.isort.cfg#L6

JLLeitschuh (approved):
LGTM!

EnricoMi (approved):
LGTM!

EnricoMi:
> one more thing:
> 
> https://github.com/PyGithub/PyGithub/blob/61f9ae6c4b943124dca93c14de54506a7661d68e/.isort.cfg#L6

@Borda what about this line length setting?

Borda:
> @Borda what about this line length setting?

thx for reminder, yes, just set in last commit to be 120 too

EnricoMi:
Awesome, thanks!

PR #2601, opened 2023-07-04
group patch for an issue into a folder

Borda:
I would suggest for navigation benefits to group all patches for specific issues to the folder to separate from other core API testing.

NOTE: also free to lose it if you do not agree with it :chipmunk: 

cc: @EnricoMi


Borda:
fyi, @EnricoMi, closing for no activity... :(

PR #2576, opened 2023-06-27
Merge some simple classes' type stubs back to source 

trim21:
This is a split of #2570, https://github.com/PyGithub/PyGithub/pull/2570 is too big, this make is easy to review.

PR #2570, opened 2023-06-25
Merge some simple class' type stubs back to source

trim21:
some classes with less equal than 5 properties

PR #2602, opened 2023-07-04
merge annotation with `retype`

Borda:
as mentioned in https://github.com/PyGithub/PyGithub/issues/2589#issuecomment-1620480319 as min Python version is 3.7 we do not need to have annotation in external files, and all can be in this package... Some advantages:

- typing, defaults, and method/function footprint lives in the same place
- reduce the number of files in the uniform folder
- simplify navigation and development as pyi are not commonly used any more


trim21:
I tried this beforee, but it won't work for some reasons:

1. circle import
2. There is a old style `... is NotSet` or `... is not NotSet` and need to be updated to `is_defined` or `is_undefined`
3. some type stub are wrong, and some inline code can't narrow type, so it failed on mypy

Borda:
> I tried this beforee, but it won't work for some reasons:
> 
> 1. circle import
> 2. There is a old style `... is NotSet` or `... is not NotSet` and need to be updated to `is_defined` or `is_undefined`
> 3. it doesn't add instance attributes
> 4. some type stub are wrong, and some inline code can't narrow type, so it failed on mypy

so how about making this per file and trying to resolve it in smaller steps?

trim21:
> > I tried this beforee, but it won't work for some reasons:
> > 
> > 1. circle import
> > 2. There is a old style `... is NotSet` or `... is not NotSet` and need to be updated to `is_defined` or `is_undefined`
> > 3. it doesn't add instance attributes
> > 4. some type stub are wrong, and some inline code can't narrow type, so it failed on mypy
> 
> so how about making this per file and trying to resolve it in smaller steps?

That's what I'm doing https://github.com/PyGithub/PyGithub/pulls/trim21

trim21:
I have done many classes in https://github.com/PyGithub/PyGithub/pull/2570 , would you mind choose the rest class, so neither of us will do useless work?

Borda:
closing in favor of #2576

Borda:
> would you mind choose the rest class, so neither of us will do useless work?

ok, I can, let's just wait till yours is merged and I can continue

PR #2607, opened 2023-07-04
Enable mypy `check_untyped_defs`

trim21:
make mypy check untyped function body.

1. fix existing type error ignored due to this option disabled
3. mypy section in tox.ini doesn't work, move it to pyproject.toml

`scripts/add_attribute.py` maybe broken, I'll fix it after https://github.com/PyGithub/PyGithub/pull/2609  is merged.

close https://github.com/PyGithub/PyGithub/issues/2628


github/RepositoryAdvisory.py
```diff
@@ -326,97 +279,75 @@ def clear_credits(self):
...
-            description, str
-        ), description
-        assert (
-            severity_or_cvss_vector_string is github.GithubObject.NotSet
-            or isinstance(severity_or_cvss_vector_string, str)
-        ), (severity_or_cvss_vector_string)
-        assert cve_id is github.GithubObject.NotSet or isinstance(cve_id, str), cve_id
-        assert vulnerabilities is github.GithubObject.NotSet or isinstance(
-            vulnerabilities, typing.Iterable
+        assert summary is NotSet or isinstance(summary, str), summary
```
trim21, line 294: this can be refactored with new type-assert utils, but I'm hoping not to touch much logic in this PR

trim21:
@EnricoMi

trim21:
@EnricoMi @JLLeitschuh any review？

github/Authorization.py
```diff
@@ -43,21 +45,22 @@ class Authorization(github.GithubObject.CompletableGithubObject):
...
 
-    _app: Attribute["AuthorizationApplication"]
-    _created_at: Attribute[datetime]
-    _id: Attribute[int]
-    _note: Attribute[Optional[str]]
-    _note_url: Attribute[Optional[str]]
-    _scopes: Attribute[str]
-    _token: Attribute[str]
-    _updated_at: Attribute[datetime]
-    _url: Attribute[str]
```
JLLeitschuh, line 54: I actually like these explicit field declarations (I'm a Java/kotlin developer primarily)

trim21: declare attribute in `def _initAttributes(self):` so we don't need to write these twice and keep them syncing

github/CWE.py
```diff
@@ -20,35 +20,27 @@
...
 
     @property
     def name(self) -> str:
-        """
-        :type: string
-        """
         return self._name.value
 
-    # noinspection PyPep8Naming
     def _initAttributes(self):
```
EnricoMi, line 40: Can we move this up to become the first method here and in `Environment`?

EnricoMi (changes requested):
LGTM, minor comment.

EnricoMi (approved):
LGTM!

PR #2608, opened 2023-07-04
add pre-commit hook https://github.com/PyCQA/autoflake to automaticly remove unused imports

trim21:


Borda:
I may suggest rather to use ruff instead #2617

trim21:
> I may suggest rather to use ruff instead #2617

I agree, ruff is a better idea.

PR #2609, opened 2023-07-05
Enable mypy `disallow_untyped_defs`

trim21:
depends on https://github.com/PyGithub/PyGithub/pull/2607


trim21:
@EnricoMi 
image

tests/ReplayData/PullRequestReview.setUp.txt
```diff
@@ -18,7 +18,7 @@ None
 None
```
EnricoMi, line 18: why are these two replay data files changing? We ave no changes in test files and github files only touch typing.

trim21: There are some code in these 2 replay data and change when running global file replace.

Reverted.

trim21:
we can also enable this option progresive instead or a huge 133 files PR

```toml
[[tool.mypy.overrides]]
module = ["github.AuthenticatedUser"]
check_untyped_defs = true
disallow_untyped_defs = true
```

EnricoMi (approved):
LGTM!

PR #2610, opened 2023-07-06
Use timezone-aware reset datetime in GithubRetry.py

soulcheck:
`reset` and the current datetime used to compute the time delta after which retry should happen were timezone-unaware and timezone-aware respectively, resulting in 

`TypeError: can't subtract offset-naive and offset-aware datetimes` being thrown.

This PR changes `datetime.utcfromtimestamp(int)` to `datetime.fromtimestamp(int, timezone.utc)` making both `reset` and current datetimes timezone-aware.

Tests have been updated to reflect that.

### DIFF
--- a/tests/GithubRetry.py
+++ b/tests/GithubRetry.py
@@ -22,7 +22,7 @@
 import contextlib
 import sys
 import unittest
-from datetime import datetime, timezone
+from datetime import datetime
 from io import BytesIO
 from unittest import mock
 
--- a/tests/GithubRetry.py
+++ b/tests/GithubRetry.py
@@ -132,7 +132,7 @@ class GithubRetry(unittest.TestCase):
         else:
             attr = "github.GithubRetry._GithubRetry__datetime"
         with mock.patch(attr) as dt:
-            dt.now = mock.Mock(return_value=datetime.fromtimestamp(now, timezone.utc))
+            dt.now = lambda tz=None: datetime.fromtimestamp(now, tz=tz)
             dt.fromtimestamp = datetime.fromtimestamp
             yield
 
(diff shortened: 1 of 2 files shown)


tests/GithubRetry.py
```diff
@@ -132,8 +132,8 @@ def mock_retry_now(self, now):
         else:
             attr = "github.GithubRetry._GithubRetry__datetime"
         with mock.patch(attr) as dt:
-            dt.now = mock.Mock(return_value=datetime.utcfromtimestamp(now))
-            dt.utcfromtimestamp = datetime.utcfromtimestamp
+            dt.now = mock.Mock(return_value=datetime.fromtimestamp(now, timezone.utc))
```
EnricoMi, line 135: The Mock is not relly needed to patch the method, let's remove it
```suggestion
            dt.now = lambda tz=None: datetime.fromtimestamp(now, tz=tz)
```

EnricoMi (changes requested):
Thanks for testing the release candidate and fix this bug. Good test coverage with bad mocking is just bad. Minor clean up while we are here.

tests/GithubRetry.py
soulcheck: Thanks!

soulcheck: Applied!

soulcheck: Removed the unused import too.

EnricoMi (approved):
LGTM!

EnricoMi:
Awesome, thanks a lot!

EnricoMi:
This has been released as pre-release 2.0.0rc1.

PR #2611, opened 2023-07-06
Add retry issue to GithubException, don't log it

EnricoMi:
Adds cause of retry failures to the exception instead of logging it. This improves identifying issues where no logging is available.

### DIFF
--- a/github/GithubRetry.py
+++ b/github/GithubRetry.py
@@ -112,7 +112,15 @@ class GithubRetry(Retry):
                         content = self.get_content(response, url)
                         content = json.loads(content)
                         message = content.get("message")
+                    except Exception as e:
+                        # we want to fall back to the actual github exception (probably a rate limit error)
+                        # but provide some context why we could not deal with it without another exception
+                        try:
+                            raise RuntimeError("Failed to inspect response message") from e
+                        except RuntimeError as e:
+                            raise GithubException(response.status, content, response.headers) from e
 
+                    try:
                         if Requester.isRateLimitError(message):
                             rate_type = (
                                 "primary"
@@ -191,11 +199,12 @@ class GithubRetry(Retry):
                     except (MaxRetryError, GithubException):
                         raise
                     except Exception as e:
-                        self.__log(
-                            logging.WARNING,
-                            "Failed to inspect response message",
-                            exc_info=e,
-                        )
+                        # we want to fall back to the actual github exception (probably a rate limit error)
+                        # but provide some context why we could not deal with it without another exception
+                        try:
+                            raise RuntimeError("Failed to determine retry backoff") from e
+                        except RuntimeError as e:
+                            raise GithubException(response.status, content, response.headers) from e
 
                     raise GithubException(response.status, content, response.headers)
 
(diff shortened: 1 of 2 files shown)


github/GithubRetry.py
```diff
@@ -112,7 +112,19 @@ def increment(
...
                         content = json.loads(content)
                         message = content.get("message")
+                    except Exception as e:
+                        # we want to fall back to the actual github exception (probably a rate limit error)
+                        # but provide some context why we could not deal with it without another exception
+                        try:
+                            raise RuntimeError(
+                                "Failed to inspect response message"
+                            ) from e
+                        except RuntimeError as e:
```
trim21, line 122: why make it so complicated here

trim21: Is this the reason you suggest this ?

https://github.com/PyGithub/PyGithub/pull/2591#issuecomment-1619670289

EnricoMi: The response from Github causes the `GithubException`, which we want to prevent by retrying, but the logic to determine the right backoff raised some exception itself, which should never happen. So I want to preserve the inner exception with stacktrace to better understand and debug the root cause, amend it with some meaningful error message (`RuntimeError`), but still provide the usual `GithubException` based on the response from Github.

The chain is more appropriate here than adding an error message to the outer `GithubException`.

I could not find a nicer syntax to do the chaining, something like `raise GithubException() from RuntimeError() from e`.

trim21: maybe exception group? https://peps.python.org/pep-0654/

trim21: I think we can add a extra message field in GithubException and use `"Failed to inspect response message"` as exception message?

EnricoMi: Yes, we should add the message field, but I would not use it here, because that error message is not coming from / caused by Github, but from PyGithub library. Hence the cause chain.

JLLeitschuh (approved):
Seems reasonable to me

github/GithubRetry.py
Borda, line 122: Maybe I miss something, but why raise an exception that you are also immediately absorbing so does not go anywhere...

github/GithubRetry.py
```diff
@@ -191,11 +203,16 @@ def get_backoff_time():
...
+                        # we want to fall back to the actual github exception (probably a rate limit error)
+                        # but provide some context why we could not deal with it without another exception
+                        try:
+                            raise RuntimeError(
+                                "Failed to determine retry backoff"
+                            ) from e
+                        except RuntimeError as e:
+                            raise GithubException(
+                                response.status, content, response.headers
+                            ) from e
```
Borda, line 215:
```suggestion
                        raise GithubException(
                            response.status, "Failed to determine retry `backoff`. " + content, response.headers
                        ) from e
```

github/GithubRetry.py
EnricoMi: To create an exception chain, GithubException caused by RuntimeError, caused by e.

github/GithubRetry.py
EnricoMi: The `GithubException` carries the response from Github (status, JSON content, headers), but the `"Failed to determine retry backoff"` bit is coming from our poor code, I think that should be separated. Github does not send that, hence the `RuntimeError`.

PR #2613, opened 2023-07-07
Add SAML exception

Borda:
recently crossed this exception and think would be nice to catch it...

cc: @EnricoMi


github/Requester.py
```diff
@@ -542,20 +541,20 @@ def createException(
...
-        elif status == 403 and message.startswith(
-            "missing or invalid user agent string"
-        ):
-            exc = GithubException.BadUserAgentException
-        elif status == 403 and cls.isRateLimitError(message):
-            exc = GithubException.RateLimitExceededException
+        if status == 401:
+            if message == "bad credentials":
+                exc = GithubException.BadCredentialsException
+            if "SAML enforcement" in message:
```
trim21, line 547: message is in lower case

trim21 (changes requested):

github/Requester.py
Borda:
```suggestion
            if "saml enforcement" in message.lower():
```

github/Requester.py
```diff
@@ -544,7 +544,7 @@ def createException(
         if status == 401:
             if message == "bad credentials":
                 exc = GithubException.BadCredentialsException
-            if "SAML enforcement" in message:
+            if "saml enforcement" in message.lower():
```
trim21, line 547: message is already lower

Borda:
```suggestion
            if "saml enforcement" in message:
```

trim21 (approved):
lgtm

EnricoMi:
Please request review so this flags up for me again.

PR #2617, opened 2023-07-07
Replace `flake8` with `ruff`

Borda:
compare to flake8 ruff also perform fixes and enable later on replace multiple other tools but this one which significantly speed up lint

cc: @EnricoMi


trim21:
I'm using ruff myself, it's good and fast, but it's not stable yet. Almost every version contain breaking change.

trim21:
If we are going to use ruff， i suggest to use a config file for  select and ignore instead of pre commit argument

trim21:
Also pyupgrade and isort can also be replaced by ruff

Borda:
> If we are going to use ruff， we should use a config file for select and ignore instead of pre commit argument.

yes, we can add it by enabling more options but this flake8 parity is just with these ignores...

trim21:
> but this flake8 parity is just with these ignores...

My point is using a config file will help developers using pip installed ruff to have same ruff config. I have no opinion on what rules we should enable or disable...

trim21:
also I think this hook should be put before black

Borda:
> My point is using a config file will help developers using pip installed ruff to have same ruff config

yes and we almost had it, but https://github.com/PyGithub/PyGithub/pull/2530#issuecomment-1620958461

trim21:
> > My point is using a config file will help developers using pip installed ruff to have same ruff config
> 
> yes and we almost had it, but [#2530 (comment)](https://github.com/PyGithub/PyGithub/pull/2530#issuecomment-1620958461)

then use `ruff.toml`...

trim21:
Or, it's still able to create a pyproject.toml file add ruff config, without break existing workflow.

Borda:
> Or, it's still able to create a pyproject.toml file add ruff config, without breaking existing workflow.

adding `pyproject.toml` and already setting line length to 120, ref: #2599

trim21:
Need to put ruff hook (with --fix) before black, some output of ruff are not black-ed

maybe also drop isort?

pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+    "E", "W",  # see: https://pypi.org/project/pycodestyle
+    "F",  # see: https://pypi.org/project/pyflakes
+    "I",  # see: isort
+]
+ignore = [
+    "E501",
+    "E741",
+]
+# Exclude a variety of commonly ignored directories.
+exclude = [
```
trim21, line 14: also doesn't look necessary

Borda:
> Need to put ruff hook (with --fix) before black, some output of ruff are not black-ed

until @s-t-e-v-e-n-k @EnricoMi or any other admin install recommit bot it does not have much meaning as you just run it twice as suggested in #2594

pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+    ".eggs",
+    ".git",
+    ".mypy_cache",
+    ".ruff_cache",
+    "__pypackages__",
+    "_build",
+    "build",
+    "dist",
+    "docs"
+]
```
Borda, line 24:
```suggestion
```

trim21:
> > Need to put ruff hook (with --fix) before black, some output of ruff are not black-ed
> 
> until @s-t-e-v-e-n-k @EnricoMi or any other admin install precommit bot it does not have much meaning as you just run it twice as suggested in #2594

This means you don't need to run `git commit -m '...'` three times, you only need to run it twice.

Borda:
> and I think ruff config also need `target-version` option

I think in such case we shall just set py version in the recommit

```
default_language_version:
  python: python3.7

File: pyproject.toml [hunk 3]
```diff
@@ -0,0 +1,14 @@
+[tool.ruff]
+line-length = 120
+# Enable Pyflakes `E` and `F` codes by default.
+select = [
+    "E", "W",  # see: https://pypi.org/project/pycodestyle
+    "F",  # see: https://pypi.org/project/pyflakes
+    "I",  # see: isort
```
trim21, line 7: If we enable ruff's isort feature we should remove isort related hooks and config from pre-commit config

trim21 (changes requested):

trim21:
and 

> Need to put ruff hook (with --fix) before black, some output of ruff are not black-ed

Borda:
> I don't understand why you against this, this not only affect pre-commit ci but also affect contributors creating git commit.

You say that ruff may not be black compatible, but are you sure that black is 100% compatibile with all ruff configs?

trim21:
> You say that ruff may not be black compatible, but are you sure that black is 100% compatibile with all ruff configs?

I still don't understand, why black need to be 100% compatibile with all ruff configs?

If a ruff rule conflicts with black then we should disable that ruff rule, black is almost unconfigurable. Just the same thing we did when using flake8.

> but also affect contributors creating git commit

in some case, the ruff will rewrite code and the result is not black-ed, when a developer commit their code, the first time it fails with ruff, the second time it fails black, then it can only pass in the third times.

pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+]
+ignore = [
+    "E501",
+    "E741",
+]
+ignore-init-module-imports = true
+unfixable = ["F401"]
```
trim21, line 14: I think we can make all rules as fixable in pre-commit hooks

Borda:
> If a ruff rule conflicts with black then we should disable that ruff rule, black is almost unconfigurable. This is just the same thing we did when using flake8.

I disagree as black is just formatting, it is the same as black is superior to flake8 which I believe is not correct

Flake8 for example shall be the last step as it validate code correctness even after black changes... 

> in some case, the ruff will rewrite code and the result is not black-ed,

Can you give me example?

Borda:
Over all, we can call ruff twice, once with fix and at the end just as check if it makes you happy...

pyproject.toml
Borda: Why? In some comment before you complained about having config in precommit...

trim21:
this one https://github.com/astral-sh/ruff/issues/5368 , for example

I don't think.

> Flake8 for example shall be the last step as it validate code correctness even after black changes...

That's because flake8 pre-commit hook doesn't change any code, it's `ruff check` without `--fix`

trim21:
> Over all, we can call ruff twice, once with fix and at the end just as check if it makes you happy...

No, if the first ruff hook change any files, this pre-commit run will finally failed.

It's meanless to run it twice.

Borda:
> for example, rewrite `Optional[T]` to `T|None` is not black-ed

True but the reason os target Python version not ruff itself...

trim21:
I just post a more real-world example, you can try it yourself.

Borda:
> > Over all, we can call ruff twice, once with fix and at the end just as check if it makes you happy...
> 
> No, if the first ruff hook change any files, this pre-commit run will finally failed.
> 
> It's meanless to run it twice.

This is meaningless to argue and waisted time, I just 

> ruff will rewrite it to this:
> 
> ```
> 
> 
> _: str | None
> ```
> 
> which is not blacked, black will format the rewrite to this:
> 
> ```
> _: str | None
> ```

Looks identical to me...

trim21:
> Looks identical to me...

black doesn't, that the point.

trim21:
There are blank lines black wants to remove, and ruff doesn't remove.

trim21:
This is only a example I have in mind, there is another simple case that ruff rewrite code and generate non black-ed code, but it will finally pass.

you can test with this example (which is already black-ed) with current PR config, to see how many times you need to run pre-commit until it pass.

```python
f = lambda x: 2 * x
_ = 1
```

pyproject.toml
trim21: because config file is for both IDE/editor plugin and pre-commit hooks, I think this rule should be unfixable in IDE/editor plugin and fixed in pre-commit hook.

trim21: This one is the https://github.com/PyGithub/PyGithub/pull/2608

Borda:
@trim21 you say that black is superior to flake8, correct? Also you say that black canot ever generate code which not complied with ruff...

trim21:
> you say that black is superior to flake8, correct? 

I'm not a native English speaker, so I don't understand why you say this, is there some misunderstand here?

I don't think we can compare black with flake8, nor "black is superior to flake8".
In this PR we want to replace flake8 with ruff, not black. black and flake8 are not same kind of tools.

> Also you say that black canot ever generate code which not complied with ruff...

I'm not sure.

But if it happened, pre-commit will still fail, and it can't pass, either it fails the balck hook or it fails the ruff hook. At that time we can adjust the ruff config.

pyproject.toml
Borda: No, you need to be consistent

Borda:
> In this PR we want to replace flake8 with ruff, not black.

That is exactly what I did but you suggest to change the hierarchy too

> black and flake8 are not same kind of tools.

Exactly, the same as black and ruff are not same kind of tools

trim21:
> Appreciate your creativity with generating examples, please report them on ruff repo, but do you have real example from this code base?

This is from the experience using buth ruff and black in pre-commit, you almost always need to run 3 times if you put ruff after black and when you trigger ruff to auto-fix your code.

But this is not a bug of ruff, ruff never says it generate code black-ed, it's a lint tool not formatter.

>> In this PR we want to replace flake8 with ruff, not black.
>
> That is exactly what I did but you suggest to change the hierarchy too
>
>> black and flake8 are not same kind of tools.
>
> Exactly, the same as black and ruff are not same kind of tools

Oh, I understand. in this case, yes. if the "superior" is pre-commit hooks order.

I think as pre-commit hooks, `ruff --fix`(any hooks not may generate non black-ed code) > `black` > `ruff`/`flake8`(check only lint hooks, if it may affect by black).

the reason is the ruff(`--fix`) change codes and it's output can and need be fixed by black, putting ruff before black could reduce roundtrip.

>> But if it happened, pre-commit will still fail, and it can't pass, either it fails the balck hook or it fails the ruff hook. At that time we can adjust the ruff config
>
> Then we will fix it, world is never perfect so why spend so much time on case which may never happen...

This is the answer to "Also you say that black canot ever generate code which not complied with ruff...", is not fixable in this PR, we are never discussing this scenario.

pyproject.toml
trim21, line 14: IMO there is no consistent need in this specific case (if a rule can be auto-fixable).

`F401` is a enabled rule and is will always raise a error in all case, it's "consistent" already.

Some developer (for example, myself) are using [ruff's IDE plugin](https://beta.ruff.rs/docs/editor-integrations/), which is able to fix ruff error on-save.

So make this rule auto-fixable is not productive and annoying. But it's productive in the pre-commit hooks

It's just like the autoflake8(`--remove-all-unused-imports`) pre-commit hook, we only want it to be auto-fixable in the pre-commit hooks, not when you are editing codes.

.pre-commit-config.yaml
```diff
@@ -33,7 +26,9 @@ repos:
...
           - --py37-plus
-  - repo: https://github.com/pycqa/flake8
-    rev: 3.9.2
+
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: 'v0.0.277'
     hooks:
-      - id: flake8
+      - id: ruff
+        args: ["--fix"]
```
trim21, line 34:
```suggestion
        args: ["--fix", "--fixable=ALL"]
```

Borda:
lets wait till @EnricoMi or @s-t-e-v-e-n-k confirm this direction before investing more time

Borda:
@sfdye @EnricoMi could you pls have look / review :flamingo:

EnricoMi:
Looks good, moving to ruff is fine with me.

EnricoMi:
https://github.com/PyGithub/PyGithub/pull/2617/files#r1263145358 maybe

trim21:
maybe also add argument `--exit-non-zero-on-fix`

and repo https://github.com/charliermarsh/ruff-pre-commit moved to https://github.com/astral-sh/ruff-pre-commit

Borda:
> [#2617 (files)](https://github.com/PyGithub/PyGithub/pull/2617/files#r1263145358) maybe

This is very subjective, so let's evaluate this non-default argument in a special PR?

Borda:
> maybe also add argument `--exit-non-zero-on-fix`
> 
> and repo [charliermarsh/ruff-pre-commit](https://github.com/charliermarsh/ruff-pre-commit) moved to [astral-sh/ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit)

let's do it as part of #2594

trim21:
> > maybe also add argument `--exit-non-zero-on-fix`
> > and repo [charliermarsh/ruff-pre-commit](https://github.com/charliermarsh/ruff-pre-commit) moved to [astral-sh/ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit)
> 
> let's do it as part of #2594

this has nothing to do with https://github.com/PyGithub/PyGithub/pull/2594

trim21:
> maybe also add argument `--exit-non-zero-on-fix`
> and repo [charliermarsh/ruff-pre-commit](https://github.com/charliermarsh/ruff-pre-commit) moved to [astral-sh/ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit)

why adding argument and fix repo address has any thing to do with pre-commit ci?

trim21:
charliermarsh/ruff-pre-commit has been moved to https://github.com/astral-sh/ruff-pre-commit

Borda:
> [charliermarsh/ruff-pre-commit](https://github.com/charliermarsh/ruff-pre-commit) has been moved to [astral-sh/ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit) , there is no reason we can't update pre-commit hook repo address in this PR

updated

EnricoMi (approved):
LGTM!

EnricoMi:
@trim21 your thoughts?

PR #2623, opened 2023-07-13
Add Secret and Variable classes

nthings:
Hi! this PR is for adding the Secret and Variable classes. Before I added the variable methods to Organization and Repository but I think is better handling this way. Let me know feedback if this is good so I can start adding test cases. Ty!


trim21:
No need to add `pyi` for new classes, just use inline types.

github/Repository.py
```diff
@@ -1617,7 +1677,7 @@ def create_variable(self, variable_name: str, value: str) -> bool:
         status, headers, data = self._requester.requestJson(
             "POST", f"{self.url}/actions/variables", input=post_parameters
         )
-        return status == 201
+        return github.Variable.Variable(self._requester, headers, data, completed=True)
 
     def update_variable(self, variable_name: str, value: str) -> bool:
```
EnricoMi, line 1682: There should also be an `edit` method in class `Variable` and `Secret`. Then it is arguable to remove this (reduces footprint of Repository class).

github/Repository.py
```diff
@@ -1599,7 +1629,37 @@ def create_secret(self, secret_name, unencrypted_value):
...
+        assert isinstance(variable_name, str), variable_name
+        return github.Variable.Variable(
+            requester=self._requester,
+            headers={},
+            attributes={"url": f"{self.url}/actions/variables/{variable_name}"},
+            completed=False,
+        )
 
     def create_variable(self, variable_name: str, value: str) -> bool:
         """
```
EnricoMi, line 1665: Please update the urls, here from

    https://docs.github.com/en/rest/reference/actions/variables#create-a-repository-variable

to

    https://docs.github.com/en/rest/actions/variables#create-a-repository-variable

github/Variable.py
```diff
@@ -0,0 +1,149 @@
...
+        return self._url.value
+
+    def delete(self):
+        """
+        :calls: `DELETE {variable_url} <https://docs.github.com/en/rest/actions/variables>`_
+        :rtype: None
+        """
+        headers, data = self._requester.requestJsonAndCheck("DELETE", self.url)
+
+    def add_repo(self, repo):
```
EnricoMi, line 99: Maybe note that this only works for organization variables with "selected" visibility, not for repo variables.

github/Organization.py
```diff
@@ -609,6 +611,36 @@ def create_repo(
...
+        """
+        assert isinstance(secret_name, str), secret_name
+        return github.Secret.Secret(
+            requester=self._requester,
+            headers={},
+            attributes={"url": f"{self.url}/actions/secrets/{secret_name}"},
+            completed=False,
+        )
+
     def create_secret(
```
EnricoMi, line 644: Maybe we put `create_secret` before `get_secret` and `get_secrets`.

github/Secret.py
```diff
@@ -0,0 +1,138 @@
...
+# along with PyGithub. If not, see <http://www.gnu.org/licenses/>.             #
+#                                                                              #
+################################################################################
+
+import github
+import github.GithubObject
+import github.Repository
+
+
+class Secret(github.GithubObject.CompletableGithubObject):
```
EnricoMi, line 28: Please see https://github.com/PyGithub/PyGithub/commit/8bdfc3a0d7fc7391ff34e60e20a6987433025417 for some examples how to inline typing into the `.py` file.

github/Secret.py
```diff
@@ -0,0 +1,138 @@
...
+    def _useAttributes(self, attributes):
+        if "name" in attributes:
+            self._name = self._makeStringAttribute(attributes["name"])
+        if "created_at" in attributes:
+            self._created_at = self._makeDatetimeAttribute(attributes["created_at"])
+        if "updated_at" in attributes:
+            self._updated_at = self._makeDatetimeAttribute(attributes["updated_at"])
+        if "visibility" in attributes:
+            self._visibility = self._makeStringAttribute(attributes["visibility"])
+        if "selected_repositories_url" in attributes:
```
EnricoMi, line 132: This means we are fetching the list of selected repositories whenever we create a secret instance, no matter if it is used or not. We like to avoid unnecessary API calls.

Further, this list of repos is paginated, so this would fetch only the first 30 repos.

Ideally, `selected_repositories` returns a paginated list of the repos:

    @property
    def selected_repositories(self) -> PaginatedList[Repository]:
        return github.PaginatedList.PaginatedList(
            github.Repository.Repository,
            self._requester,
            self.selected_repositories_url,
            {},
            list_item="repositories",
        )

github/Variable.py
EnricoMi: Alternatively, we could have `class OrganizationVariable(Variable)`, which has `add_repo` and `remove_repo`, and `Organization.get_variable` returns `OrganizationVariable`.

EnricoMi (changes requested):
I like having Secret and Variable classes.

nthings:
Addressed most of comments can I get a re-review? @EnricoMi @trim21 Also bit of help needed into the implementation of `PaginatedList` on the `selected_repositories` property

github/OrganizationVariable.py
```diff
@@ -0,0 +1,138 @@
...
+        :param repo: github.Repository.Repository
+        :rtype: bool
+        """
+        if self.visibility != "selected":
+            return False
+        self._requester.requestJsonAndCheck("DELETE", f"{self.url}/repositories/{repo.id}")
+        self._selected_repositories.value.remove(repo)
+        return True
+
+    def _initAttributes(self) -> None:
```
trim21, line 117: put `_initAttributes` before `__repr__`

trim21 (changes requested):

trim21:
Sorry, I mean move `_initAttributes`, and make it the first method of class, don't move `__repr__`

nthings:
> Sorry, I mean to move `_initAttributes`, and make it the first method of class, don't move `__repr__`

Ok. Below the properties right?

trim21:
> > Sorry, I mean to move `_initAttributes`, and make it the first method of class, don't move `__repr__`
> 
> Ok. Below the properties right?

nthings:
@trim21 reordered those methods. Do you have example of a class that one of the properties is a PaginatedList?

trim21:
> @trim21 reordered those methods. Do you have example of a class that one of the properties is a PaginatedList?

https://github.com/PyGithub/PyGithub/blob/13e178a3abcdc1acfb3dd63aecc2f5e22b7772ae/github/MainClass.py#L262-L270

nthings:
> > @trim21 reordered those methods. Do you have example of a class that one of the properties is a PaginatedList?
> 
> https://github.com/PyGithub/PyGithub/blob/13e178a3abcdc1acfb3dd63aecc2f5e22b7772ae/github/MainClass.py#L262-L270

But if you check the mypy errors and code. How can I add or delete repos from the `selected_repositories` property on the class, that logic worked with simple list before

nthings:
@trim21 I think I got it working. Please let me know any feedback so I can go ahead with test cases

EnricoMi (approved):
LGTM!

Ideally, we would have tests for `get_secret`, `get_secrets`, `create_secrets`, `get_variable`, `get_variables` and `create_variables` on `Organization` and `Repository`, assert all attributes of `Secret`, `OrganizationSecret` and `Variable` (at least in one test), and call `selected_repositories` on `OrganizationSecret` and `OrganizationVariable`.

nthings:
@EnricoMi @trim21 Test cases added!

EnricoMi (approved):
LGTM!

Issue #2445, opened 2023-03-03
Download Release Asset?


neel-m:
Can't reopen #1541 and my post to it doesn't seem to be attracting any attention.

It seems like the ability to download a release asset would be a good thing.  Here is what I think it needs:

1.  Add the function ```download_asset(self, path)``` to GitReleaseAsset.py (and .pyi).  This function just calls requestFile() in its _requester object.
2. Add the implementation of ```requestFile(self, verb, url, parameters, cnx, path)``` to Requester.py (and.pyi).

For this particular function verb, parameters, and cnx are not really needed.  The other functions in GitReleaseAsset don't seem to be passing in cnx and just get a new one from the requester.  The verb isn't really needed for this function, but might be needed for something in the future?

Not really sure what else I might be missing.  Is there anyone willing to do a review of a PR?

neel-m:
So, no interest in this at all?

PR #2624, opened 2023-07-13
Add `download_asset` to `GitReleaseAsset`

thejcannon:
This change adds `download_asset` to `GitReleaseAsset` to close the gap on missing functionality.

Unfortunately, the current code isn't designed to handle this use-case so some refactoring was necessary. The changes should only be to private methods/attributes. Ideally, the `RequestsResponse` object was handed around more freely, but I didn't want to break any public-facing APIs.

Fixes https://github.com/PyGithub/PyGithub/issues/2445


thejcannon:
Commits are useful to review independently.

thejcannon:
On second thought. Meh. `pygithub` already depends on `requests` so it's not any harder to use `requests.get(asset.browser_download_url)`.

PR #2626, opened 2023-07-14
Adding feature for enterprise consumed license

YugoHino:
Hi,

This PR is for adding the enterprise consumed licenses class which is similar to the Organisation class. This feature is only for GitHub enterprise administration users, but this might be useful for them. 
https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license?apiVersion=2022-11-28

Please let me know your feedback! Thank you so much!


github/MainClass.py
```diff
@@ -361,6 +363,22 @@ def get_organizations(self, since=github.GithubObject.NotSet):
             url_parameters,
         )
 
+    def get_enterprise_consumed_licenses(self, login):
```
EnricoMi, line 366: I'd prefer to create an `Enterprise` class that wraps the `/enterprise/{enterprise}` path prefix. That class can provide all endpoints that start with this prefix.

Then, `MainClass` would have only `get_enterprise(enterprise: str) -> Enterprise`.

Returned `Enterprise` instance then has `get_consumed_licenses() -> EnterpriseConsumedLicenses`.

github/EnterpriseConsumedLicenses.pyi
```diff
@@ -0,0 +1,23 @@
+from github.GithubObject import CompletableGithubObject
```
EnricoMi, line 1: Please embed typing of new files in the `py` file. For examples, see https://github.com/PyGithub/PyGithub/commit/8bdfc3a0d7fc7391ff34e60e20a6987433025417

EnricoMi (changes requested):

github/MainClass.py
YugoHino: Thank you for your quick feedback! Sure, I agree with that, will check and update it.

github/EnterpriseConsumedLicenses.pyi
YugoHino: Sure, will embed it in the `py` file without the `pyi` file.

YugoHino:
Hi, I updated which is based on the feedback. Please check it out. Thank you!

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        assert isinstance(self.login, str), self.login
+        firstURL = f"/enterprises/{self.login}/consumed-licenses"
+        headers, data = self._requester.requestJsonAndCheck("GET", firstURL)
+        # The response doesn't have the key of login and url, manually add it to data.
+        data["login"] = self.login
+        data["url"] = self.url + "/consumed-licenses"
+        return EnterpriseConsumedLicenses(self._requester, headers, data, completed=True)
+
+    def _initAttributes(self) -> None:
```
trim21, line 58: put `_initAttributes` before `__repr__`

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+        self._login = NotSet
+        self._url = NotSet
```
trim21, line 60:
```suggestion
        self._login: Attribute[...] = NotSet
        self._url: Attribute[...] = NotSet
```

trim21: These to suggestions also apply on `github/EnterpriseConsumedLicenses.py` and `github/NamedEnterpriseUser.py`

github/NamedEnterpriseUser.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+
+    def __repr__(self):
+        return self.get__repr__({"login": self._github_com_login.value})
+
+    @property
+    def github_com_login(self):
+        """
+        :type: string
+        """
```
trim21, line 46:
```suggestion
```

we have typing, this is not needed

github/Enterprise.py
YugoHino: Thank you. Fixed it.

github/NamedEnterpriseUser.pyi
```diff
@@ -0,0 +1,43 @@
+from typing import Any, Dict
```
trim21, line 1: remove this, use inline typing

YugoHino: Thank you. did it.

github/NamedEnterpriseUser.py
YugoHino: Thank you. did it.

github/Enterprise.py
YugoHino: Thank you. did it.

doc/examples/MainClass.rst
```diff
@@ -39,6 +39,16 @@ Get organization by name
...
 
+Get enterprise consumed licenses by name
+------------------------
+
+.. code-block:: python
+
+    >>> enterprise = g.get_enterprise_consumed_licenses("PyGithub")
+    >>> enterprise_consumed_licenses = enterprise.get_enterprise_consumed_licenses()
+    >>> enterprise_consumed_licenses.total_seats_consumed
+    u'5000'
```
trim21, line 50: why there is a `u` prefix?

Are you using python2?

YugoHino: No, I use python3. I just followed others in MainClass.rst such like u'Wan Liuyang' in Get user by name section.

trim21: I think we should remove this `u` prefix, we don't support python2 anymore.

YugoHino: Sure. Will remove it. May I remove others in the file as well?

trim21: maybe leave others as-is, not related to this PR.

YugoHino: I just removed it only this PR's one.

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+
+from github.EnterpriseConsumedLicenses import EnterpriseConsumedLicenses
+from github.GithubObject import Attribute, CompletableGithubObject, NotSet
+
+
+class Enterprise(CompletableGithubObject):
+    """
+    This class represents Enterprises. The reference can be found here https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license#list-enterprise-consumed-licenses
+    """
+
```
EnricoMi, line 32: The thing with object `Enterprise` is that it does not exist w.r.t. the Github API. We merely have this class to collect all endpoints hat start with `/enterprise/{enterprise}` under one class.

Firstly, this is `NonCompletableGithubObject`, secondly, we create it from a requester and the enterprise name:
```suggestion
class Enterprise(NonCompletableGithubObject):
    """
    This class represents Enterprises. Such objects do not exist in the Github API, so this class merely collects all endpoints the start with /enterprise/{enterprise}/. See methods below for specific endpoints and docs.
    """

    def __init__(
        self,
        requester: "Requester",
        slug_or_id: str,
    ):
        super().__init__(requester, {}, {"slug_or_id": slug_or_id, "url": s"/enterprise/{slug_or_id}"}, True)

```

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+    def _initAttributes(self) -> None:
+        self._login: Attribute[str] = NotSet
```
EnricoMi, line 34: Docs call he `enterprise` parameter

> The slug version of the enterprise name. You can also substitute this value with the enterprise id.

https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license?apiVersion=2022-11-28#list-enterprise-consumed-licenses

So I'd rather go for `slug_or_id`.

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+    @property
+    def url(self) -> str:
+        self._completeIfNotSet(self._url)
+        return self._url.value
+
+    def get_enterprise_consumed_licenses(self) -> EnterpriseConsumedLicenses:
+        """
+        :calls: `GET /enterprise/{enterprise}/consumed-licenses <https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license#list-enterprise-consumed-licenses>`_
+        """
+        assert isinstance(self.login, str), self.login
```
EnricoMi, line 54: such assertion should go into `__init__`.

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+        firstURL = f"/enterprises/{self.login}/consumed-licenses"
+        headers, data = self._requester.requestJsonAndCheck("GET", firstURL)
```
EnricoMi, line 56: rather use `self.url`:
```suggestion
        headers, data = self._requester.requestJsonAndCheck("GET", self.url + "/consumed-licenses")
```

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+        # The response doesn't have the key of login and url, manually add it to data.
+        data["login"] = self.login
```
EnricoMi, line 58: I would not add `login` to the returned instance if it is not part of the response data.

github/EnterpriseConsumedLicenses.py
```diff
@@ -0,0 +1,85 @@
...
+#                                                                              #
+################################################################################
+from typing import Any, Dict, Union
+
+from github.GithubObject import Attribute, CompletableGithubObject, NotSet, _NotSetType
+from github.NamedEnterpriseUser import NamedEnterpriseUser
+from github.PaginatedList import PaginatedList
+
+
+class EnterpriseConsumedLicenses(CompletableGithubObject):
```
EnricoMi, line 29:
```suggestion
class EnterpriseConsumedLicenses(NonCompletableGithubObject):
```

github/Enterprise.py
```diff
@@ -0,0 +1,66 @@
...
+    def login(self) -> str:
+        self._completeIfNotSet(self._login)
+        return self._login.value
+
+    @property
+    def url(self) -> str:
+        self._completeIfNotSet(self._url)
+        return self._url.value
+
+    def get_enterprise_consumed_licenses(self) -> EnterpriseConsumedLicenses:
```
EnricoMi, line 50:
```suggestion
    def get_consumed_licenses(self) -> EnterpriseConsumedLicenses:
```

github/EnterpriseConsumedLicenses.py
```diff
@@ -0,0 +1,85 @@
...
+        self._completeIfNotSet(self._login)
+        return self._login.value
+
+    @property
+    def url(self) -> str:
+        self._completeIfNotSet(self._url)
+        return self._url.value
+
+    def get_enterprise_users(
+        self, filter_: Union[str, _NotSetType] = NotSet, role: Union[str, _NotSetType] = NotSet
```
EnricoMi, line 62: Where are these filters documented?

github/EnterpriseConsumedLicenses.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        assert filter_ is NotSet or isinstance(filter_, str), filter_
+        assert role is NotSet or isinstance(role, str), role
+
+        url_parameters = {}
+        if filter_ is not NotSet:
+            url_parameters["filter"] = filter_
+        if role is not NotSet:
+            url_parameters["role"] = role
+        return PaginatedList(NamedEnterpriseUser, self._requester, self.url, url_parameters, None, "users")
```
EnricoMi, line 75: That is weird, we are requesting the same url as in `Enterprise.get_consumed_licenses`.

From the docs, it looks like we retrieve the `EnterpriseConsumedLicenses` object with an `users` attributes, for which we have retrieved the first page. There is no other endpoint that behaves like that.

I think this requires some extension of PaginatedList, or we fetch the first page again when accessing the `users` attribute (as you propose).

github/MainClass.py
```diff
@@ -341,6 +343,22 @@ def get_organizations(self, since=github.GithubObject.NotSet):
...
+        """
+        assert isinstance(login, str), login
+        firstURL = f"/enterprises/{login}"
+        # There is no native "/enterprise/{enterprise}" api, so this function is a hub for under apis of enterprise administration.
+        headers = {}
+        data = {}
+        # The response doesn't have the key of login and url, manually add it to data.
+        data["login"] = login
+        data["url"] = self.__requester.base_url + firstURL
+        return github.Enterprise.Enterprise(self.__requester, headers, data, completed=True)
```
EnricoMi, line 360: This then simplifies to
```suggestion
        assert isinstance(login, str), login
        # There is no native "/enterprise/{enterprise}" api, so this function is a hub for apis that start with "/enterprise/{enterprise}".
        return github.Enterprise.Enterprise(self.__requester, login)
```

EnricoMi (commented):
Really strange endpoint.

github/EnterpriseConsumedLicenses.py
YugoHino: Thank you for the feedback! I also thought that's weird to request the same url for both of them. However, unfortunatelly, there are two types of responses in this API. When a user want to get just "total seat puchased", the user don't have to do pagenation because the value of total seat purchased is already in the first page of response. There is a wasted time while the pagenation roop. 
> or we fetch the first page again when accessing the users attribute (as you propose). 

That's why I proposed it. This kind of mixed result is rare case, I think we shouldn't update PaginatedList for this.

github/EnterpriseConsumedLicenses.py
YugoHino: That's my mistake. Removed it.

github/Enterprise.py
YugoHino: Thank you. Fixed it.

github/Enterprise.py
YugoHino: Thank you. Fixed it.

github/Enterprise.py
YugoHino: Thank you. Fixed it.

github/Enterprise.py
YugoHino: Thank you. Fixed it.

github/Enterprise.py
YugoHino: Thank you. Fixed it.

github/MainClass.py
YugoHino: Thank you. Fixed it.

github/EnterpriseConsumedLicenses.py
EnricoMi: Before we start extending PaginatedList, can I ask you to create some simple tests for `MainClass.get_enterprise` and `Enterprise.get_consumed_licenses`? For now without iterating over `users`? This requires recording test data, which will help me understand how exactly the pagination of `/enterprises/{enterprise}/consumed-licenses` might work. Please feel free to obfuscate enterprise specific content.

github/EnterpriseConsumedLicenses.py
```diff
@@ -0,0 +1,77 @@
...
+    def enterprise(self) -> str:
+        self._completeIfNotSet(self._enterprise)
+        return self._enterprise.value
+
+    @property
+    def url(self) -> str:
+        self._completeIfNotSet(self._url)
+        return self._url.value
+
+    def get_enterprise_users(self) -> PaginatedList[NamedEnterpriseUser]:
```
EnricoMi, line 61:
```suggestion
    def get_users(self) -> PaginatedList[NamedEnterpriseUser]:
```

github/EnterpriseConsumedLicenses.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Any, Dict
+
+from github.GithubObject import Attribute, CompletableGithubObject, NotSet
+from github.NamedEnterpriseUser import NamedEnterpriseUser
+from github.PaginatedList import PaginatedList
+
+
+class EnterpriseConsumedLicenses(CompletableGithubObject):
+    """
+    This class represents Enterprises. The reference can be found here https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license#list-enterprise-consumed-licenses
```
EnricoMi, line 31:
```suggestion
    This class represents licenses consumed by enterprises. The reference can be found here https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license#list-enterprise-consumed-licenses
```

github/Enterprise.py
```diff
@@ -0,0 +1,69 @@
...
+    @property
+    def url(self) -> str:
+        return self._url.value
+
+    def get_consumed_licenses(self) -> EnterpriseConsumedLicenses:
+        """
+        :calls: `GET /enterprises/{enterprise}/consumed-licenses <https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/license#list-enterprise-consumed-licenses>`_
+        """
+        headers, data = self._requester.requestJsonAndCheck("GET", self.url + "/consumed-licenses")
+        data["url"] = self.url + "/consumed-licenses"
```
EnricoMi, line 62:
```suggestion
        if "url" not in data:
            data["url"] = self.url + "/consumed-licenses"
```

github/EnterpriseConsumedLicenses.py
YugoHino: I can create some simple tests but I actually don't have any enterprise ids for testing. Is there any of them for testing here? Or just create tests with blank on enterprise id?

EnricoMi: I thought you'd have access to some enterprise instance, otherwise, how would you know these changes work?

YugoHino: I mean, I have our company own enterprise instance and I can access and test with it, but our company cannot give the access to not a company members.

EnricoMi: There is no need  to give access to any one, simply record the data for the tests against your enterprise instance, obfuscate data where really needed and add the test files to your pull request. Once recorded, nobody needs access to the targeted enterprise instance.

YugoHino: Understand. I haven't been doing this kind of testing without actual api call, but will try to find out how to do this on.

YugoHino: @EnricoMi Sorry for late. I created the data and script for testing about Enterprise class. Please check it with it. Thank you so much.

EnricoMi:
@YugoHino good work! Thanks for the test data. I have extended `PaginatedList` to reuse the first page, see removal of redundant call that fetches first page: https://github.com/PyGithub/PyGithub/pull/2626/commits/8fa29e7bd4a7744ca4a2eb5760337d8322b5f90d#diff-38a7da26fba598a6a06f75040b037a8b9fad73dbda23e88f8f2f53e3b5c21b16L12-L22

EnricoMi (approved):
LGTM!

PR #2630, opened 2023-07-14
Add `is_alphanumeric` attribute to Autolink and `Repository.create_autolink`

janssonoskar:
Added _is_alphanumeric_ to Autolink to be able to choose between alphanumeric (default) and numeric


github/Repository.py
```diff
@@ -992,20 +992,20 @@ def compare(self, base, head):
             self._requester, headers, data, completed=True
         )
 
-    def create_autolink(self, key_prefix, url_template):
+    def create_autolink(self, key_prefix, url_template, is_alphanumeric=True):
```
trim21, line 995: please use `NotSet` as default value

trim21 (changes requested):

github/Repository.py
janssonoskar: Fixed

janssonoskar:
@EnricoMi Hi, I haven't got authentication to work to record tests, is that something you could help me with?

EnricoMi (approved):
LGTM!

PR #2641, opened 2023-07-17
Fix assertion creating pull request review comment

EnricoMi:
@heitorpolidoro looks like assertions for `start_side` and `subject_type` were wrong: https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request


heitorpolidoro:
Thanks for the fix

JLLeitschuh (approved):
LGTM!

Issue #2640, opened 2023-07-17
Review comment for a pull request with subject_type


vanya20074:
Hi,
I am trying to create comment with `subject_type:FILE` but got an error:
```
{
    "message": "Invalid request.\n\nNo subschema in \"oneOf\" matched.\n\"position\" wasn't supplied.\n\"in_reply_to\" wasn't supplied.\n\"subject_type\" is not a permitted key.\n\"line\" wasn't supplied.\nFILE is not a member of [\"line\", \"file\"].",
    "documentation_url": "https://docs.github.com/rest/pulls/comments#create-a-review-comment-for-a-pull-request"
}
```
At the same time `subject_type:file` working fine but [create_review_comment](https://github.com/PyGithub/PyGithub/blob/main/github/PullRequest.py#L450) have check with uppercase:

```
        assert subject_type is github.GithubObject.NotSet or subject_type in [
            "LINE",
            "FILE",
            "side",
        ], subject_type
```
Am I doing something wrong?

EnricoMi:
The docs say it's uppercase, so the assertion looks right: https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request

Please check with Github Support whether upper or lowercase is expected. You could also try to execute both calls via curl to know for sure what is expected by the Github API. Please report back with your findings.

EnricoMi:
Btw., #2641.

vanya20074:
I have tested calls via curl with upper and lowercase.
With lower is ok - 201 created
With uppercase - 422Unprocessable Entity
Also asked in github for [clarification](https://github.com/github/docs/issues/26844), but I think it's only mistake in documentation

EnricoMi:
I have commented on that issue, waiting for confirmation, but I think we have to fix those assertions then.

Would you mind open a PR?

vanya20074:
Yes, sure, please take a look https://github.com/PyGithub/PyGithub/pull/2642

EnricoMi:
Fixed via #2642

PR #2642, opened 2023-07-18
Fix assertion for subject_type in creating pull request review comment

vanya20074:
Fix assertion for subject_type 
https://github.com/PyGithub/PyGithub/issues/2640
https://github.com/github/docs/issues/26844


EnricoMi:
Can we also fix `side` and `start_side` while we are touching this?

vanya20074:
> Can we also fix `side` and `start_side` while we are touching this?

I have tested lowercase value for `side` and `start_side` and got this error in response:

`right is not a member of [\"LEFT\", \"RIGHT\"]`

It can be concluded that in this case the values in the documentation are indicated correctly and must be in uppercase.

EnricoMi:
In that case, we should relax the assertion to

    subject_type.lower() in ["line", "file", "side"]

so that both upper and lowercase is accepted by PyGithub and Github response tells users which is right.

I think they will have to harmonize those consts to either always lowercase or upper case them, and in order not to break user code, they will go for case-insensitive consts.

vanya20074:
> In that case, we should relax the assertion to
> 
> ```
> subject_type.lower() in ["line", "file", "side"]
> ```
Done

EnricoMi:
I meant relax the assertion for all string consts (`side`, `start_side`), so that our code looks consistent in some sense.

github/PullRequest.py
```diff
@@ -447,9 +447,9 @@ def create_review_comment(
             "RIGHT",
         ], start_side
         assert in_reply_to is github.GithubObject.NotSet or isinstance(in_reply_to, int), in_reply_to
-        assert subject_type is github.GithubObject.NotSet or subject_type in [
-            "LINE",
-            "FILE",
+        assert subject_type is github.GithubObject.NotSet or subject_type.lower() in [
```
EnricoMi, line 450: On a second thought, lets remove this `lower` so we encode the current known API requirement:
```suggestion
        assert subject_type is github.GithubObject.NotSet or subject_type in [
```
Once docs are updated, we can fix all string assertions.

PR #2622, opened 2023-07-13
CI: label PRs that have conflicts

Borda:
This will simplify users to navigate what is needed

cc: @EnricoMi

PR #2654, opened 2023-07-18
CI: update labels action name and version

Borda:
resolves https://github.com/PyGithub/PyGithub/pull/2622#issuecomment-1640568285

Cc @EnricoMi

### DIFF
--- a/.github/workflows/label-conflicts.yml
+++ b/.github/workflows/label-conflicts.yml
@@ -1,3 +1,5 @@
+name: Label conflicts
+
 on:
   push: { branches: ["main"] }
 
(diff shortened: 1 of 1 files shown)


EnricoMi (approved):
LGTM!

EnricoMi:
The action failed again, a rerun fixed it. Shall we increase the wait to 60s and retry to 10? The action is fine to run for 10 minutes, as long as it succeeds.

Then we should add (at workflow level):

    concurrency:
      group: ${{ github.ref }}
      cancel-in-progress: true

So that concurrent runs are killed and only the action on master HEAD continues.

https://docs.github.com/en/actions/using-jobs/using-concurrency

EnricoMi:
Removed in #2669.

PR #2658, opened 2023-07-18
Merge `File` type stub back to source

trim21:


github/File.py
```diff
@@ -29,110 +29,76 @@
...
 
     @property
-    def status(self):
-        """
-        :type: string
-        """
+    def status(self) -> str:
         return self._status.value
 
-    def _initAttributes(self) -> None:
```
Borda, line 123: Why is this moved? let's keep the dif minimal for easier review...

trim21: EnricoMi think this is better

EnricoMi: @borda you are right, minimal diffs are better in general. In this particular case, having all attributes and types defined at the top of the class makes it more comprehensible, so while we are touching all files anyway, we are moving this method up.

EnricoMi (approved):
LGTM!

PR #2660, opened 2023-07-18
Merge `Event` type stub back to source

trim21:


github/Event.py
```diff
@@ -27,90 +27,69 @@
...
-        """
+    def org(self) -> github.Organization.Organization:
         return self._org.value
 
     @property
-    def payload(self):
-        """
-        :type: dict
-        """
+    def payload(self) -> dict[str, Any]:
```
Borda, line 77:
```suggestion
    def payload(self) -> Dict[str, Any]:
```

trim21: no, `from __future__ import annotations`

EnricoMi (approved):
LGTM!

PR #2693, opened 2023-07-20
fix: allow app installation to use custom certificate authority file

goodtune:
Pass the `verify` value from the requester, or it will default to `True`.

<details>
<summary>Minimal code example and explanation</summary>

Assume I have a `ca.pem` for my company, and it won't be able to verify `api.github.com`

```python
auth = Auth.AppAuth(
    app_id,
    private_key,
).get_installation_auth(installation_id)

Github(auth=auth, verify="ca.pem")

# expect a blow up
print(auth.token)
```

It does not blow up, because at the time of evaluation, `verify=True`, and the default CA (which can validate `api.github.com`) is used instead. Why?

It turns out that `verify` was not being passed into the `GithubIntegration` (a couple of layers down in the call stack).
</details>

### DIFF
--- a/github/Auth.py
+++ b/github/Auth.py
@@ -256,6 +256,7 @@ class AppInstallationAuth(Auth, WithRequester["AppInstallationAuth"]):
         self.__integration = GithubIntegration(
             auth=self._app_auth,
             base_url=requester.base_url,
+            verify=self._WithRequester__requester._Requester__verify,
         )
 
         return self


goodtune:
Although this works, the linting error made me investigate further (I realise it was already hacky) and found that I can work with the `GithubIntegration` directly.

```python
auth = Auth.AppAuth(app_id, private_key)
app = GithubIntegration(auth=auth, verify="ca.pem")
authz = app.get_access_token(installation_id)
print(authz.token)
```

I still think the `verify` should be properly passed on, but I don't have the time to solve that at the moment. I will write up an issue, in case anyone else encounters it.

EnricoMi:
> I still think the `verify` should be properly passed on

What you probably want is

    gh = app.get_github_for_installation(installation_id)

The returned `github.Github` instance has a requester identically configured as that of `app` and `auth`, except that it is authenticated as the app installation. No need to call `get_access_token`.

EnricoMi:
From your pull request, I can see the issue. I will create a fix.

PR #2702, opened 2023-07-28
Add support for global advisories and unify some shared logic with repository advisories

crimsonknave:
This PR adds support for the new Global Advisories api: https://docs.github.com/en/rest/security-advisories/global-advisories. It adds the `get_global_advisory` and `get_global_advisories` methods.

It creates a new `GlobalAdvisory` class  and pulls the shared fields into a class both Global and Repository Advisories inherit from. The payloads of the Global and Repository Advisories are similar, but not quite the same. Repository Advisories have fields designed to allow users to submit using the payloads that are returned. Global Advisories have additional fields that do not apply to advisories attached to a repository. I also implemented a few more fields that are present in both, but were not included (CVSS and identifiers).

I have attempted to follow the code style in the other files, happy to update anything where I'm not following convention. 

-----

The fact that I renamed some of the files means that the header script is removing the copyright from anyone who was listed in the original file, which seems wrong. But, I'm not sure what the right fix is, any manual editing would be overwritten by the next person to run the script.

Side note: `script/fix_headers.py` adds `# -*- coding: utf-8 -*-` which is then removed by the `pyupgrade` step of the pre-commit hooks.


tests/Github_.py
```diff
@@ -1,16 +1,15 @@
 ############################ Copyrights and license ############################
 #                                                                              #
-# Copyright 2012 Vincent Jacques <vincent@vincent-jacques.net>                 #
```
EnricoMi, line 3: removing these lines does not look right

github/AdvisoryCredit.py
```diff
@@ -1,6 +1,6 @@
 ############################ Copyrights and license ############################
 #                                                                              #
-# Copyright 2023 Jonathan Leitschuh <Jonathan.Leitschuh@gmail.com>             #
```
EnricoMi, line 3: removing this line does not look right either

github/CVSS.py
```diff
@@ -0,0 +1,57 @@
...
+    This class represents a CVSS.
+    The reference can be found here <https://docs.github.com/en/rest/security-advisories/global-advisories>
+    """
+
+    def _initAttributes(self) -> None:
+        self._vector_string: Attribute[str] = NotSet
+        self._score: Attribute[Decimal] = NotSet
+
+    @property
+    def score(self) -> Decimal:
```
EnricoMi, line 40: Why not float?

EnricoMi (changes requested):

tests/Github_.py
crimsonknave: Yup, that was done by the automated script. I can manually fix it, but as I noted in the description anyone that runs the script will just revert it. It looks like this file was renamed a few years ago and the script does not follow the rename backwards, so it only sees anyone who touched the new file.

github/AdvisoryCredit.py
crimsonknave: Same issue here, this file was renamed so the script removed it.

github/CVSS.py
crimsonknave: Fundamentally the score is a user facing number so having it be a decimal means that it will always look nice. There won't be a CVSS score that needs more than 1-2 decimal places, so we don't need any of the float precision.

Additionally, decimal will ensure that we don't end up with 5.00000000000001 (or whatever) when doing any math or converting with a score. It also means I can be confident that any comparisons I do against the score won't be impacted by the same issue (so, if I check if the score is 5 less for example, the above float value.

The [decimal docs](https://docs.python.org/3/library/decimal.html) go into more detail into the pros and cons.

crimsonknave:
Happy to make whatever changes y'all would like. I just wasn't sure what the right answer was around the comments.

crimsonknave:
@EnricoMi I'm happy to resolve the copyright comments however you think is best. Aside from that and the merge conflicts, is there anything else I should do to get this PR ready to merge?

tests/Github_.py
EnricoMi: Fixed in #2728, you can ignore this.

github/CVSS.py
EnricoMi: Sounds good, thanks!

EnricoMi:
@JLLeitschuh are you happy with this refactoring of your advisories?

EnricoMi:
@crimsonknave I am happy with this work, please resolve the conflicts and we are ready to go ahead (pending @JLLeitschuh approval).

tests/Github_.py
crimsonknave: Ignore this as in "these changes are fine to commit as is" or, "revert and just add myself and they'll be updated once that PR is merged"?

EnricoMi: Fine to ignore, they will be restored in #2801.

JLLeitschuh (approved):
This looks good to me! Happy to merge!

JLLeitschuh:
@EnricoMi can I merge this?

JLLeitschuh:
@EnricoMi friendly ping!

EnricoMi (approved):
LGTM!

Issue #1700, opened 2020-09-21
403 error for otherwise functional token


seanbreckenridge:
The following fails on my machine:

```python
import json
from github import Github

with open("./token.json", "r") as f:
    auth_token = json.load(f)["token"]

g = Github(auth_token)

repo = next(iter(g.get_user().get_repos()))
# this fails with a 403
print(repo.get_views_traffic())
```

```
{'message': 'Must have push access to repository', 'documentation_url': 'https://docs.github.com/rest/reference/repos#get-repository-clones'}
Traceback (most recent call last):
  File "./requests_fails.py", line 13, in <module>
    r.raise_for_status()
  File "/home/sean/.local/share/virtualenvs/pygithub_requests_error-nXm25YLC-/home/sean/.local/share/pyenv/shims/python/lib/python3.8/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.github.com/repos/seanbreckenridge/albums/traffic/clones
```

The token has the `repo` scope, and if I request the same using `curl`, it works fine.

![](https://camo.githubusercontent.com/7eff8b96b08c3576dd21f49fbb215c57f719f80b/68747470733a2f2f692e696d6775722e636f6d2f4f4d46724e6e4b2e706e67)

This was working fine at one point, and then just stopped. I originally found this while using <https://github.com/karlicoss/ghexport/issues/4>, you can see the process of debugging it in that issue.

I've debugged this quite a bit in the above issue, created <https://github.com/seanbreckenridge/pygithub_requests_error> which has code that runs in a virtualenv which causes the error *on my machine*. Strangely, @karlicoss, who I debugged this with doesn't have the same error.

See the [log file with errrors](https://github.com/seanbreckenridge/pygithub_requests_error/blob/master/log.txt), for [this script](https://github.com/seanbreckenridge/pygithub_requests_error/blob/master/run), which runs:

- a basic `curl` with the token, which succeeds ([link](https://github.com/seanbreckenridge/pygithub_requests_error/blob/master/curl_test.sh))
-  `PyGithub` which fails ([link](https://github.com/seanbreckenridge/pygithub_requests_error/blob/master/pygithub_requests_fails.py))
- `requests` which fails ([link](https://github.com/seanbreckenridge/pygithub_requests_error/blob/master/requests_fails.py))
[...]

Issue #1150, opened 2019-06-21
Fetching a private repo fails (404)


ddevault:
```python
>>> from github import Github
>>> github = Github("oauth token")
>>> github.get_repo("ddevault/test-repo") # GitHubException (404)
```

A similar request with curl succeeds:

```
$ curl -v -A 'PyGitHub/Python' \
    -H Accept:'application/vnd.github.v3+json' \
    -H Authorization:'token [redacted]' \
    https://api.github.com/repos/ddevault/test-repo # works
```

When trying to direct pygithub through mitmproxy to gather more information, it succeeds (200):

```
$ mitmproxy -m reverse:https:/api.github.com
```

```python
>>> from github import Github
>>> github = Github("oauth token", base_url="http://localhost:8080")
>>> github.get_repo("ddevault/test-repo") # works
```

ddevault:
Got it, it was reading my ~/.netrc and bunging things up. It probably should not do that.

PR #2703, opened 2023-07-28
Removing use of .netrc from default behaviour. 

DanielNaaykens:
While attempting to use PyGitHub, I noticed that it would ignore the given credentials and always used my machine's .netrc credentials.

This PR adds an assert to ensure that at least one login method is provided, and ensures that the provided login method is used (via the method proposed in PR #910).

If using a .netrc file is desired as a fallback, the assert could be removed. If that is the case, a flag could be added to the constructor or the documentation updated to explain this.

## Expected Behaviour:

```
User@Computer:~$ python3
Python 3.11
>>> from github import Auth, Github
>>> gh = Github()
>>> print("Using user account {}".format(gh.get_user().login))
    (...)
  File "/home/user/.local/lib/python3.11/site-packages/github/Requester.py", line 487, in __check
    raise self.__createException(status, responseHeaders, data)
github.GithubException.GithubException: 401 {"message": "Requires authentication", "documentation_url": "https://docs.github.com/rest/users/users#get-the-authenticated-user"}
>>>
```

## Actual Behaviour:

```
User@Computer:~$ python3
Python 3.11
>>> from github import Auth, Github
>>> gh = Github()
>>> print("Using user account {}".format(gh.get_user().login))
Using user account DanielNaaykens
>>>
```

## New Behaviour:

```
User@Computer:~$ python3
Python 3.11
>>> from github import Auth, Github
>>> gh = Github()
>>> print("Using user account {}".format(gh.get_user().login))
    (...)
  File "/home/user/.local/lib/python3.11/site-packages/github/MainClass.py", line 139, in __init__
    assert (
AssertionError: At least one authorization method is required.
>>>
```

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -141,6 +141,16 @@ class Github:
         assert seconds_between_writes is None or seconds_between_writes >= 0
         assert auth is None or isinstance(auth, Auth.Auth), auth
 
+        # The 'requests' package can attempt to use a .netrc or other method to log in to github.
+        # This check minimizes user error where no login credentials are provided.
+        assert (
+            login_or_token is not None
+            or password is not None
+            or jwt is not None
+            or app_auth is not None
+            or auth is not None
+        ), "At least one authorization method is required."
+
         if password is not None:
             warnings.warn(
                 "Arguments login_or_token and password are deprecated, please use "
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -118,6 +118,7 @@ class HTTPSRequestsConnectionClass:
         self.timeout = timeout
         self.verify = kwargs.get("verify", True)
         self.session = requests.Session()
+        self.session.auth = self.__authorization_config
 
         if retry is None:
             self.retry = requests.adapters.DEFAULT_RETRIES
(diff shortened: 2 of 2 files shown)


github/MainClass.py
```diff
@@ -141,6 +141,16 @@ def __init__(
         assert seconds_between_writes is None or seconds_between_writes >= 0
         assert auth is None or isinstance(auth, Auth.Auth), auth
 
+        # The 'requests' package can attempt to use a .netrc or other method to log in to github.
```
EnricoMi, line 144: There are endpoints that do not require any authentication.

EnricoMi:
This .netrc magic should really be controllable, I agree. We have seen many issues with it, that were hard to debug and understand: #1753, #1700, #1150.

Can we somehow integrate the .netrc authentication into `github.auth.Auth`, e.g. by implementing `github.auth.NetRc`?

    auth = github.auth.NetRc()
    g = github.Github(auth=auth)

This explicitly uses `.netrc` for authentication. There should be no automatic use any more (breaking change).

EnricoMi (changes requested):

EnricoMi:
@DanielNaaykens thanks for pointing me to the solution, fixed in #2739.

PR #2711, opened 2023-08-10
Search Workflows by Name

armartirosyan:
The existing `get_workflow()` method only works with the input of a workflow ID or workflow filename. This pull request (PR) serves to elevate the method's functionality by enabling the utilization of workflow names as an alternative approach to identifying the relevant workflow.

```
>>> repo.get_workflow(2008038)
Workflow(url="https://api.github.com/repos/PyGithub/PyGithub/actions/workflows/2008038", name="Publish to PyPI")
>>> repo.get_workflow("2008038")
Workflow(url="https://api.github.com/repos/PyGithub/PyGithub/actions/workflows/2008038", name="Publish to PyPI")
>>> repo.get_workflow("Publish to PyPI")
Workflow(url="https://api.github.com/repos/PyGithub/PyGithub/actions/workflows/2008038", name="Publish to PyPI")
```


github/Repository.py
```diff
@@ -3215,6 +3216,14 @@ def get_workflow(self, id_or_name):
         :rtype: :class:`github.Workflow.Workflow`
         """
         assert isinstance(id_or_name, int) or isinstance(id_or_name, str), id_or_name
+        if (
+            type(id_or_name).__name__ == "str"
```
trim21, line 3220: why not `isinstance`?

github/Repository.py
```diff
@@ -3215,6 +3216,14 @@ def get_workflow(self, id_or_name):
...
+            and not id_or_name.isnumeric()
+            and (".yml" not in id_or_name and ".yaml" not in id_or_name)
+        ):
+            for workflow in self.get_workflows():
```
EnricoMi, line 3224: Multiple workflows can have the same name, this finds only the first.

To be honest, I don't like this magic here. `id_or_name` is a unique identifier (the int id or the string file name): https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#get-a-workflow

Search for a workflow that matches the name is 1) something that can be easily done in user code 2) should be a different method like `search_workflow`, which I don't see much benefit in adding to the overcrowded `Repository` class.

github/Repository.py
armartirosyan: Because I was very tired and not sure what I was thinking :) 
Thanks for catching it!

github/Repository.py
armartirosyan: Thanks for your feedback @EnricoMi!

My two cents:
It is possible but very unlikely that two or more workflows will be named the same (at least, I have not encountered such case). The reason I created this code is that I was searching the workflow name and was getting a 404 error, and I could not understand why this is. The method's parameter expects an id or name (`id_or_name`), and I was intuitively providing a name. 

In my opinion,  it is an additional capability that fits perfectly into the `get_worklfow()` method.

EnricoMi:
In that case, the parameter `id_or_name` should better be renamed to `id_or_file_name` to avoid the impression the workflow name can be used to get the workflow. Also, the method doc string could emphasise this.

As I said, finding a workflow by name can easily done in user code, which then can handle duplicates (whether they may exist or not).

armartirosyan:
> In that case, the parameter `id_or_name` should better be renamed to `id_or_file_name` to avoid the impression the workflow name can be used to get the workflow. Also, the method doc string could emphasise this.
> 
> As I said, finding a workflow by name can easily done in user code, which then can handle duplicates (whether they may exist or not).

Removed the logic, renamed the parameter, and updated the docstring.
Thanks!

github/Repository.py
```diff
@@ -3208,23 +3208,15 @@ def get_workflows(self):
...
-        if (
-            isinstance(id_or_name, str)
-            and not id_or_name.isnumeric()
-            and (".yml" not in id_or_name and ".yaml" not in id_or_name)
-        ):
-            for workflow in self.get_workflows():
-                if workflow.name == id_or_name:
-                    return workflow
-        headers, data = self._requester.requestJsonAndCheck("GET", f"{self.url}/actions/workflows/{id_or_name}")
+        assert isinstance(id_or_file_name, int) or isinstance(id_or_file_name, str), id_or_file_name
```
trim21, line 3218:
```suggestion
        assert isinstance(id_or_file_name, (int, str)), id_or_file_name
```

trim21:
so this PR doesn't actually change anything， just update argument name and add documents…

armartirosyan:
> so this PR doesn't actually change anything， just update argument name and add documents…

Appears that way! Adds a test case for `Repository.get_workflow()`

EnricoMi (approved):
LGTM!

PR #2712, opened 2023-08-10
Add `include_all_branches` field to create repo from template function

nckslvrmn:
Adds the boolean `include_all_branches` to the request object being sent in the create from template request. This has been in the API spec for a while.

I was unable to get a test account to record the new outputs of the tests as a lot of the organization tests are associated with an organization I cant see.


nckslvrmn:
As expected, tests failed in the tests I updated, likely because I wasn't able to record the new outputs. Can someone assist with that?

PR #2715, opened 2023-08-11
fix: required_conversation_resolution assertion

### LINKED ISSUE #2714: Error when setting `required_conversation_resolution`
After setting `required_conversation_resolution=True` in `edit_protection`, one gets an error
```
     required_conversation_resolution is github.GitHubObject.NotSet
AsertionError: True
```

terenho:
`required_conversation_resolution` assertion checks if it is boolean, instead of checking `required_conversation_resolution` is boolean.

fixes #2714


EnricoMi (approved):
LGTM, but I doubt this fixes / caused #2714.

PR #2720, opened 2023-08-15
add is_alphanumeric attribute to autolink

ofrankg:
GitHub API contains an is_alphanumeric attribute that is missing. The attribute has been added using the script and tested by editing the appropriate files in tests folder


EnricoMi:
For reference, there is another PR which is still in draft state: #2630 @janssonoskar

EnricoMi:
Fixed in #2630.

Issue #2436, opened 2023-02-23
GithubObjects are no longer serializable with `pickle`


josh-fell:
To support data exchange, being able to serialize non-JSON-serializable `GithubObject`s has been easily achieved in past with `pickle`. However, starting with v1.58, `GithubObject`s like `Repository`, `NamedUser`, etc. are no longer serializable with with `pickle`,

For example,
```py
import github
import pickle

repo = github.Github().get_repo(full_name_or_id="PyGithub/PyGithub")
pickle.dumps(repo)
```
raise the following exception:
```
Traceback (most recent call last):
  File "/Users/joshfell/code/daedalus/tempCodeRunnerFile.python", line 15, in <module>
    pickle.dumps(repo)
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/multiprocessing/synchronize.py", line 101, in __getstate__
    context.assert_spawning(self)
  File "/usr/local/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/multiprocessing/context.py", line 373, in assert_spawning
    raise RuntimeError(
RuntimeError: RLock objects should only be shared between processes through inheritance
```

The underyling `Requester` object now has an `RLock()` attribute, `__auth_lock`, which looks to be preventing this serialization.

EnricoMi:
Pickle allows `GithubObject` to exclude the lock from being serialized:
https://docs.python.org/3/library/pickle.html#handling-stateful-objects

And we should add tests that proof pickle support.

josh-fell:
@EnricoMi, Thanks for taking a look so quickly and the tip on the `pickle` workaround!

PR #2724, opened 2023-08-17
Close connections after use

### LINKED ISSUE #1859: GitHub Class Doesn't Clean-up Connections After Use
The following code in unit test throws ResourceWarning for open connections and I cannot see a method to use to close these down. Am I missing something, or is this a possible memory/connection leak oversight?

```
repository = "someuser\somerepository"
connection = Github (self.token)
connection.get_repo (repository).get_latest_release ().get_assets () [0]
del connection
```

hemslo:
Currently connections are closed once `Github` instance is GC, this change introduces a way to explicitly close connections and with a context manager for MainClass. Also makes `Requester` thread safe when creating new connections concurrently.

Usages:

```python
   with Github(auth=auth) as g:
     # do something
```

or

```python
    g = Github(auth=auth)
    # do something
    g.close()
```

Fix #1859


EnricoMi:
We would need to do the same for `GithubIntegration`, the "other main class" of this project.

The `g.close()` should be added at the end of the example in https://github.com/PyGithub/PyGithub/blob/main/README.md and https://github.com/PyGithub/PyGithub/blob/main/doc/introduction.rst. The `MainClass.close` could use a docstring showing the `with` example as an alternative to calling `close`.

Last  time we introduced a lock we broke thinks for users serializing `GithubObjects`: https://github.com/PyGithub/PyGithub/issues/2436

(We should really add tests for serializable `GithubObjects`).

hemslo:
@EnricoMi `pickle` is already broken since there is a `lambda` in `Requester`, fixed in https://github.com/PyGithub/PyGithub/pull/2724/commits/3b57f2cd8e706f01dcc47e88959c4745371b51f8 together.

doc/introduction.rst
```diff
@@ -35,6 +35,9 @@ Then play with your Github objects::
         # to see all the available attributes and methods
         print(dir(repo))
 
+To close connections after use::
+
+    g.close()
 Download and install
```
EnricoMi, line 41:
```suggestion

Download and install
```

github/Requester.py
```diff
@@ -357,14 +370,17 @@ def __init__(
...
         self.__scheme = o.scheme
         if o.scheme == "https":
             self.__connectionClass = self.__httpsConnectionClass
         elif o.scheme == "http":
             self.__connectionClass = self.__httpConnectionClass
         else:
             assert False, "Unknown URL scheme"
+        self.__connection: Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = None
+        self.__connection_lock = threading.Lock()
+        self.__custom_connections: Deque[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = deque()
```
EnricoMi, line 383: Do you see problems that we pickle these connection classes including their sessions?

tests/Requester.py
```diff
@@ -132,6 +132,37 @@ class TestAuth(github.Auth.AppAuth):
...
+        self.assertIsNone(requester._Requester__connection)
+
+    def testCloseGithubIntegration(self):
+        mocked_connection = mock.MagicMock()
+        mocked_custom_connection = mock.MagicMock()
+
+        class TestAuth(github.Auth.AppAuth):
+            pass
+
+        auth = TestAuth(123, "key")
```
EnricoMi, line 155: Why not using `github.Auth.AppAuth` directly here?

EnricoMi (commented):
Looks good, some minor comments.

tests/Pickle.py
```diff
@@ -0,0 +1,14 @@
...
+
+REPO_NAME = "PyGithub/PyGithub"
+
+
+class TestPickle(unittest.TestCase):
+    def test_pickle_github(self):
+        gh = github.Github()
+        gh2 = pickle.loads(pickle.dumps(gh))
+        self.assertIsInstance(gh2, github.Github)
+        self.assertIsNotNone(gh2._Github__requester._Requester__connection_lock)
```
EnricoMi, line 14: Can we test pickling a `GithubObject` as well?
```suggestion
        self.assertIsNotNone(gh2._Github__requester._Requester__connection_lock)

    def test_pickle_repository(self):
        gh = github.Github()
        repo = gh.get_repo(REPO_NAME)

        repo2 = pickle.loads(pickle.dumps(repo))
        self.assertIsInstance(repo2, Repository)
        self.assertIsNotNone(repo2._requester._Requester__connection_lock)
```
You can reuse replay data from `Requester.testBaseUrlPrefixRedirection.txt`:

    tail -n 10 tests/ReplayData/Requester.testBaseUrlPrefixRedirection.txt > tests/ReplayData/Pickle.test_pickle_repository.txt

doc/introduction.rst
hemslo: Fixed in https://github.com/PyGithub/PyGithub/pull/2724/commits/63956924eaab9ab233bff9b6503c2275d85596cb

tests/Pickle.py
hemslo: added in https://github.com/PyGithub/PyGithub/pull/2724/commits/63956924eaab9ab233bff9b6503c2275d85596cb

github/Requester.py
hemslo: `__connection` was pickled previously without issues, I tested pickle `__custom_connections` by append `self.__connection` to it manually, `pickle.loads` and `pickle.dumps` are all good.

EnricoMi: I am more thinking about what does it mean to pickle these connection instances and reuse them remotely.

hemslo: right, if reuse them remotely, then those connections can't be reused. If it's used for save and restore locally, then maybe those connections are still valid. To support both use cases, I think it's ok to ignore them during pickle. I'll push a new change soon.

hemslo: Ignored in https://github.com/PyGithub/PyGithub/pull/2724/commits/7e17681d87775ecd4b38760487f646ec928eef5f, is there a good test case I can use for testing `__custom_connections`?

hemslo: Found an issue with previous test, use `unittest.TestCase` won't use replay data, use `Framework.TestCase` will cause pickle error because of lambda in test [setup](https://github.com/PyGithub/PyGithub/blob/48d2009c8141e3e32def3db35ecf7d91ce1e687a/tests/Framework.py#L280).
So to avoid external requests in test, I switched to use `lazy=True` in pickle tests.
see https://github.com/PyGithub/PyGithub/pull/2724/commits/49d27405f3e65c877d8d5a9a9ed13d1c203c66e5

EnricoMi: The `testPickleRepository` test works just fine for me without `lazy=True`.

hemslo: if you turn off your network, that test will fail as it will send real http request to github

hemslo: Another way to verify is to update the replay data to return `404`, the test can still pass without `lazy=True`, because the file is not being used.
Since we don't expect unitests to send real http requests out, and `Framework.TestCase` is not pickable, I updated to be `lazy=True` and removed replay data.

EnricoMi: Right, testing with `lazy=True` is not ideal, but I agree, test setup is not made for this.

EnricoMi (approved):
LGTM!

PR #2726, opened 2023-08-18
Print usage when no arguments given to add_attribute script

EnricoMi:


scripts/add_attribute.py
```diff
@@ -38,6 +38,14 @@
 import os.path
 import sys
 
+if len(sys.argv) == 1:
+    print(f"usage: {sys.argv[0]} <class> <attribute> <attribute-type> [<attribute-class-type>]")
+    print()
+    print("examples:")
+    print(f"  {sys.executable} {sys.argv[0]} Repository name string")
```
trim21, line 45: Just python maybe?no need to use full executable path

PR #2731, opened 2023-08-18
Fix the Pip install command

Skptak:
On the front page of the [README Install](https://github.com/PyGithub/PyGithub#install) Instructions there is a button to copy  `$ pip install PyGithub` command to the clipboard
Due to the leading `$` this command fails when pasted into a terminal
Removing the leading `$` from the command so it can be copy pasted

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This library enables you to manage [GitHub] resources such as repositories, user
 ## Install
 
 ```bash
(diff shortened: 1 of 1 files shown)


A-Zaba (approved):

EnricoMi (approved):
Right, this might be confusing. Let's be safe.

PR #2732, opened 2023-08-20
feat: add per_page and page to Repository.get_pulls

psyclaudeZ:
It looks like this specific API isn't in sync with [what's available](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests) right now, hence this PR to patch it up.

Tested in the repl:

```
>>> from github import Auth, Github
>>> import os
>>> key = os.environ.get("GITHUB_ACCESS_TOKEN")
>>> auth = Auth.Token(key)
>>> g = Github(auth=auth)
>>> repo = g.get_repo("PyGithub/PyGithub")
>>> res = repo.get_pulls(state="all", per_page=100, page=2)
>>> for p in res:
...     print(p)
...
PullRequest(title="Fix the Pip install command", number=2731)
PullRequest(title="Add attribute add attribute lists", number=2729)
PullRequest(title="Improve fix headers", number=2728)
PullRequest(title="Add import for new class attribute", number=2727)
PullRequest(title="Print usage when no arguments given to add_attribute script", number=2726)

```

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -2932,6 +2932,8 @@ class Repository(github.GithubObject.CompletableGithubObject):
         direction=github.GithubObject.NotSet,
         base=github.GithubObject.NotSet,
         head=github.GithubObject.NotSet,
+        per_page=github.GithubObject.NotSet,
+        page=github.GithubObject.NotSet,
     ):
         """
         :calls: `GET /repos/{owner}/{repo}/pulls <https://docs.github.com/en/rest/reference/pulls>`_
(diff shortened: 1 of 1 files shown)


EnricoMi:
Any endpoint that returns a paginated list takes arguments `per_page` and `page`. Therefore, they are implemented once for all paginated lists and not for each endpoint individually.

The `per_page` argument is set when constructing `github.Github`, the `page` is used to iterate over the pages. A specific page can be accessed vie `PaginatedList.get_page(page)`.

Above example then reads:
```python
>>> from github import Auth, Github
>>> import os
>>> key = os.environ.get("GITHUB_ACCESS_TOKEN")
>>> auth = Auth.Token(key)
>>> g = Github(auth=auth, per_page=100)
>>> repo = g.get_repo("PyGithub/PyGithub")
>>> res = repo.get_pulls(state="all").get_page(2)
...
```

PR #2734, opened 2023-08-21
Remove un-used lambda from Requester that prevents serialization

EnricoMi:
Looks like this lambda is not used but breaks serialization. Thanks to @hemslo for pointing this out.

Unit tests will be added in #2724.


EnricoMi:
Fixed by #2724, won't be backported so this fix is not needed.

PR #2746, opened 2023-09-04
Add capability for global laziness

### LINKED ISSUE #2334: [Feature Request] Allow setting default value for lazy for all API calls
As I'm already familiar with lazy objects & workflows, I would like it to have `lazy==True` to be the default. But because this may be not the best for everyone, I propose adding a default lazy argument to the initial `Github` class constructor which becomes the default for every method called by it. If I gain some time, I will look into implementing it myself. However if someone else is faster to implement it, feel free to implement & submit it yourself.

This may be implemented by adding a variable to `Github` which holds the global default, can be set by using the constructor and defaults to `False` (to keep the current behavior the default). Then every method supporting lazy behavior may either look up this global default by itself or a decorator is implemented which replaces the lazy parameter by itself. The default for the method's argument should be set to `None`. Then the following comparison code can be used to resolve the value to use:
```python
lazy_global_default if lazy_method_arg is None else lazy_method_arg
```

(This issue serves as a reminder for myself.)

PS: This library looks really good & is easy to use, thanks for the work.

EnricoMi:
Adds capability to configure global laziness Github objects through `Github(lazy=…)` and `GithubIntegration(lazy=…)`. Does not add any laziness capability to any particular Github object.

Reworks `CompletableGithubObject` to support following features:
- It fetches initial complete JSON, which allows to replace
```python
headers, data = self.__requester.requestJsonAndCheck("GET", f"/users/{login}")
return github.NamedUser.NamedUser(self.__requester, headers, data, completed=True)
```
with
```python
return github.NamedUser.NamedUser(self.__requester, url=f"/users/{login}")
```
Which allows that returned `CompletableGithubObject` to be lazy.

Removes any code related to `CompletableGithubObject` from `NonCompletableGithubObject`.
- Which requires replacing (for `NonCompletableGithubObject` only)
```python
        return github.Branch.Branch(self._requester, headers, data, completed=True)
```
for `NonCompletableGithubObject` with
```python
        return github.Branch.Branch(self._requester, headers, data)
```

Fixes #2334


EnricoMi:
@trim21 what do think about this?

trim21:
good idea

EnricoMi:
@trim21 can you help me with this one?

    Incompatible return value type (got "CompletableGithubObject", expected "TGithubObject")

https://github.com/PyGithub/PyGithub/pull/2746/files#diff-cd558fc3c3c4b4ddfcbb8e34bcd6a3fa37d85cb07d510f46efa74f0b8c24fdfeR777

trim21:
> @trim21 can you help me with this one?
> 
> ```
> Incompatible return value type (got "CompletableGithubObject", expected "TGithubObject")
> ```
> 
> https://github.com/PyGithub/PyGithub/pull/2746/files#diff-cd558fc3c3c4b4ddfcbb8e34bcd6a3fa37d85cb07d510f46efa74f0b8c24fdfeR777

just use `# type: ignore`... mypy can't narrow type in this case.

EnricoMi:
> just use `# type: ignore`... mypy can't narrow type in this case.

Thanks, done

github/GithubObject.py
```diff
@@ -367,10 +408,27 @@ def _completeIfNeeded(self) -> None:
     def __complete(self) -> None:
         if self._url.value is None:
             raise IncompletableObject(400, message="Returned object contains no URL")
-        headers, data = self._requester.requestJsonAndCheck("GET", self._url.value)
+        headers, data = self._requester.requestJsonAndCheck("GET", self._url.value, headers=self.__completeHeaders)
         self._storeAndUseAttributes(headers, data)
         self.__completed = True
 
+    def do_complete_unless_lazy(self, lazy: Opt[bool]) -> Self:
```
JLLeitschuh, line 415: This could use it's own dedicated pydoc. Also why is `lazy` optional, but not given a default value of `None`

github/MainClass.py
```diff
@@ -769,7 +773,9 @@ def create_from_raw_data(
         if headers is None:
             headers = {}
 
-        return klass(self.__requester, headers, raw_data, completed=True)
+        if issubclass(klass, CompletableGithubObject):
+            return klass(self.__requester, headers, raw_data, completed=True)  # type: ignore
```
JLLeitschuh, line 777: You've been removing `completed` from many constructors, why is it still needed here?

github/Repository.py
```diff
@@ -1993,15 +1993,18 @@ def get_assignees(self) -> PaginatedList[NamedUser]:
         """
         return PaginatedList(github.NamedUser.NamedUser, self._requester, f"{self.url}/assignees", None)
 
-    def get_branch(self, branch: str) -> Branch:
+    def get_branch(self, branch: str, lazy: Opt[bool] = NotSet) -> Branch:
```
JLLeitschuh, line 1996: Why add this additional argument? Can't the end-user decide this by calling the additional method afterwards? Why do they need to specify it here?

JLLeitschuh (commented):
It seems excessive to add the `lazy` argument everywhere across the API, and also prone to making mistakes in the future. Additionally, since it's positional, if additional arguments get added later, it would be an API breaking change to not add those arguments after this new `lazy` argument.

Shouldn't everything be lazy by default, and only get realized when things on that object are accessed?

github/GithubObject.py
```diff
@@ -151,49 +152,51 @@ def __init__(
...
 
     @property
-    def raw_data(self) -> Dict[str, Any]:
+    def raw_data(self) -> dict[str, Any]:
         """
         :type: dict
         """
-        self._completeIfNeeded()
+        if isinstance(self, CompletableGithubObject):
+            self._completeIfNeeded()
```
JLLeitschuh, line 168: Why not do this through override, instead of instance check?

github/MainClass.py
```diff
@@ -205,8 +210,20 @@ def __init__(
...
+        """
+        Create a Github instance with identical configuration but the given lazy setting.
+        :param lazy: completable objects created from this instance are lazy,
+                     as well as completable objects created from those, and so on
+        :return: new Github instance
+        """
+        kwargs = self.__requester.kwargs
+        kwargs.update(lazy=lazy)
+        return Github(**kwargs)
+
```
JLLeitschuh, line 226: Love this!

tests/CommitCombinedStatus.py
```diff
@@ -33,8 +33,10 @@ class CommitCombinedStatus(Framework.TestCase):
     def setUp(self):
         super().setUp()
         self.combined_status = (
-            self.g.get_repo("edx/edx-platform", lazy=True)
+            self.g.withLazy(lazy=True)
+            .get_repo("edx/edx-platform")
             .get_commit("74e70119a23fa3ffb3db19d4590eccfebd72b659")
+            .complete()
```
JLLeitschuh, line 39: Is this actually needed? Wont the next method call cause the object to be completed anyway?

github/GithubObject.py
```diff
@@ -327,25 +326,64 @@ def _initAttributes(self) -> None:
...
+            headers = {}
+        if attributes is None:
+            attributes = {}
+        if url is not None:
+            attributes["url"] = url
+        super().__init__(requester, headers, attributes)
+        self.__completed = completed if isinstance(completed, bool) else False
+        self.__completeHeaders = {"Accept": accept} if accept else None
+
+        # complete this completable object when requester indicates laziness and
```
EnricoMi, line 383:
```suggestion
        # complete this completable object when requester indicates non-laziness and
```

github/GithubObject.py
EnricoMi: As in 4bd755fefbc6125fb721724a94f2f8285db1e960?

github/MainClass.py
EnricoMi: This can probably be removed as well.

tests/CommitCombinedStatus.py
EnricoMi: This is just to keep the recorded test data. I could simply remove `.complete()` (getting the commit), which simplifies the test and proves the laziness works as expected: 8e02440aa9affc689cf23b74f461e9dd9a7ad543

EnricoMi:
@JLLeitschuh is this good to be approved?

PR #2747, opened 2023-09-05
Close replay test data file silently when test is failing already

EnricoMi:
Otherwise closing the test data file may raise an assertion error, which will eclipse the root failure.

Further adds the file name of the failing replay data file to the assertion message.


JLLeitschuh (approved):
LGTM

PR #2748, opened 2023-09-07
Make Branch.edit_* functions return objects

### LINKED ISSUE #2735: Make `Branch.edit_*` return Branch objects
The API returns the `BranchProtection` object, the `Branch.edit_protection` method should return it:
https://github.com/PyGithub/PyGithub/blob/294c0cc9c5a0a5f127dd12ab2cc7ac7ca9c3e59c/github/Branch.py#L272-L277

The same holds for the other `edit_` methods in `Branch`.

luna215:
## Description
Returning BranchProtection for `edit_*` function in `Branch.py`
```python
return github.BranchProtection.BranchProtection(self._requester, headers, data, completed=True)
```

resolves #2735

### DIFF
--- a/github/Branch.py
+++ b/github/Branch.py
@@ -299,7 +299,7 @@ class Branch(NonCompletableGithubObject):
         self,
         strict: Opt[bool] = NotSet,
         contexts: Opt[list[str]] = NotSet,
-    ) -> BranchProtection:
+    ) -> RequiredStatusChecks:
         """
         :calls: `PATCH /repos/{owner}/{repo}/branches/{branch}/protection/required_status_checks <https://docs.github.com/en/rest/reference/repos#branches>`_
         """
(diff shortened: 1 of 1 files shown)


github/Branch.py
```diff
@@ -297,7 +299,7 @@ def edit_required_status_checks(
         self,
         strict: Opt[bool] = NotSet,
         contexts: Opt[list[str]] = NotSet,
-    ) -> None:
+    ) -> BranchProtection:
```
EnricoMi, line 302: It looks like the Github API returns `RequiredStatusChecks` here: https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#delete-branch-protection

github/Branch.py
```diff
@@ -341,7 +345,7 @@ def edit_required_pull_request_reviews(
         dismiss_stale_reviews: Opt[bool] = NotSet,
         require_code_owner_reviews: Opt[bool] = NotSet,
         required_approving_review_count: Opt[int] = NotSet,
-    ) -> None:
+    ) -> BranchProtection:
```
EnricoMi, line 348: `RequiredStatusChecks` here as well.

EnricoMi (changes requested):
Thanks for the contribution, I think return types should be different types, but in general, this is the right approach.

EnricoMi (approved):
LGTM!

PR #2749, opened 2023-09-07
Merge `Repository.pyi` back to source

trim21:
this file is too huge(4271 lines), so I didn't change code unless it will fail mypy.

Also merges overloads of `create_pull`.


EnricoMi:
LGTM! Please go ahead merging the `create_pull` methods.

trim21:
@EnricoMi done

tests/Repository.py
```diff
@@ -1296,11 +1296,11 @@ def testGetSubscribers(self):
...
-            "Pull request created by PyGithub",
-            "Body of the pull request",
-            "topic/RewriteWithGeneratedCode",
-            "BeaverSoftware:master",
-            True,
+            title="Pull request created by PyGithub",
+            body="Body of the pull request",
+            base="topic/RewriteWithGeneratedCode",
+            head="BeaverSoftware:master",
+            maintainer_can_modify=True,
```
EnricoMi, line 1303: with
```suggestion
            draft=False,
            maintainer_can_modify=True,
```
there would be no need to change the replay data, and we would test that attribute

trim21: good idea

PR #2751, opened 2023-09-07
Allow two-dot git diff

siddharthab:
This could be relevant in some contexts.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons


siddharthab:
Looks like this is not supported in the REST API.

PR #2753, opened 2023-09-10
Adding `code_search` to `RateLimit.py`

### LINKED ISSUE #2745: Add `code_search` to `RateLimit`
There is a separate rate limit for code search:
https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28

It is not in the `RateLimit` object:
```
(Pdb) gh.get_rate_limit().code_search
*** AttributeError: 'RateLimit' object has no attribute 'code_search'
```

But it is available in the data:
```
(Pdb) gh.get_rate_limit().raw_data["code_search"]
{'limit': 10, 'used': 0, 'remaining': 10, 'reset': 1693823606}
```

luna215:
We now have `code_search` as a property for RateLimit types.

```python
@property
def code_search(self) -> Rate:
     """
     Rate limit for code-searchig API
        
     :type: class:`github.Rate.Rate`
     """
     return self._core_search
```

resolves #2745


EnricoMi (changes requested):
While you are there, can you add all other missing rate limits as well? https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28#about-rate-limits

luna215:
@EnricoMi the rest of the properties should be added now

PR #2755, opened 2023-09-11
Remove unusable commit input from `Workflow.create_dispatch`

gbhand:
`Workflow.create_dispatch` currently allows `ref` to be a `github.Commit.Commit` type. 

It appears that this option was added when workflow dispatch support was introduced https://github.com/PyGithub/PyGithub/pull/1625#issuecomment-664364599 but [the GitHub API did not end up supporting it](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event):

>The git reference for the workflow. The reference can be a branch or tag name.

I've tested this behavior with PyGithub and manually invoking the API, both showing the dispatch creation fails when a commit is provided. The Github CLI gives a more verbose response confirming that their definition of ref does not support commits:

```shell
$ gh workflow --ref "<a commit that exists in the repo>" run build.yml
could not create workflow dispatch event: HTTP 422: No ref found for: <a commit that exists in the repo> (https://api.github.com/repos/<redacted>/<redacted>/actions/workflows/<redacted>/dispatches)
```

### DIFF
--- a/github/Workflow.py
+++ b/github/Workflow.py
@@ -25,7 +25,6 @@ from datetime import datetime
 from typing import Any
 
 import github.Branch
-import github.Commit
 import github.GithubObject
 import github.NamedUser
 import github.Tag


EnricoMi:
I think the `Commit` import can be removed:
https://github.com/PyGithub/PyGithub/pull/2755/files#diff-b21087810495ec178373d557b34b9f1f64e04d2c74ebcd0905879761f2fdf1c7R28

@trim21 looks like the linting does not detect unused imports, can we add that?

trim21:
> I think the `Commit` import can be removed: https://github.com/PyGithub/PyGithub/pull/2755/files#diff-b21087810495ec178373d557b34b9f1f64e04d2c74ebcd0905879761f2fdf1c7R28
> 
> @trim21 looks like the linting does not detect unused imports, can we add that?

no existing tools support this

EnricoMi:
> no existing tools support this

Well, there is: #2756

trim21:
> > no existing tools support this
> 
> Well, there is: #2756

I tested autoflake and `--remove-all-unused-imports`

trim21:
> I think the `Commit` import can be removed: https://github.com/PyGithub/PyGithub/pull/2755/files#diff-b21087810495ec178373d557b34b9f1f64e04d2c74ebcd0905879761f2fdf1c7R28
> 
> @trim21 looks like the linting does not detect unused imports, can we add that?

ruff can remove unused imports, but neigher ruff nor autoflake can work at this case.

EnricoMi (changes requested):
Can you remove the `import github.Commit` please?

gbhand:
@EnricoMi changes are addressed 😄

PR #2756, opened 2023-09-12
Remove unused imports, variables, expand star imports

EnricoMi:


EnricoMi:
Looks like this does not remove unused `github` imports, though it claims it does:

>   --remove-all-unused-imports: remove all unused imports (not just those from the standard library)

Issue #2760, opened 2023-09-15
PullRequest.is_merged does not handle token expiration


jorge-delphix:
# Context

I am trying to build some automation using PyGithub and a Github App to authenticate to our organization.
The token I can get has an expiration of 1 hour.

The idea of my workflow is to do the following steps:
1. Get a token from Github App
2. Create pull request
3. Notify reviewers
4. Wait for PR to be merged with a flexible timeout to allow reviewers time to review.
5. Once PR is merged, do more steps

The problem arises in 4 if the timeout exceeds 1 hour.

# Problem

In the scenario above I noticed that even after the authentication token has expired, calling `pr.is_merged()` returns `False`.

Looking at the code the behavior I was seeing makes sense:
```
    def is_merged(self) -> bool:
        """
        :calls: `GET /repos/{owner}/{repo}/pulls/{number}/merge <https://docs.github.com/en/rest/reference/pulls>`_
        """
        status, headers, data = self._requester.requestJson("GET", f"{self.url}/merge")
        return status == 204
```
since the status of the request with expired token would be 401.

I see two issues with this:
1. False negatives - a PR could have been merged but `is_merged` returns False. 
2. Consumer is not notified in any way that their authentication has expired.

This also differs from other methods that I have been using like `pr.get_commits().get_page(0)` or `pr.get_labels().get_page(0)` which fail with the expected `github.BadCredentialsException` as soon as the token expires.

I also tested with a token that has no access to my organization, but in that case an expected 403 was raised which seems to be handled by `github.GithubRetry`.

# Steps to Reproduce

1. Create a Github Token on your account with `repo` access
2. Use token to create a client and get a pull request `pr`
3. Remove Github token from your account
4 . Run `pr.is_merged` .

For example:
```
import github
github_client = github.Github("<YOUR TOKEN>")

upstream_repo = github_client.get_repo("PyGithub/PyGithub")
pr = upstream_repo.get_pull(2757)
pr.is_merged()
False
pr.get_commits().get_page(0)
[Commit(sha="7569ac6c33e3682fa42ddf7aee1640198461a557"), Commit(sha="626a0435d063b12f9a9501a282e7371fe8783494")]
#
# Remove token from account
#
pr.is_merged()
False
pr.get_commits().get_page(0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../site-packages/github/PaginatedList.py", line 258, in get_page
[...]

PR #2763, opened 2023-09-19
Add a requester method to explicitly validate status codes

jorge-delphix:
# Problem

Described a specific scenario in detail here: https://github.com/PyGithub/PyGithub/issues/2760

The general problem is when `requestJson` is used to get a status code and then comparing that to an expected
status code via `status == <expected status>`. This pattern does not handle codes that would be returned due to errors like Authentication errors ( 401, 403) , Validation errors (422), Not found (404), etc. Instead of an appropriate error being raised to the user so they can act on it, the error is swallowed and the call returns False.

# Evaluation

From my understanding of the code, the `is_merged` method uses `requestJson` instead of `requestJsonAndCheck` since the latter would raise an `UnknownObjectException` when a pull request is not merged. This is due to the APIs use of 404 to signify a pull request has not been merged. See [here](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#check-if-a-pull-request-has-been-merged) for details.

The other issue is that `requestJsonAndCheck` does not return the `status` so a comparison on that with an expected code is currently not possible. 

This seems to me like a generic approach is needed that returns `status` and raises exceptions but can ignore certain expected 400+ codes.

# Proposed Solution

I see two possible solutions:
1. Modify `requestJsonAndCheck` to return `status` and ignore certain codes like 404 in the is_merged case.
2. Create a new method that returns `status` and validates it's within the expected codes if not it raises an exception.

I opted not to pursue Solution 1 since it is a breaking change and every call to `requestJsonAndCheck` would need to handle the new `status` returned.

Opted for Solution 2 since it can be implemented in a rolling basis in the scenarios that require it and `requestJsonAndCheck` could remain as is.

**NOTE**: I am new to the code base so if there's a better way of solving this issue that I have missed let me know.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -75,6 +75,7 @@ from typing import (
     ItemsView,
     List,
     Optional,
+    Set,
     Tuple,
     Type,
     TypeVar,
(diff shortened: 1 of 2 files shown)


github/Requester.py
```diff
@@ -614,6 +615,22 @@ def __structuredFromJson(self, data: str) -> Any:
...
 
+    def requestJsonAndValidateStatus(
+        self,
+        verb: str,
+        url: str,
+        parameters: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, Any]] = None,
+        input: Optional[Any] = None,
+        cnx: Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = None,
+        valid_codes: Optional[Set[int]] = None,
```
trim21, line 626: This should be required parameter I think.

jorge-delphix: Sure, just modified.

PR #2764, opened 2023-09-21
Add `actor` property to WorkflowRun

gbhand:
The GitHub API [returns an `actor` field when listing workflow runs](https://docs.github.com/en/free-pro-team@latest/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow) but `WorkflowRun` does not include `actor`. 

This PR adds the `actor` property of `NamedUser` type.

Not sure about field ordering conventions so just followed the API response schema.


github/WorkflowRun.py
```diff
@@ -318,6 +325,8 @@ def _useAttributes(self, attributes: dict[str, Any]) -> None:
             self._created_at = self._makeDatetimeAttribute(attributes["created_at"])
         if "updated_at" in attributes:  # pragma no branch
             self._updated_at = self._makeDatetimeAttribute(attributes["updated_at"])
+        if "actor" in attributes:  # pragma no branch
+            self._actor = self._makeClassAttribute(github.NamedUser.NamedUser, attributes["actor"])
```
trim21, line 329: Need to import this class I guess

gbhand: Yeah wasn't too sure what is needed here. I saw later on `github.Repository.Repository` was referenced but only imported by name with `TYPE_CHECKING`.

Would adding `import github.NamedUser` be the right move?

trim21: yes

gbhand: Addressed [841daed](https://github.com/PyGithub/PyGithub/pull/2764/commits/841daed3e18a5db15e9cdbc153ec505fb10d8147)

trim21:
lgtm

EnricoMi (changes requested):
Can we add this new attribute to https://github.com/PyGithub/PyGithub/blob/main/tests/WorkflowRun.py#L40? LGTM besides this.

EnricoMi:
@gbhand?

gbhand:
Apologies @EnricoMi, haven't checked this in quite a while. Added to the unit test in [6347cb2](https://github.com/PyGithub/PyGithub/pull/2764/commits/6347cb221bb28f1a1b4bbb6b12bd0de8e8cd5d30).

EnricoMi (approved):
LGTM!

PR #2772, opened 2023-09-29
Add `GithubObject.last_modified_datetime` to have `last_modified` as a `datetime`

chouetz:
As nearly all dates are returned as `datetime.datetime`, it would be convenient to have `last_modified` as a `datetime.datetime` as well.
As discussed below, let's introduce `last_modified_datetime` new argument to cover this need.


EnricoMi:
This would be a breaking change, and `last_modified` could rightly be considered a header attribute of type string,

I'd prefer to introduce a separate property `last_modified_datetime: Optional[datetime]`. Using `_makeDatetimeAttribute` would also help to support more than just a single hard-coded format.

chouetz:
> This would be a breaking change, and `last_modified` could rightly be considered a header attribute of type string,
> 
> I'd prefer to introduce a separate property `last_modified_datetime: Optional[datetime]`. Using `_makeDatetimeAttribute` would also help to support more than just a single hard-coded format.
> 
> And here should be some asserts on this property in tests.

Many thanks for your comment, I'll update the PR this way and let you know when it's ready

chouetz:
Probably needs #2783 for the failing job, @EnricoMi can you check the PR content again please?
Many thanks in advance

github/GithubObject.py
```diff
@@ -302,6 +302,13 @@ def last_modified(self) -> Optional[str]:
         """
         return self._headers.get(Consts.RES_LAST_MODIFIED)  # type: ignore
 
+    @property
+    def last_modified_datetime(self) -> Optional[datetime]:
+        """
+        :type: datetime
+        """
+        return self._makeDatetimeAttribute(self._headers.get(Consts.RES_LAST_MODIFIED))  # type: ignore
```
EnricoMi, line 310: We can simplify this to
```suggestion
        return self._makeDatetimeAttribute(self.last_modified)  # type: ignore
```

EnricoMi: Method `self._makeDatetimeAttribute` returns an `Attribute`, but we want to return the value, so we need to
```suggestion
        return self._makeDatetimeAttribute(self._headers.get(Consts.RES_LAST_MODIFIED)).value  # type: ignore
```

EnricoMi (changes requested):

github/GithubObject.py
```diff
@@ -302,6 +302,13 @@ def last_modified(self) -> Optional[str]:
         """
         return self._headers.get(Consts.RES_LAST_MODIFIED)  # type: ignore
 
+    @property
+    def last_modified_datetime(self) -> Optional[datetime]:
+        """
+        :type: datetime
+        """
+        return self._makeDatetimeAttribute(self.last_modified)  # type: ignore
```
chouetz, line 310:
```suggestion
        return self._makeDatetimeAttribute(self.last_modified).value  # type: ignore
```

chouetz:
Thanks for your time @EnricoMi

chouetz:
OK for review @EnricoMi, many thanks in advance

EnricoMi (approved):
LGTM!

PR #2788, opened 2023-10-11
Add `make_latest` to `GitRelease.update_release`

fb929:
1. The problem:
def "update_release" can not make release as "latest"

2. How solving the problem:
adding optional parameter "make_latest", from original api doc: https://docs.github.com/en/free-pro-team@latest/rest/releases/releases?apiVersion=2022-11-28#update-a-release--code-samples

### DIFF
--- a/github/GitRelease.py
+++ b/github/GitRelease.py
@@ -165,6 +165,7 @@ class GitRelease(CompletableGithubObject):
         message: str,
         draft: bool = False,
         prerelease: bool = False,
+        make_latest: Opt[bool] = NotSet,
         tag_name: Opt[str] = NotSet,
         target_commitish: Opt[str] = NotSet,
     ) -> GitRelease:
@@ -188,6 +189,9 @@ class GitRelease(CompletableGithubObject):
             "draft": draft,
             "prerelease": prerelease,
         }
+        if make_latest is not NotSet:
+            assert isinstance(make_latest, bool), make_latest
+            post_parameters["make_latest"] = make_latest
         # Do not set target_commitish to self.target_commitish when omitted, just don't send it
         # altogether in that case, in order to match the Github API behaviour. Only send it when set.
         if target_commitish is not NotSet:


trim21:
can you also update document link in docstring? current link is out of data.

pfarrell:
@fb929 I was just about to submit a PR to do exactly this.  
Are you planning to complete this PR?  This is a very feature I needed and had to implement by calling the github rest api directly.  I'd much prefer to have support in PyGithub.

github/GitRelease.py
```diff
@@ -188,6 +189,9 @@ def update_release(
             "draft": draft,
             "prerelease": prerelease,
         }
+        if make_latest is not NotSet:
+            assert isinstance(make_latest, bool), make_latest
```
EnricoMi, line 193: documentation says, this is a string
```suggestion
            assert make_latest in ["true", "false", "legacy"], make_latest
```

EnricoMi (changes requested):
While you are here, can you please add `discussion_category_name` as well?

EnricoMi:
@pfarrell feel free to fork @fb929's repo and branch and create a new PR to finish this. This way, @fb929's contribution is credited and we get this over the finish line.

EnricoMi:
Looks like `message` has been renamed to `body`. This should be reflected in another PR to avoid this to break in the future. https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28

github/GitRelease.py
```diff
@@ -165,6 +165,7 @@ def update_release(
         message: str,
         draft: bool = False,
         prerelease: bool = False,
+        make_latest: Opt[bool] = NotSet,
```
EnricoMi, line 168: Adding this here may break user code, can you add this at the end of the argument list?

EnricoMi:
@pfarrell @fb929 can we get this over the finish line?

treee111:
I tackled the PR review of this PR in the linked PR. @trim21 and @EnricoMi feel free to review.

> Looks like `message` has been renamed to `body`. This should be reflected in another PR to prevent this from breaking in the future. https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28

we already send/post that as `body` to the Github API and have `message` only as input to PyGithub if I am correct? So no need for adjustment (because that would break user code). See:
https://github.com/PyGithub/PyGithub/blob/7e7653fb7cb054d3b03dd5f89dea1a2a79350827/github/GitRelease.py#L204-L210

pfarrell:
> @pfarrell @fb929 can we get this over the finish line?

@EnricoMi, apologies for missing the alerts on this, thanks @@treee111 for picking this up

treee111:
You are welcome @pfarrell. Feel free to test the coding and comment if it worked for you! Helps the PR getting merged :-)

EnricoMi:
Closed in favour of #2888.

PR #2801, opened 2023-10-21
run fix-headers.py

trim21:


EnricoMi:
Can we still have this since we have changed `fix-headers.py` and headers are so out-dated?

trim21:
I don't want to resolve conflicting, you can run scripts from main branch

PR #2804, opened 2023-10-28
Remove dependency on python-dateutil

lazka:
According to https://docs.github.com/en/rest/overview/resources-in-the-rest-api#schema the GitHub API only returns timestamps in the format YYYY-MM-DDTHH:MM:SSZ which we can parse with the Python stdlib back to 3.7 if we handle the "Z" ourselves.

This removes the test cases using fractional seconds since the stdlib (< 3.11) can't parse them and according to the above GitHub doesn't return them anyway.

### DIFF
--- a/github/GithubObject.py
+++ b/github/GithubObject.py
@@ -50,7 +50,6 @@ from decimal import Decimal
 from operator import itemgetter
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
-from dateutil import parser
 from typing_extensions import Protocol, TypeGuard
 
 from . import Consts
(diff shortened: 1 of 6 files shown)


lazka:
Just did this because I noticed the new dependencies (dateutil/six) being installed an this seemed possible, but I can understand if this is deemed too risky/not worth it :)

EnricoMi:
Thanks for the pointer to the timestamp schema, I was looking for that but could not find it. With that, we could simplify parsing the timestamps as you suggest.

However, in #2772 we are going to parse the last-modified header timestamp, which has not such a constraint. With that, we would still have to depend on the `python-dateutil` package.

However, I think we can still apply your simplification. Let's first merge #2772 and then adjust your PR.

What do you think?

lazka:
> However, in #2772 we are going to parse the last-modified header timestamp, which has not such a constraint. With that, we would still have to depend on the `python-dateutil` package.

I went digging a bit how other libs handle this, and for example in werkzeug the parsing is done here: https://github.com/pallets/werkzeug/blob/eafbed0ce2a6bdf60e62de82bf4a8365188ac334/src/werkzeug/http.py#L961-L987 as can be seen here: https://github.com/pallets/werkzeug/blob/eafbed0ce2a6bdf60e62de82bf4a8365188ac334/src/werkzeug/sansio/http.py#L51

Maybe something to consider.

> However, I think we can still apply your simplification. Let's first merge #2772 and then adjust your PR.
> 
> What do you think?

Sure, sounds good. I'm happy to update the PR if needed.

EnricoMi:
#2772 has been merged, go for it!

lazka:
I've split the changes into 4 commits now, see the commit messages for details.

The first one ports HTTP header parsing, the second GitHub API parsing, then porting of some unrelated tests, and finally the dependency is removed.

EnricoMi (approved):
Very clean PR, thanks for the contribution!

lazka:
Thanks!

PR #2806, opened 2023-11-01
Add `internal` as valid Repository visibility value

AndrewJDawes:
Issue:

- PyGithub currently throws an `AssertionError` when attempting to create or edit a Repository where param `visibility="internal"`.
- This is an issue because `"internal"` is a valid Repository visibility value for GitHub Enterprise users.
- [https://docs.github.com/en/enterprise-cloud@latest/rest/repos/repos?apiVersion=2022-11-28#create-an-organization-repository](https://docs.github.com/en/enterprise-cloud@latest/rest/repos/repos?apiVersion=2022-11-28#create-an-organization-repository)

Example of the issue:

```python
org = gh.get_organization(organization)
org.create_repo(name="test-repo", visibility="internal")
# throws AssertionError and fails to create repo
```

Proposed fix (in pull request): 

- Allow `"internal"` as a valid value for `visibility`

Example:

```python
org = gh.get_organization(organization)
org.create_repo(name="test-repo", visibility="internal")
# does not throw assertion error and successfully creates repo
```


AndrewJDawes:
Fixed #2810

EnricoMi (approved):
LGTM!

PR #2811, opened 2023-11-06
Add mTLS support

ukaj:
Issue:
- Currently PyGitHub doesn't allow to authenticate using client certs (mTLS), because cert property is not passed to requests call

Example:
```python
gh = Github(cert='/tmp/client_cert.pem')
TypeError: Github.__init__() got an unexpected keyword argument 'cert'

```

Proposed fix:
- Add cert property to Github object, which will allow to use cert for requests as described in documentation https://requests.readthedocs.io/en/latest/user/advanced/#client-side-certificates


EnricoMi:
Does GH or GH Enterprise support mTLS authentication? Can this be combined with any other available authentication like token or login?

EnricoMi:
@ukaj ping

EnricoMi:
Once we have access to the `session` of the `Requester`, this can be implemented as an `auth.Auth`.

Issue #2815, opened 2023-11-10
Wrong URL path formation at variable edition/deleting


appsec-igm-sbf:
https://github.com/PyGithub/PyGithub/blob/dc37d5c1852a1c2f7083118adeda22760d575420/github/Variable.py#L99C25-L99C55

# Description: 

The self.url is already actions/variables/{variable_name} and does not need to repeat the path as it is in the code. 

# Evidences: 

Malformed URL: 
PATCH https://api.github.com/repos/grupo-sbf/pci-api/actions/variables/CX_ADDITIONAL_PARAMETERS/actions/variables/CX_ADDITIONAL_PARAMETERS

`{
    "headers": {
        "Authorization": "token (oauth token removed)",
        "User-Agent": "PyGithub/Python",
        "Content-Type": "application/json"
    },
    "data": {
        "name": "CX_ADDITIONAL_PARAMETERS",
        "value": "oie"
    },
    "status": 404,
    "responseHeaders": {
        "server": "GitHub.com",
        "date": "Fri, 10 Nov 2023 16:12:00 GMT",
        "content-type": "application/json; charset=utf-8",
        "transfer-encoding": "chunked",
        "x-oauth-scopes": "admin:org_hook, codespace, project, read:discussion, read:org, read:packages, read:user, repo, user:email, workflow",
        "x-accepted-oauth-scopes": "repo",
        "github-authentication-token-expiration": "2023-12-01 13:25:29 UTC",
        "x-github-media-type": "github.v3; format=json",
        "x-github-api-version-selected": "2022-11-28",
        "x-ratelimit-limit": "5000",
        "x-ratelimit-remaining": "4973",
        "x-ratelimit-reset": "1699635340",
        "x-ratelimit-used": "27",
        "x-ratelimit-resource": "core",
        "access-control-expose-headers": "ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset",
        "access-control-allow-origin": "*",
        "strict-transport-security": "max-age=31536000; includeSubdomains; preload",
        "x-frame-options": "deny",
        "x-content-type-options": "nosniff",
[...]

Issue #2785, opened 2023-10-10
Variable modification results in 404 due to malformed URL


a-chaudhari:
The `url` attribute in the Variable class already has the form of `https://api.github.com/repos/OWNER/REPO/actions/variables/VAR_NAME`

However the edit and delete commands also append `/actions/variables/VAR_NAME` onto this leading to `https://api.github.com/repos/OWNER/REPO/actions/variables/VAR_NAME/actions/variables/VAR_NAME`

This causes a 404 and so the commands don't work.  Either the edit/delete commands need to no longer append this string or the self.url attribute should not include it

EnricoMi:
Would you mind sketching a quick PR to fix this?

Issue #2834, opened 2023-11-18
404 and other errors when calling `delete()` method on `Github.Secret.Secret` and `GitHub.Variable.Variable`


AndrewJDawes:
**Methods called on `Secret` and `Variable` model methods are not requesting the correct GitHub REST API endpoint URLs. Invoking these methods causes 404 errors.**

- For `Variable`, there is a bug that appends additional incorrect parts to the variable's REST API endpoint URL, resulting in 404 errors for `delete` (HTTP `DELETE`) and `edit` (HTTP `PATCH`).
  - Example:
    - Expected: `/repos/{owner}/{repo}/actions/variables/{name}`
    - Actual: `/repos/{owner}/{repo}/actions/variables/{name}/actions/variables/{name}`
- For both `Variable` and `Secret`, fetching a `PaginatedList` of either from the `Repository` model's `get_variables` or `get_secrets` methods fails to correctly set the `_url` attribute on the paginated list of models. Therefore, calling such methods will also fail.

Below are some examples for how to reproduce the unexpected behavior. Please substitute `"AndrewJDawes/PyGithub"` with the owner/name of a repository to which you have access.

```python
# Setup
repo = self.g.get_repo("AndrewJDawes/PyGithub")
variable = repo.create_variable("VARIABLE_NAME", "VARIABLE_VALUE")
repo = self.g.get_repo("AndrewJDawes/PyGithub")
variable = repo.get_variable("VARIABLE_NAME")


# Issue
variable.delete() # github.GithubException.UnknownObjectException: 404 {"message": "Not Found", "documentation_url": "https://docs.github.com/rest"}
```

```python
repo = self.g.get_repo("AndrewJDawes/PyGithub")
repo.create_variable("VARIABLE_ONE_NAME", "VARIABLE_ONE_VALUE")
repo.create_variable("VARIABLE_TWO_NAME", "VARIABLE_TWO_VALUE")
repo = self.g.get_repo("AndrewJDawes/PyGithub")
variables = repo.get_variables()
for variable in variables:
    if variable.name in ["VARIABLE_ONE_NAME", "VARIABLE_TWO_NAME"] and variable.value in ["VARIABLE_ONE_VALUE", "VARIABLE_TWO_VALUE"]:
        variable.delete() # Fails to delete Variable
```

```python
  repo = self.g.get_repo("AndrewJDawes/PyGithub")
  repo.create_secret("SECRET_ONE_NAME", "SECRET_ONE_VALUE")
[...]

PR #2835, opened 2023-11-18
Fix Variable and Secret url bugs

AndrewJDawes:
Fixes: #2815, #2834

Tests added to `tests/Repository.py`:

- `testRepoSecrets`
- `testRepoVariables`

Note:

- This fix for Variable and Secret urls is a dependency for new functionality to fetch `Environment` variables and secrets
  - i.e., [https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28#list-environment-variables](https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28#list-environment-variables)
- Adding the optional `attributeTransformer` callable argument will also make it possible to support this Variable and Secret functionality on the `Environment` model.
- I am currently working on a pull request to add support for `get_secrets` and `get_variables` on the `Environment` model (tests are not added yet)!

**If you have any questions or if there is anything I can help with, would you please let me know?**

We utilize this library at my work place to for automations against repos in our GitHub Enterprise organizations, so we are eager to get some of these bugs fixed and new functionality added to fully support Environment variables and secrets.


AndrewJDawes:
Also fixes #2785

EnricoMi:
Thanks for the contribution.

I'd prefer to move the logic that builds the Variable url from the Repository (that uses an attributes transformer), into the Variable class. 

For this:
- add attribute `variables_url` to Variable
- in `Variable.url`, construct `url` from `variables_url` and `name`, if `self._url.` is not set
- in `Repository`, inject `variables_url` into paginated elements via `attributes_transformer_override_from_dictionary`

EnricoMi:
Btw, in 1b85bfc I have restored compatibility of `PaginatedList` with Pickle, which is good to have.

AndrewJDawes:
Hey, @EnricoMi !

Pushed up a new commit to implement your suggestions!

EnricoMi:
> Pushed up a new commit to implement your suggestions!

Looks exactly like what I meant! Now lets iterate over the PR some more...

github/PaginatedList.py
```diff
@@ -153,12 +154,18 @@ def __init__(
...
 
         first_page = []
         if firstData is not None and firstHeaders is not None:
             first_page = self._getPage(firstData, firstHeaders)
         super().__init__(first_page)
 
+    def _transformAttribute(self, attribute: Any) -> Any:
+        if self._attributesTransformer is None:
+            return attribute
+        return self._attributesTransformer(attribute)
```
EnricoMi, line 167: I got the naming / semantics wrong, I think, this is more accurate:
```suggestion
    def _transformAttributes(self, element: Dict[str, Any]) -> Dict[str, Any]:
        if self._attributesTransformer is None:
            return element
        return self._attributesTransformer(element)
```

github/PaginatedList.py
```diff
@@ -270,5 +276,29 @@ def get_page(self, page: int) -> List[T]:
...
+            self.__contentClass(self.__requester, headers, self._transformAttribute(element), completed=False)
+            for element in data
+        ]
 
-        return [self.__contentClass(self.__requester, headers, element, completed=False) for element in data]
+    @classmethod
+    def attributes_transformer_override_from_dictionary(
+        cls, transformer: Callable[[Dict[str, Any]], Dict[str, Any]], overrides: dict
+    ) -> Callable[[Dict[str, Any]], Dict[str, Any]]:
+        def attributes_transformer(attributes: Dict[str, Any]) -> Dict[str, Any]:
```
EnricoMi, line 288:
```suggestion
        def attributes_transformer(element: Dict[str, Any]) -> Dict[str, Any]:
```

github/PaginatedList.py
```diff
@@ -140,6 +140,7 @@ def __init__(
         list_item: str = "items",
         firstData: Optional[Any] = None,
         firstHeaders: Optional[Dict[str, Union[str, int]]] = None,
+        attributesTransformer: Optional[Callable[[Any], Any]] = None,
```
EnricoMi, line 143:
```suggestion
        attributesTransformer: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
```

github/PaginatedList.py
```diff
@@ -270,5 +276,29 @@ def get_page(self, page: int) -> List[T]:
...
             data = data[self.__list_item]
+        return [
+            self.__contentClass(self.__requester, headers, self._transformAttribute(element), completed=False)
+            for element in data
+        ]
 
-        return [self.__contentClass(self.__requester, headers, element, completed=False) for element in data]
+    @classmethod
+    def attributes_transformer_override_from_dictionary(
+        cls, transformer: Callable[[Dict[str, Any]], Dict[str, Any]], overrides: dict
```
EnricoMi, line 286:
```suggestion
        cls, transformer: Callable[[Dict[str, Any]], Dict[str, Any]], overrides: Dict[str, Any]
```

github/PaginatedList.py
```diff
@@ -270,5 +276,29 @@ def get_page(self, page: int) -> List[T]:
...
+    ) -> Callable[[Dict[str, Any]], Dict[str, Any]]:
```
EnricoMi, line 287: Why do we need this extra `transformer`? Can we simplify this to

```suggestion
    def override_attributes(cls, overrides: Dict[str, Any]) -> Callable[[Dict[str, Any]], Dict[str, Any]]:
```

and then use it as

```python
attributesTransformer=PaginatedList.override_attributes({"secrets_url": f"{self.url}/actions/secrets"}),
```

AndrewJDawes:
Hey, @EnricoMi !

Those suggestions make sense to me! I think it reads cleaner.

I submitted another commit that implements your suggestions. **Would you take a look when you get a chance?**

*Context on why the attributes transformers were initially accepting a `Callable`:*

- I am still hoping to submit a follow-up PR t after this one to provide the same functionality for Repository Environments and their Variables and Secrets.
- To that end, I wanted to make sure there these enhancements wouldn't prevent us from adding that functionality. - Specifically, I wasn't sure about moving control over the attributes transformations into `PaginatedList` and removing the possibility of a decorator pattern to wrap transformers (that was why `lambda` functions and each transformer accepted a `Callable`).
- However, after looking things over, I think moving control to `PaginatedList` and removing the decorator pattern should be fine and we should still be able to add support for Repository Environments and their Variables/Secrets.
- **So, I implemented your suggestions!**

AndrewJDawes:
Hey, @EnricoMi - just noting that I removed the extra unwanted `transformer`. Thought I had done that already - my mistake!

Would you please let me know if there is anything else I can help with?

EnricoMi (approved):
This looks really good, thanks for the contribution!

PR #2851, opened 2023-12-05
Support editing PullRequestReview

ColasGael:
Corresponding GitHub API: https://docs.github.com/en/rest/pulls/reviews#update-a-review-for-a-pull-request

---

Hello there, this is my first contribution to the project.
Happy to address any feedback!


github/PullRequestReview.py
```diff
@@ -100,6 +100,22 @@ def delete(self) -> None:
...
+        assert isinstance(body, str), body
+        post_parameters = {
+            "body": body,
+        }
+        headers, data = self._requester.requestJsonAndCheck(
+            "PUT",
+            f"{self.pull_request_url}/reviews/{self.id}",
+            input=post_parameters,
+        )
+        self._useAttributes(data)
```
ColasGael, line 117: I am wondering why we do not call:
```
        self._useAttributes(data)
```
in: `dismiss` above?

Since the `state` should change to `DISMISSED`.

EnricoMi: Looks like we should, can you add this here too?

EnricoMi (approved):
LGTM!

github/PullRequestReview.py
ColasGael: Done here: https://github.com/PyGithub/PyGithub/pull/2854

ColasGael:
@EnricoMi Can we merge this?

Since a dedicated PR has been opened for the other fix: https://github.com/PyGithub/PyGithub/pull/2854

PR #2868, opened 2024-01-05
Create release with optional name and message when generate_release_notes is true

heitorpolidoro:
Name and Message are not required when generate_release_notes it true in the API: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release

Broke if you want to use generate_release_notes without passing name or message to generate automatically


github/Repository.py
```diff
@@ -1207,10 +1207,10 @@ def create_git_tag_and_release(
         self,
         tag: str,
         tag_message: str,
-        release_name: str,
-        release_message: str,
         object: str,
         type: str,
+        release_name: Opt[str] = NotSet,
```
EnricoMi, line 1212: This is a breaking change, but I cannot see how to avoid that...

EnricoMi: Please adjust the order of the `:param`s in the docstring.

github/Repository.py
```diff
@@ -1264,8 +1264,8 @@ def create_git_release(
         :rtype: :class:`github.GitRelease.GitRelease`
         """
         assert isinstance(tag, str), tag
-        assert isinstance(name, str), name
-        assert isinstance(message, str), message
+        assert is_optional(name, str), name
```
EnricoMi, line 1267: Should we consider the value of `generate_release_notes`? As you said, it is optional only if `generate_release_notes` is true.

github/Repository.py
```diff
@@ -1273,8 +1273,8 @@ def create_git_release(
         :rtype: :class:`github.GitRelease.GitRelease`
         """
         assert isinstance(tag, str), tag
-        assert isinstance(name, str), name
-        assert isinstance(message, str), message
+        assert generate_release_notes and is_optional(name, str) or isinstance(name, str), name
+        assert generate_release_notes and is_optional(message, str) or isinstance(message, str), message
         assert isinstance(draft, bool), draft
         assert isinstance(prerelease, bool), prerelease
         assert isinstance(generate_release_notes, bool), generate_release_notes
```
EnricoMi, line 1280: I think if we swap the `or` elemens, this reads more clearly:
```suggestion
        assert isinstance(generate_release_notes, bool), generate_release_notes
        assert isinstance(name, str) or generate_release_notes and is_optional(name, str), name
        assert isinstance(message, str) or generate_release_notes and is_optional(message, str), message
        assert isinstance(draft, bool), draft
        assert isinstance(prerelease, bool), prerelease
```
And we need to assert `generate_release_notes` before using it as a `bool`.

github/Repository.py
EnricoMi: I am afraid we have to revert this breaking change. Please replace
```
        release_name: str,
        release_message: str,
```
with
```
        release_name: Opt[str],
        release_message: Opt[str],
```
instead.

Calling code will have to explicitly provide `release_name=None, release_message=None`.

EnricoMi (changes requested):
Happy to approve, please see
- https://github.com/PyGithub/PyGithub/pull/2868/files#r1497080554
- https://github.com/PyGithub/PyGithub/pull/2868/files#r1497083233

heitorpolidoro:
Sorry for the delay in fixing

github/Repository.py
```diff
@@ -1273,23 +1274,27 @@ def create_git_release(
         :rtype: :class:`github.GitRelease.GitRelease`
         """
         assert isinstance(tag, str), tag
-        assert isinstance(name, str), name
-        assert isinstance(message, str), message
+        assert isinstance(generate_release_notes, bool), generate_release_notes
+        assert isinstance(name, str) or generate_release_notes and (name is None or is_optional(name, str)), name
+        assert (
+            isinstance(message, str) or generate_release_notes and (message is None or is_optional(message, str))
+        ), message
```
EnricoMi, line 1281: This project doesn't use `None` for unset optional arguments but `NotSet`. Staying with the convention:
```suggestion
        assert isinstance(name, str) or generate_release_notes and is_optional(name, str), name
        assert (
            isinstance(message, str) or generate_release_notes and is_optional(message, str)
        ), message
```

EnricoMi (changes requested):
Does this work for you?

PR #2876, opened 2024-01-17
Fixing the tests and Update repository public key Config

smkillen:
Added the missing replay files for creating Org and Repo secrets for Actions/Dependabot.

Updated the Repository Code to pass the secret_type to the public_key retrieval function


smkillen:
Raised in Error

Issue #2870, opened 2024-01-11
GraphQL API for Enterprise GitHub


shicholas:
When using the GraphQL API such as for features like `enable_auto_merge`, a user should be allowed to specify the URL. AFAICT it's hardcoded https://github.com/PyGithub/PyGithub/blob/603896f44541740d72aa441ed60c698760a1393e/github/Requester.py#L543C81-L543C81

samad-yar-khan:
@EnricoMi  I have been using PyGithub in our project for over a year. If no one is working on this, can I open a PR and try to resolve this issue ?

EnricoMi:
@shicholas are you planning to create a PR?

samad-yar-khan:
@EnricoMi Thanks, I will look into into it and try solving this issue.

samad-yar-khan:
@shicholas @EnricoMi I couldn't open a PR within time as I was still navigating the cod, had some other time constraints and saw a PR already opened for the issue. Thanks for the PRs, went through them to learn a bit about the codebase.

PR #2878, opened 2024-01-19
Use the hostname and port for GraphQL connections

shicholas:
Addresses https://github.com/PyGithub/PyGithub/issues/2870

@EnricoMi, I noticed that there are some ConnectionClasses, is that what you had envisioned? should the `requestJsonAndCheck` method be moved to those classes?

Thanks for this awesome lib!


github/Requester.py
```diff
@@ -540,7 +540,7 @@ def graphql_query(self, query: str, variables: Dict[str, Any]) -> Tuple[Dict[str
         """
         input_ = {"query": query, "variables": {"input": variables}}
 
-        response_headers, data = self.requestJsonAndCheck("POST", "https://api.github.com/graphql", input=input_)
+        response_headers, data = self.requestJsonAndCheck("POST", f"https://{self.__hostname}:{self.__port}/graphql", input=input_)
```
EnricoMi, line 543: Right, that is easier than I thought, but that should do it.

I'd suggest to construct the GraphQl URL in the `__init__` method right after `o = urllib.parse.urlparse(base_url)` (line 397) like this:
```
o = urllib.parse.urlparse(base_url)
self.__graphql_url = f"{o.scheme}://{o.hostname}:{o.port}/graphql"
```

or maybe safer would be this
```
o = urllib.parse.urlparse(base_url)
self.__graphql_url = urlunparse(o._replace(path="graphql"))
```

And then use `self.__graphql_url` here instead:
```suggestion
        response_headers, data = self.requestJsonAndCheck("POST", self.__graphql_url, input=input_)
```

EnricoMi:
We could use some test here.

github/Requester.py
EnricoMi: Above suggestion using `urlunparse` should also fix the test error `ValueError: Port could not be cast to integer value as 'None'`.

EnricoMi:
I have fixed the imports...

EnricoMi:
I have added a few tests. Please test the code against your enterprise server.

EnricoMi:
What is the path of your v3 REST API? Is it `/api/v3`?

EnricoMi:
If you don't want the attribution, I have to do this without your commits: #2880

EnricoMi:
Fixed in #2880.

EnricoMi:
Fixed in #2880.

PR #2882, opened 2024-01-24
Make `Repository.compare().commits` return paginated list

### LINKED ISSUE #2027: [enhancement] Repository.compare() function should return PaginatedList of commits
I have a script that uses PyGithub's `Repository.compare()` function to generate a list of changes between two commits.  Our repository has become fairly busy of late, and I only recently discovered that this API returns a page size of 250 commits, by default.  Documentation is [here](https://docs.github.com/en/rest/reference/repos#compare-two-commits) on the corresponding GitHub API.

When PyGithub consumes this API, the `commits` property in the `Comparison` object is just a plain old list.  It would be great if this could be a `PaginatedList` instead, so that we can consume all of the commits without worrying about the pagination.  I think this may be a bit weird, because the commits list is only part of the returned response.  There's also a list of files, which doesn't support pagination.  Additionally, according to the documentation, the sort order of the commits returned changes when pagination parameters are provided.

It would be fantastic if all of this weirdness could be hidden away from users of PyGithub!  But I'm not sure that it's straightforward to do because of the weirdness.  I'm happy to work on a PR to provide this, but would prefer not to do so without a better understanding from PyGithub's maintainers about how it should work.

For anyone who may come across this in the very near future, I'm working around this in a really terrible way by appending the pagination parameters onto the `head` parameter with each call, i.e. `repo.compare(base, f"{head}?page={page}")`.

EnricoMi:
Fixes #2027.


JLLeitschuh (approved):
LGTM!

PR #2873, opened 2024-01-16
Add missing branch protection fields

treee111:
## overview

this is a collection of currently open pull requests targeting branch protection field updates with the goal to have a consistent code base of these fields in PyGithub. Consistent code base means both reading the fields and updating them through PyGithub.

I cherry picked the commits so that the contributors of the superseded PR's don't lose their credits.

## related PyGithub issues

supersedes PR'S #2793, #2716, #2378
closes #2360, closes #2373, closes #2862, closes #2299, closes #2239

## details
this table shows the state of the fields in the actual implementation (current `main` branch)
| attribute  | read actual impl. (BranchProtection.py) | write actual impl. (Branch.py) | read source  PR | write source PR  |
|---|---|---|---|---|
| `required_linear_history` | exists | exists  | - | - |
| `allow_force_pushes` | - | exists  | 2793 | - |
| `allow_deletions` | - | - | 2793 |  2716 |
| `allow_fork_syncing` | -  | exists | 2793 | - |
| `lock_branch` | - | exists | 2793 | - | OK 
| `required_conversation_resolution` | - | exists | 2793 | - |
| `require_last_push_approval` | - (RequiredPullRequestReviews.py) | - | 2716 | 2716 |

## testing
I tested all field with my scripts and manually checked Github UI as well as the return of PyGithub via debugger.
In addition, all unittests are also green.

PR #2884, opened 2024-01-26
Add branch protection support for `required_status_checks.checks` object

### LINKED ISSUE #2157: Add branch protection support for required_status_checks `checks` object
It's possible to set [branch protections](https://pygithub.readthedocs.io/en/latest/github_objects/BranchProtection.html#github.BranchProtection.BranchProtection.required_status_checks) using the [contexts](https://pygithub.readthedocs.io/en/latest/github_objects/RequiredStatusChecks.html#github.RequiredStatusChecks.RequiredStatusChecks.contexts) field within the `RequiredStatusChecks` class, however, according to the documentation, [contexts is deprecated](https://docs.github.com/en/rest/reference/branches#update-branch-protection).

`checks` (an array of objects) will replace `contexts` as it allows you to configure the source of the check. 

An example of the raw request to set required status checks using the `checks` object is:

```
    "required_status_checks":{
        "strict": false,
        "checks": [
            {
                "context": "A Required Check set to a given app id",
                "app_id": 3414
            },
            {
                "context": "A Required Check set to any source",
                "app_id": -1
            }
        ]
    }
```

image

treee111:
as outlined in https://github.com/PyGithub/PyGithub/pull/2873#issuecomment-1906795864:
> Looking at the [docs](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#update-branch-protection) I found that argument `required_status_checks.contexts` is deprecated in the API and replaced by `required_status_checks.checks`.

I've done:
- [x] adds `checks` as another argument
- [x] `checks` should be of type `List[Union[str, Tuple[str, int]]]` so that this allows for `checks=["check1", ("check2", -1), ("check3", 123456)]`
- [x] if only `context` is given, it's values should be used to populate `required_status_checks.checks.context`, rather than `required_status_checks.contexts`
- [x] if both `checks` and `context` are given, `context` should be ignored

Fixes #2157


treee111:
Hi @EnricoMi could you please have a look at the coding if this is the right direction and what needs to be adjusted. Thanks!

github/Branch.py
```diff
@@ -130,6 +130,7 @@ def edit_protection(
         self,
         strict: Opt[bool] = NotSet,
         contexts: Opt[list[str]] = NotSet,
+        checks: Opt[list[Union[str, Tuple[str, int]]]] = NotSet,
```
EnricoMi: Please add at the end of the signature, or you may break user code.

github/Branch.py
```diff
@@ -180,14 +181,25 @@ def edit_protection(
...
+            elif is_defined(checks):
+                post_parameters["checks"] = []
+                for check in checks:
+                    if isinstance(check, tuple):
+                        context, app_id = check
+                        post_parameters["checks"].append({"context": context, "app_id": app_id})
+                    else:
+                        post_parameters["checks"].append({"context": check})
+            elif is_defined(contexts):
+                post_parameters["checks"] = [{"context": context} for context in contexts]
```
EnricoMi, line 198: I think you want to populate list `checks` here, as used below to set `post_parameters["required_status_checks"]`.

tests/ReplayData/Branch.testEditRequiredStatusChecksChecks.txt
```diff
@@ -0,0 +1,21 @@
+https
+PATCH
+api.github.com
+None
+/repos/jacquev6/PyGithub/branches/integrations/protection/required_status_checks
+{'Authorization': 'Basic login_and_password_removed', 'Content-Type': 'application/json', 'User-Agent': 'PyGithub/Python'}
+{"checks": [{"context": "check1"}, {"context": "check2", "app_id": -1}, {"context": "check3", "app_id": 123456}]}
+200
+[('status', '200 OK'), ('x-ratelimit-remaining', '4994'), ('content-length', '0'), ('server', 'nginx/1.0.13'), ('connection', 'keep-alive'), ('x-ratelimit-limit', '5000'), ('etag', '"81ba94a48ad867b26d48c023ac584f43"'), ('date', 'Mon, 07 May 2018 13:42:41 GMT'), ('content-type', 'application/json; charset=utf-8')]
+''
```
EnricoMi, line 10: looks like this is not correct

github/Branch.py
```diff
@@ -327,14 +339,29 @@ def edit_required_status_checks(
...
+                if isinstance(check, tuple):
+                    context, app_id = check
+                    post_parameters["checks"].append({"context": context, "app_id": app_id})
+                else:
+                    post_parameters["checks"].append({"context": check})
+
+        elif is_defined(contexts):
+            post_parameters["checks"] = [{"context": context} for context in contexts]
 
-        post_parameters: dict[str, Any] = NotSet.remove_unset_items({"strict": strict, "contexts": contexts})
```
EnricoMi, line 337: please keep the `NotSet.remove_unset_items` pattern where it exists, preferred over `is_defined + post_parameters = ... above`.

EnricoMi (changes requested):
Generally right direction, just minor style comments.

github/Branch.py
```diff
@@ -327,14 +339,29 @@ def edit_required_status_checks(
...
         strict: Opt[bool] = NotSet,
         contexts: Opt[list[str]] = NotSet,
+        checks: List[Union[str, Tuple[str, int]]] = NotSet,
     ) -> RequiredStatusChecks:
         """
         :calls: `PATCH /repos/{owner}/{repo}/branches/{branch}/protection/required_status_checks <https://docs.github.com/en/rest/reference/repos#branches>`_
         """
         assert is_optional(strict, bool), strict
         assert is_optional_list(contexts, str), contexts
+        # assert is_optional_list(checks, Union[str, Tuple[str, int]]), checks
```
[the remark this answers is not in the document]
treee111: thanks, have done the `if checks is not None:` in another way because that way it didn't worked for `None`. Is this OK?

github/Branch.py
treee111: thanks, didn't think about that - fixed!

tests/ReplayData/Branch.testEditRequiredStatusChecksChecks.txt
treee111: how to create the file automatically? This does not work for me in macOS
`Branch.testEditRequiredStatusChecksChecks --record --auth_with_token`

```
(pygithub-dev) <user>@<computer> PyGithub % pytest -k Branch.testEditRequiredStatusChecksChecks --record --auth_with_token
=============================================================================================== test session starts ===============================================================================================
platform darwin -- Python 3.10.13, pytest-7.4.4, pluggy-1.3.0
rootdir: /Users/<user>/VSCode/python/PyGithub
configfile: pyproject.toml
plugins: github-actions-annotate-failures-0.2.0, subtests-0.11.0, cov-4.1.0
collected 890 items / 890 deselected / 0 selected                                                                                                                                                                 

============================================================================================= 890 deselected in 0.32s =============================================================================================
```

github/Branch.py
treee111: tried to comply with the pattern but didn't found a example in PyGithub which uses kind of calculation of the parameters like I do here.
I dont think https://github.com/PyGithub/PyGithub/pull/2884/commits/8776ae5e421d5b7a739d9113e63ab05f97f01efd makes it easier to read, maybe you could make it clearer what I have to adjust based on that.

micahbeeman-wf:
@treee111 @EnricoMi Any update on this? Would love this functionality.

EnricoMi:
I have reworked checks post parameters and recorded test data.

EnricoMi (approved):
LGTM!

treee111:
thanks for reworking this PR @EnricoMi! 🥇

PR #2886, opened 2024-01-28
Release v2.2.0

EnricoMi:
- Updates file headers
- Updates `doc/changes.rst`
- Script `manage.sh` is outdated, renamed and reduced to `scripts/prepare-release`.


scripts/prepare_release.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
sfdye, line 1: Can you pls update https://github.com/PyGithub/PyGithub/blob/main/DEPLOY.md as well?

JLLeitschuh (approved):
I suggest making the change suggested by @sfdye, other than that, this LGTM!

scripts/prepare_release.sh
EnricoMi: Done

PR #2888, opened 2024-01-30
Add `make_latest` to `GitRelease.update_release`

treee111:
### The problem:

- def "update_release" can not make release as "latest"

### How solving the problem:

- adding optional parameter "make_latest", from original api doc: https://docs.github.com/en/free-pro-team@latest/rest/releases/releases?apiVersion=2022-11-28#update-a-release--code-samples

### additions

- documentation URL has changed
- `discussion_category_name` has been added

superseds #2788


github/GitRelease.py
```diff
@@ -187,9 +187,11 @@ def update_release(
...
         target_commitish: Opt[str] = NotSet,
+        make_latest: Opt[str] = NotSet,
+        discussion_category_name: Opt[str] = NotSet,
     ) -> GitRelease:
         """
-        :calls: `PATCH /repos/{owner}/{repo}/releases/{release_id} <https://docs.github.com/en/rest/reference/repos#update-a-release>`_
+        :calls: `PATCH /repos/{owner}/{repo}/releases/{release_id} <https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release>`_
         """
         assert tag_name is NotSet or isinstance(tag_name, str), "tag_name must be a str/unicode object"
         assert target_commitish is NotSet or isinstance(
```
EnricoMi, line 197: Please add the assertion here, remove the next to lines, and move all these four assertions below `assert isinstance(prerelease, bool), prerelease`, to reflect  the order of arguments in the method signature:

```suggestion
        assert is_optional(tag_name, str), "tag_name must be a str/unicode object"
        assert is_optional(target_commitish, str), "target_commitish must be a str/unicode object"
        assert make_latest in ["true", "false", "legacy", NotSet], make_latest
        assert is_optional(discussion_category_name, str), discussion_category_name
```

github/GitRelease.py
```diff
@@ -208,6 +210,12 @@ def update_release(
             "draft": draft,
             "prerelease": prerelease,
         }
+        if make_latest is not NotSet:
```
EnricoMi, line 213: Let's not mix setting `post_parameter` with assertions, see above.

EnricoMi (changes requested):

github/GitRelease.py
treee111: done. I also resorted the `post_parameters` statements to have the same order. Only the defaulting of `tag_name` is out of order but I think it more unified now, thanks for the heads-up!

jantmanAtCox:
This would be really useful for me, any hope of having it merged?

EnricoMi (approved):
LGTM!

github/GitRelease.py
```diff
@@ -73,7 +73,6 @@ class GitRelease(CompletableGithubObject):
 
     The reference can be found here
     https://docs.github.com/en/rest/reference/repos#releases
-
```
EnricoMi, line 76: The doc style requires this line, breaks lint test.

github/GitRelease.py
```diff
@@ -73,7 +73,6 @@ class GitRelease(CompletableGithubObject):
...
     """
```
EnricoMi, line 76: lets revert this:
```suggestion

    """
```

github/GitRelease.py
treee111: thanks for correcting!

jmaixl:
When can this be released?

Issue #2807, opened 2023-11-02
Add support for repositories in app access token creation


sfc-gh-hpathak:
GitHub allows changing the scope of permissions for an app while generating an access token for an app. 
[API](https://github.com/PyGithub/PyGithub/blob/main/github/GithubIntegration.py#L197)

Note that this takes `repositories` as the key also so we can scope down the access token to only specific repositories. This, however, is not supported in PyGithub today.  This is because the code below adds the permissions `str` directly to the `permissions` key and sends it as the body.

[code](https://github.com/PyGithub/PyGithub/blob/main/github/GithubIntegration.py#L197) 

It would be great if we can specify the repo list also as expected by the Github API

PR #2892, opened 2024-02-03
Add Ability to Scope Install Access Tokens to a List of Repositories

bakosa:
## Description

As you can see here in the [GitHub Docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app#generating-an-installation-access-token) you can now pass in ether a list of repos via the `repositories` or `repository_ids` body parameters when you create an app installation access token. 

The problem that this is trying to solve is we need to generate least privileged tokens scoped to only specific repositories and specific permissions. 

This PR adds new `token_repositories` and `token_repository_ids` arguments to the `AppInstallationAuth` class and uses them when invoking the `get_access_token` method to get a scoped app install access token.

## Things to Note
I tired to follow the same pattern that is being used for the `permissions` body parameter. Instead of defaulting to `{}` when nothing is passed in, None is preferred here since that will translate to null which drive the expected outcome (tested this). Also, you can pass a combination of both `repositories` and `repository_ids` in because 🦄 

## Related open issues

https://github.com/PyGithub/PyGithub/issues/2807


bakosa:
Is there anything I can do to help get this merged in 😃

mchieco:
Any updates on this issue/PR? Was looking around for this functionality and if we could get this merged in that would be great so I don't have to fork for this feature

github/AppAuthentication.py
```diff
@@ -49,9 +49,13 @@ def __init__(
         private_key: str,
         installation_id: int,
         token_permissions: Optional[Dict[str, str]] = None,
+        token_repositories: Optional[List[str]] = None,
+        token_repository_ids: Optional[List[int]] = None,
```
EnricoMi, line 53: What about unifying:
```suggestion
        token_repositories: Optional[List[str | int]] = None,
```
just because 🦄.

github/GithubIntegration.py
```diff
@@ -212,18 +218,29 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
+            raise GithubException(status=400, data={"message": "Invalid repositories"}, headers=None)
+
+        if repository_ids is not None and not isinstance(repository_ids, list):
+            raise GithubException(status=400, data={"message": "Invalid repository_ids"}, headers=None)
+
         if not isinstance(permissions, dict):
             raise GithubException(status=400, data={"message": "Invalid permissions"}, headers=None)
 
-        body = {"permissions": permissions}
+        body = {"permissions": permissions, "repositories": repositories, "repository_ids": repository_ids}
```
EnricoMi, line 243: This will remove the elements that are `None`:
```suggestion
        body = {"permissions": permissions}
        if repositories is not None:
          body["repositories"] = repositories
        if repository_ids is not None:
          body["repository_ids"] = repository_ids
```

EnricoMi (changes requested):

bakosa:
This should be good to go now!

github/Auth.py
```diff
@@ -190,15 +190,17 @@ def get_installation_auth(
         installation_id: int,
         token_permissions: Optional[Dict[str, str]] = None,
         requester: Optional[Requester] = None,
+        token_repositories: Optional[List[Union[str, int]]] = None,
```
EnricoMi, line 193: Though this method is public, the `requester` argument is internal in a sense that there should be no user code calling this method with `requester` given. Hence, it is safe (and preferable) to add `token_repositories` before `requester`:
```suggestion
        token_permissions: Optional[Dict[str, str]] = None,
        token_repositories: Optional[List[Union[str, int]]] = None,
        *,
        requester: Optional[Requester] = None,
```

Adding the `*` makes the `requester` argument mandatory named, which avoids breaking user code in the future when adding an argument before it.

This requires code that provides `requester` to add `requester=...` throughout the project.

github/Auth.py
```diff
@@ -190,15 +190,17 @@ def get_installation_auth(
...
     ) -> "AppInstallationAuth":
         """
         Creates a github.Auth.AppInstallationAuth instance for an installation.
         :param installation_id: installation id
         :param token_permissions: optional permissions
         :param requester: optional requester with app authentication
+        :param token_repositories: optional repositories or repository ids
```
EnricoMi, line 200:
```suggestion
        :param token_permissions: optional permissions
        :param token_repositories: optional repositories or repository ids
        :param requester: optional requester with app authentication
```

github/Auth.py
```diff
@@ -256,16 +258,19 @@ def __init__(
         installation_id: int,
         token_permissions: Optional[Dict[str, str]] = None,
         requester: Optional[Requester] = None,
+        token_repositories: Optional[List[Union[str, int]]] = None,
```
EnricoMi, line 261:
```suggestion
        token_permissions: Optional[Dict[str, str]] = None,
        token_repositories: Optional[List[Union[str, int]]] = None,
        *,
        requester: Optional[Requester] = None,
```

github/Auth.py
```diff
@@ -190,15 +190,17 @@ def get_installation_auth(
...
         """
         Creates a github.Auth.AppInstallationAuth instance for an installation.
         :param installation_id: installation id
         :param token_permissions: optional permissions
         :param requester: optional requester with app authentication
+        :param token_repositories: optional repositories or repository ids
         :return:
         """
-        return AppInstallationAuth(self, installation_id, token_permissions, requester)
+        return AppInstallationAuth(self, installation_id, token_permissions, requester, token_repositories)
```
EnricoMi, line 203:
```suggestion
        return AppInstallationAuth(self, installation_id, token_permissions, token_repositories, requester=requester)
```

github/GithubIntegration.py
```diff
@@ -174,10 +174,13 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
...
     def get_github_for_installation(
-        self, installation_id: int, token_permissions: dict[str, str] | None = None
+        self,
+        installation_id: int,
+        token_permissions: dict[str, str] | None = None,
+        token_repositories: list[str | int] | None = None,
     ) -> github.Github:
         # The installation has to authenticate as an installation, not an app
-        auth = self.auth.get_installation_auth(installation_id, token_permissions, self.__requester)
+        auth = self.auth.get_installation_auth(installation_id, token_permissions, self.__requester, token_repositories)
```
EnricoMi, line 183:
```suggestion
        auth = self.auth.get_installation_auth(installation_id, token_permissions, token_repositories, requester=self.__requester)
```

github/GithubIntegration.py
```diff
@@ -212,18 +215,41 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
         if not isinstance(permissions, dict):
             raise GithubException(status=400, data={"message": "Invalid permissions"}, headers=None)
 
-        body = {"permissions": permissions}
+        body["permissions"] = permissions
+        if repositories:
+            repository_names = []
+            repository_ids = []
+            for r in repositories:
+                print(r, type(r))
```
EnricoMi, line 242: this debug line should be removed

github/GithubIntegration.py
```diff
@@ -212,18 +215,41 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
+                if isinstance(r, str):
+                    repository_names.append(r)
+                elif isinstance(r, int):
+                    repository_ids.append(r)
+
+            if repository_names:
+                body["repositories"] = repository_names
+            if repository_ids:
+                body["repository_ids"] = repository_ids
```
EnricoMi, line 251: What about list comprehension?
```suggestion
            if repository_names:
                body["repositories"] = [r for r in repositories if isinstance(r, str)]
            if repository_ids:
                body["repository_ids"] = [r for r in repositories if isinstance(r, int)]
```

EnricoMi (changes requested):

EnricoMi:
Note that #2912 will add some `Requester`-related assertions to `github.Auth` that might conflict with your changes.

github/GithubIntegration.py
bakosa: Wow, that is just embarrassing 🤦‍♂️

github/Auth.py
bakosa: I was wondering the same thing when I went to add the new argument. This makes sense! thanks for the deeper explanation.

bakosa:
> Note that #2912 will add some `Requester`-related assertions to `github.Auth` that might conflict with your changes.

Sounds good, hopefully this PR is now ready! Happy to handle any conflicts if #2912 gets in first. Curious the reasoning behind all the runtime type checking?

EnricoMi:
Merge, happy conflict chasing!

bakosa:
Resolved 🚀

github/GithubIntegration.py
```diff
@@ -212,18 +218,35 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
     ) -> InstallationAuthorization:
         """
         :calls: `POST /app/installations/{installation_id}/access_tokens <https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app>`
         """
+        body: dict[str, Any] = {}
+
         if permissions is None:
             permissions = {}
 
+        if repositories is not None and not isinstance(repositories, list):
```
EnricoMi, line 234:
```suggestion
        if repositories is not None and not (isinstance(repositories, list) and all(isinstance(r, (str, int)) for r in repositories)):
```

github/GithubIntegration.py
```diff
@@ -212,18 +218,35 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
+            raise GithubException(status=400, data={"message": "Invalid repositories"}, headers=None)
+
         if not isinstance(permissions, dict):
             raise GithubException(status=400, data={"message": "Invalid permissions"}, headers=None)
 
-        body = {"permissions": permissions}
+        body["permissions"] = permissions
```
EnricoMi, line 240:
```suggestion
        body: dict[str, Any] = {"permissions": permissions}
        body["permissions"] = permissions
```

github/GithubIntegration.py
```diff
@@ -212,18 +218,35 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
+        self,
+        installation_id: int,
+        permissions: dict[str, str] | None = None,
+        repositories: list[str | int] | None = None,
     ) -> InstallationAuthorization:
         """
         :calls: `POST /app/installations/{installation_id}/access_tokens <https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app>`
         """
+        body: dict[str, Any] = {}
+
```
EnricoMi, line 230: I think this can be merged with the line below.
```suggestion
```

tests/GithubIntegration.py
```diff
@@ -268,10 +268,30 @@ def testGetAccessTokenWithInvalidPermissions(self):
...
-        with self.assertRaises(github.GithubException) as raisedexp:
+
+        with self.assertRaises(github.GithubException) as raisedPermissionExp:
             github_integration.get_access_token(self.repo_installation_id, permissions="invalid_data")
+        self.assertEqual(raisedPermissionExp.exception.status, 400)
+
+        with self.assertRaises(github.GithubException) as raisedRepositoriesExp:
+            github_integration.get_access_token(self.repo_installation_id, repositories="invalid_data")
+        self.assertEqual(raisedRepositoriesExp.exception.status, 400)
+
```
EnricoMi, line 279: Lets also test for invalid list element types
```suggestion

        with self.assertRaises(github.GithubException) as raisedRepositoriesExp:
            github_integration.get_access_token(self.repo_installation_id, repositories=[False])
        self.assertEqual(raisedRepositoriesExp.exception.status, 400)

```

EnricoMi (changes requested):
Minor changes, LGTM overall!

EnricoMi:
I have added the new argument to the examples docs: 80407cc4a36756c48b80e269edbd8c1eda8b0111

github/GithubIntegration.py
```diff
@@ -212,18 +218,34 @@ def create_jwt(self, expiration: int | None = None) -> str:
...
+        permissions: dict[str, str] | None = None,
+        repositories: list[str | int] | None = None,
     ) -> InstallationAuthorization:
         """
         :calls: `POST /app/installations/{installation_id}/access_tokens <https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app>`
         """
         if permissions is None:
             permissions = {}
 
+        if repositories is not None and not (isinstance(repositories, list) and all(isinstance(r, (str, int)) for r in repositories)):
```
bakosa, line 232: The unit tests fails here because in Python, booleans are subclass of integers 🙃 

https://docs.python.org/3/c-api/bool.html

bakosa, line 232: The unit test fails here because in Python, booleans are subclass of integers 🙃 

https://docs.python.org/3/c-api/bool.html

We could ether be explicit about the type check (exclude subclasses..)

```suggestion
        if repositories is not None and not (isinstance(repositories, list) and all(type(r) in (str, int) for r in repositories)):
```

Or change the unit test. Not sure why someone would pass in a subclassed object or str or int into the list of repositories but I will leave it up to you!

EnricoMi: Let's stick to `isinstance` as we use it everywhere while it is not accurate, as you pointed out. Can you fix the unittest by simply using `object()`?

bakosa: Makes sense, updated the test.

PR #2894, opened 2024-02-05
fixes a temporary directory in source tree to crash building PyGithub

### LINKED ISSUE #2887: [2.2.0] Extra directory in source tree trashes build
Just found that after create some temp/ directory in source tree I was unable to build module. Looks like some bits are missing in pyproject.toml.
Here is test case:
```console
+ mkdir temp
+ /usr/bin/python3 -sBm build -w --no-isolation
* Getting build dependencies for wheel...
error: Multiple top-level packages discovered in a flat-layout: ['github', 'temp'].

To avoid accidental inclusion of unwanted files or directories,
setuptools will not proceed with this build.

If you are trying to create a single distribution with multiple packages
on purpose, you should not rely on automatic discovery.
Instead, consider the following options:

1. set up custom discovery (`find` directive with `include` or `exclude`)
2. use a `src-layout`
3. explicitly set `py_modules` or `packages` with a list of names

To find more information, look for "package discovery" on setuptools docs.

ERROR Backend subprocess exited when trying to invoke get_requires_for_build_wheel
```

treee111:
- closes #2887
- explicitly name the modules that should be build in pyproject.toml to tell the build system which modules it should build
- works when creating a empty dir with name `temp` in root of PyGithub
- for reference see https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,9 @@ integrations = []
 
 [tool.setuptools_scm]
 
+[tool.setuptools]
+packages = ["github"]
+
 [tool.setuptools.package-data]
 github = ["py.typed", '*.pyi']
 


EnricoMi (approved):
LGTM!

EnricoMi:
@treee111 thanks for fixing this!

kloczek:
> feel free to test if this also works in your environment @kloczek

Just tested that and it works as expected 👍

PR #2896, opened 2024-02-06
Get pulls by commit sha

OksanaTka:
Add ability to get related pull request by commit


EnricoMi (changes requested):
Can we add a test for this?

EnricoMi:
@OksanaTka?

EnricoMi:
Sorry, I just found that `repo.get_commit(sha).get_pulls()` exists already. I think that is sufficient. Any objection?

PR #2898, opened 2024-02-08
implement method to obtain logs download link from workflow run

jonapich:
This implements 2 new endpoints:

- Get download link for workflow run (latest attempt)
- Get download link for workflow run (WorkflowRun instance's attempt)

I also supplied a quick example of how to consume this download link, but I'm open to include this as a separate "courtesy" function (e.g.: `def get_logs_for_this_attempt() -> Dict[PurePath, str]: ...`).

I wasn't able to record the tests because I need admin access to the repository. Can an admin do that for me? 🥺

### DIFF
--- a/github/WorkflowRun.py
+++ b/github/WorkflowRun.py
@@ -191,6 +191,10 @@ class WorkflowRun(CompletableGithubObject):
         return self._logs_url.value
 
     @property
+    def run_attempt_logs_url(self) -> str:
+        return self.logs_url[:-4] + f"attempts/{self.run_attempt}/logs"
+
+    @property
     def check_suite_url(self) -> str:
         self._completeIfNotSet(self._check_suite_url)
         return self._check_suite_url.value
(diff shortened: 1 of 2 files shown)


github/WorkflowRun.py
```diff
@@ -190,6 +190,10 @@ def logs_url(self) -> str:
         self._completeIfNotSet(self._logs_url)
         return self._logs_url.value
 
+    @property
+    def run_attempt_logs_url(self) -> str:
+        return self.logs_url[:-4] + f"attempts/{self.run_attempt}/logs"
```
EnricoMi, line 195: Why not
```suggestion
        return self.url + f"/attempts/{self.run_attempt}/logs"
```

EnricoMi:
Looks like there is precedence for following the `Location` link to download file content: https://github.com/PyGithub/PyGithub/blob/0b8435fccbcc98404f08e146b6e259bd20065c98/github/Repository.py#L2147-L2149

jonapich:
> Looks like there is precedence for following the `Location` link to download file content:
> 
> https://github.com/PyGithub/PyGithub/blob/0b8435fccbcc98404f08e146b6e259bd20065c98/github/Repository.py#L2147-L2149

Doing it this way will prevent users from using a streaming response. Large files need a separate download method.

PR #2901, opened 2024-02-17
document Properties in class docstrings

wakamex:
part one

### DIFF
--- a/github/AccessToken.py
+++ b/github/AccessToken.py
@@ -47,7 +47,26 @@ from github.GithubObject import Attribute, NonCompletableGithubObject, NotSet
 
 class AccessToken(NonCompletableGithubObject):
     """
-    This class represents access tokens.
+    Encapsulates access token details for API authentication.
+
+    Properties
+    ----------
+    token : str
+        Access token for API requests.
+    type : str
+        Type of token, usually 'bearer'.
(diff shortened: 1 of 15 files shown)


EnricoMi:
Is this auto-generated or hand-crafted?

wakamex:
good question. it does look very suspect.

a bit of both?

I started off by looking up the documentation for CommitStats because I wanted to know which stats are available. I was surprised that isn't listed in the documentation, so decided to add it. CommitStats is super vanilla, since it's just 3 items. Then I noticed none of the other classes documented their attributes in docstrings, so I started flipping through them. To make that go faster I used GPT4 to generate descriptions from the _initAttributes, asking it to be concise, and reviewed them very briefly. I'm no expert on these objects, especially the more esoteric ones like CVE advisories, but they look right to me. Despite the problems of genAI hallucination, in this case I'm fairly confident GPT4 has been trained on PyGithub (likely code and examples). To be 100% certain and avoid any risk of incorrect description hallucinated by these models, I can cut out all the docstrings except for CommitStats, which I'm confident in.

My thinking was some attribute documentation is better than none, especially when flipping through the Github object documentation here: https://pygithub.readthedocs.io/en/stable/github_objects/IssueEvent.html.

PR #2912, opened 2024-03-03
Assert requester argument type in Auth

EnricoMi:
Type of `requester` argument is not asserted in `github.Auth`. This fixes this and removes some redundant assertions.


EnricoMi:
CC @JLLeitschuh

JLLeitschuh (approved):
LGTM!

EnricoMi:
Thanks!

PR #2923, opened 2024-03-19
chore: fix docs import

wurstbrot:
didn't work without it for me

### DIFF
--- a/doc/examples/Authentication.rst
+++ b/doc/examples/Authentication.rst
@@ -9,6 +9,8 @@ All authentication methods require this import:
 .. code-block:: python
 
     >>> from github import Auth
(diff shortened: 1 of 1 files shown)


doc/examples/Authentication.rst
```diff
@@ -9,6 +9,8 @@ All authentication methods require this import:
 .. code-block:: python
 
     >>> from github import Auth
+    >>> from github import Github
+
```
EnricoMi, line 13: I think, while we are there, `GithubIntegration` is also needed:
```suggestion
    >>> from github import Github
    >>> from github import GithubIntegration
```

EnricoMi (approved):
LGTM!

PR #2927, opened 2024-03-22
Requester: fix type annotation for requestBlobAndCheck

fsouza:
requestBlob returns `tuple[int, dict[str, Any], str]`, so requestBlobAndCheck should return `tuple[dict[str, Any], str]`.


fsouza:
oh wait I misread the code. closing

PR #2929, opened 2024-03-24
Fix requesting urls containing parameters with parameters dict

### LINKED ISSUE #1136: Paginated list reversed only gives the first page of results backwards
When I do
```
g = Github("xxx")
repo = g.get_repo("xxx")

for comm in repo.get_commits(since=datetime).reversed:
    ....
```
it only goes through the most recent 30 results in reverse order.

I was able to fix this with
```
def reverse_github_results(paginated_list):
    for i in range(paginated_list.totalCount//30, -1, -1):
        page = paginated_list.get_page(i)
        page.reverse()
        for item in page:
            yield item
```
but this possibly runs a bit slower and I thought I should let you guys know

EnricoMi:
Requesting an url that contains parameters (query part of the url) did not support giving a parameters dict:

    Requester.requestJson(verb, "https://api.github.com/?per_page=10", {"per_page": 20})

Now, parameters given in the URL have precedence over the dict.

Iterating over reversed `PaginatedList` is affected by this.

Fixes #1136.


JLLeitschuh (commented):
Why allow duplicate request parameters at all, why not throw an exception?

If we're doing this internally, this seems like a bug?

EnricoMi:
You are right, calling this method this way does not make much sense. But the code path that ends up in this situation is complex. Other code paths might suffer from the same issue, and it is not trivial to identify those.

It is expected that the values of the parameters are identical, so this is the right thing to do in that situation. I could throw an exception when parameter values contradict. But then this becomes a breaking change for some people where this just behaved incorrectly.

Felixoid:
Hello. Is there a chance we can have it merged?

EnricoMi:
@JLLeitschuh you are right, now `add_parameters_to_url` gives precedence to the given parameters and overwrites params that exist in the url. The code path that produced the conflict now removes all params from `parameters` that are contained in the url, giving precedence in that particular situation only (which is the right thing to do there).

PR #2930, opened 2024-03-26
Add rerun-failed-jobs for WorkflowRun

alvicsam:
GitHub API provides an ability to rerun only failed jobs in a workflow: https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#re-run-failed-jobs-from-a-workflow-run

PR adds `rerun_failed_jobs` attribute. I also tried to add tests but it seems I need some help with them or a hint how can I fix them.

### DIFF
--- a/github/WorkflowRun.py
+++ b/github/WorkflowRun.py
@@ -86,6 +86,7 @@ class WorkflowRun(CompletableGithubObject):
         self._check_suite_url: Attribute[str] = NotSet
         self._cancel_url: Attribute[str] = NotSet
         self._rerun_url: Attribute[str] = NotSet
+        self._rerun_failed_jobs_url: Attribute[str] = NotSet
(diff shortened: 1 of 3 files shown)


alvicsam:
I tried to fix tests but current approach won't work for `rerun-failed-jobs`. 
As I understood tests are recording json answer from a request to https://api.github.com/repos/PyGithub/PyGithub/actions/runs/<run_number> and then they are looking for a specific URL (for example https://api.github.com/repos/PyGithub/PyGithub/actions/runs/3910280793/rerun). However, url https://api.github.com/repos/PyGithub/PyGithub/actions/runs/3910280793/rerun-failed-jobs exists and works but api [doesn't provide it in it's answer](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run).

PR #2936, opened 2024-04-04
Link Commit to correct upstream documentation

xmo-odoo:
`github.Commit.Commit` represents a github-level commit, but the link is to git-level commits. The semantics and *especially* the metadata are very different between the two e.g. a github-level commit provides the *github users* for the author and committer, as well as links to the github comments.

The github-level commit is a layer around the git-level commit, which is (partially) provided via the `commit` attribute.


EnricoMi (approved):
LGTM!

PR #2939, opened 2024-04-05
Make Commit.files return PaginatedList

### LINKED ISSUE #1745: commit.files result has a limit of 300 files, can I get infomation of all files?
commit.files result has a limit of 300 files.
I have a large commit that has 926 changed files and I need to list the changes.
when I use         
files = commit.files
print(len(commit.files))
I got 300.
Can I get all the files?

iarspider:
[`/commits/{ref}` endpoint](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit) is  paginated, meaning you can get up to 3000 files per commit:
 
> Note: If there are more than 300 files in the commit diff and the default JSON media type is requested, the response will include pagination link headers for the remaining files, up to a limit of 3000 files. Each page contains the static commit information, and the only changes are to the file listing.

As suggested [here](https://github.com/PyGithub/PyGithub/issues/1745#issuecomment-1991754882).

Fixes #1745.


tests/BadAttributes.py
```diff
@@ -97,15 +97,6 @@ def testBadAttributeInClassAttribute(self):
             owner.avatar_url
         self.assertEqual(raisedexp.exception.actual_value, 42)
 
-    def testBadTransformedAttributeInList(self):
```
EnricoMi, line 100: I think instead of removing this test case, we should find a different object that returns a list of a class and use that to test the bad transform case.

EnricoMi (approved):
LGTM!

iarspider:
Thanks!

PR #2940, opened 2024-04-05
Fix docstring formatting in `create_file` and `update_file`

xmo-odoo:
Formatting is currently broken as can be seen on [the published documentation](https://pygithub.readthedocs.io/en/v2.3.0/github_objects/Repository.html#github.Repository.Repository.create_file): line-breaking to the first column makes rST thing the info list has ended and go back to regular content, making a hash of the rest of the commit until reaching the next indented section which is interpreted as a quote, following which the new info fields are seen again.

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -2387,8 +2387,7 @@ class Repository(CompletableGithubObject):
         """
         Create a file in this repository.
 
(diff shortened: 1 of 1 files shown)


EnricoMi:
This project uses a custom field `:calls:` that docformatter does not recognize: https://github.com/PyCQA/docformatter/issues/279#issuecomment-2122227941

PR #2945, opened 2024-04-08
Resolve login when repr-ing AuthenticatedUser

xmo-odoo:
It's extremely confusing to dump an `AuthenticatedUser` and get `AuthenticatedUser(login=None)` even though the `AuthenticatedUser` is correctly set up, and necessarily has a login.

This risks incurring additional requests, but I think in the contexts where it would be used (mostly at the shell or when debugging services) it's more helpful to actually show the information.

### DIFF
--- a/github/AuthenticatedUser.py
+++ b/github/AuthenticatedUser.py
@@ -177,7 +177,7 @@ class AuthenticatedUser(CompletableGithubObject):
         self._two_factor_authentication: Attribute[bool] = NotSet
 
     def __repr__(self) -> str:
-        return self.get__repr__({"login": self._login.value or "<UNKNOWN>"})
+        return self.get__repr__({"login": self._login.value or "(UNKNOWN)"})
 
     @property
     def avatar_url(self) -> str:


EnricoMi:
I like the `UNKNOWN` representation when login is not set.

github/AuthenticatedUser.py
```diff
@@ -177,7 +177,7 @@ def _initAttributes(self) -> None:
         self._two_factor_authentication: Attribute[bool] = NotSet
 
     def __repr__(self) -> str:
-        return self.get__repr__({"login": self._login.value})
+        return self.get__repr__({"login": self._login.value or "<UNKNOWN>"})
```
EnricoMi, line 180: I think `self._login.value or "<UNKNOWN>"` does not work as you would expect.
```suggestion
        return self.get__repr__({"login": self._login.value if is_defined(self._login.value) else "<UNKNOWN>"})
```

EnricoMi, line 180: I'd prefer `(UNKNOWN)`.
```suggestion
        return self.get__repr__({"login": self._login.value or "(UNKNOWN)"})
```

EnricoMi (changes requested):
Please add a test that asserts the repr before and after fetching the login.

PR #2956, opened 2024-04-22
Add return values to Team operations where the API specifies a return value, but we didn't honor it so far

jdelic:
These are just small improvements.

Specifically:
* [Adding a team membership](https://docs.github.com/en/rest/teams/members?apiVersion=2022-11-28#add-or-update-team-membership-for-a-user) returns the same schema/data structure as getting a team membership, so we should return a `Membership` instance.
* [Deleting a team](https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#delete-a-team) and [removing a team membership](https://docs.github.com/en/rest/teams/members?apiVersion=2022-11-28#remove-team-membership-for-a-user) both return HTTP 204 on success and so, like other examples in the API, these methods should return a `bool` to show whether they were succesful.


github/Team.py
```diff
@@ -298,11 +300,15 @@ def update_team_repository(self, repo: Repository, permission: str) -> bool:
...
+    def delete(self) -> bool:
         """
         :calls: `DELETE /teams/{id} <https://docs.github.com/en/rest/reference/teams#delete-a-team>`_
         """
-        headers, data = self._requester.requestJsonAndCheck("DELETE", self.url)
+        status, headers, data = self._requester.requestJson(
+            "DELETE",
+            self.url,
+        )
+        return status == 204
```
EnricoMi, line 311: This is bad practise: #2760

github/Team.py
```diff
@@ -411,12 +417,13 @@ def has_in_repos(self, repo: Repository) -> bool:
...
 
-    def remove_membership(self, member: NamedUser) -> None:
+    def remove_membership(self, member: NamedUser) -> bool:
         """
         :calls: `DELETE /teams/{team_id}/memberships/{username} <https://docs.github.com/en/rest/reference/teams#remove-team-membership-for-a-user>`_
         """
         assert isinstance(member, github.NamedUser.NamedUser), member
-        headers, data = self._requester.requestJsonAndCheck("DELETE", f"{self.url}/memberships/{member._identity}")
+        status, headers, data = self._requester.requestJson("DELETE", f"{self.url}/memberships/{member._identity}")
+        return status == 204
```
EnricoMi, line 426: same here

github/Team.py
```diff
@@ -218,6 +219,7 @@ def add_membership(self, member: NamedUser, role: Opt[str] = NotSet) -> None:
         headers, data = self._requester.requestJsonAndCheck(
             "PUT", f"{self.url}/memberships/{member._identity}", input=put_parameters
         )
+        return github.Membership.Membership(self._requester, headers, data, completed=True)
```
EnricoMi, line 222: Happy to approve this change, but can you extend the tests in `tests/Team.py`? Tests already call `add_membership`, so simply assert some attributes of the returned `Membership` instance.

EnricoMi (changes requested):

PR #2965, opened 2024-05-06
Add workflow create_dispatch method that raises exceptions

dblanchette:
We've been having issues with creating workflow dispatch with missing/too many arguments.

When using Workflow.create_dispatch(), we only get a boolean response even though the GitHub API returns detailed error messages.

This creates a new method that returns an exception with the error data when it fails. I did that not to break the existing method.


dblanchette:
Broke the PR by rebasing our fork, I'll open a new one

PR #2966, opened 2024-05-07
Add workflow create_dispatch method that raises exceptions

dblanchette:
(Reopen broken #2965)

We've been having issues with creating workflow dispatch with missing/too many arguments.

When using Workflow.create_dispatch(), we only get a boolean response even though the GitHub API returns detailed error messages.

This creates a new method that returns an exception with the error data when it fails. I did that not to break the existing method.


EnricoMi:
See #2760 as the general design issue.

EnricoMi:
Maybe we should add attribute `throw: bool = False` to `create_dispatch`, which when `False` exhibits the current (flawed) behaviour (`return status == 204`), and when set `True` calls into `requestJsonAndCheck`, which will raise errors when status is not 2xx.

ngc92:
this would have saved me quite a bit of time.

dblanchette:
> Maybe we should add attribute `throw: bool = False` to `create_dispatch`, which when `False` exhibits the current (flawed) behaviour (`return status == 204`), and when set `True` calls into `requestJsonAndCheck`, which will raise errors when status is not 2xx.
> 
> In v3, `throw` will be set to `True` as the default (breaking change) and will eventually be removed.

I agree. The changes are made.

github/Workflow.py
```diff
@@ -160,10 +158,27 @@ def create_dispatch(
...
+        :calls: `POST /repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches <https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event>`_
+
+        """
+        try:
+            self._create_dispatch(ref, inputs)
+        except GithubException.GithubException:
+            if throw:
+                raise
+            return False
+        return True
```
EnricoMi, line 181: Shouldn't we still check for 202 status?

```suggestion
        return status == 204
```

EnricoMi (changes requested):
Please merge `_create_dispatch` back into `create_dispatch` where only

    self._requester.requestJsonAndCheck

is inside they `try` branch.

github/Workflow.py
```diff
@@ -160,10 +158,27 @@ def create_dispatch(
             ref = ref.name
         if inputs is NotSet:
             inputs = {}
-        status, _, _ = self._requester.requestJson(
-            "POST", f"{self.url}/dispatches", input={"ref": ref, "inputs": inputs}
-        )
-        return status == 204
+        self._requester.requestJsonAndCheck("POST", f"{self.url}/dispatches", input={"ref": ref, "inputs": inputs})
+
+    def create_dispatch(
```
EnricoMi, line 163:
```suggestion
    # v3: default throw to True
    def create_dispatch(
```

github/Workflow.py
dblanchette: The doc says the only possible return status is 204 https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event

github/Workflow.py
```diff
@@ -141,11 +142,18 @@ def url(self) -> str:
...
+        self,
+        ref: github.Branch.Branch | github.Tag.Tag | github.Commit.Commit | str,
+        inputs: Opt[dict] = NotSet,
+        throw: bool = False,
     ) -> bool:
         """
+        Call Create Dispatch, raises or return False without details on error, depending on the "throw" parameter.
+
         :calls: `POST /repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches <https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event>`_
+
```
EnricoMi: To align with other methods' docstrings:
```suggestion
        :calls: `POST /repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches <https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event>`_
        Note: raises or return False without details on error, depending on the ``throw`` parameter.
```

github/Workflow.py
```diff
@@ -162,10 +170,13 @@ def create_dispatch(
...
-            "POST", f"{self.url}/dispatches", input={"ref": ref, "inputs": inputs}
-        )
-        return status == 204
+        try:
+            self._requester.requestJsonAndCheck("POST", f"{self.url}/dispatches", input={"ref": ref, "inputs": inputs})
+        except GithubException.GithubException:
+            if throw:
+                raise
+            return False
+        return True
```
EnricoMi, line 179: Revisiting this, non-exceptional code paths that use exceptions is bad code design, let's avoid the check if throw is false:
```suggestion
        if throw:
            self._requester.requestJsonAndCheck("POST", f"{self.url}/dispatches", input={"ref": ref, "inputs": inputs})
            return True
        else:
            status, _, _ = self._requester.requestJson(
                "POST", f"{self.url}/dispatches", input={"ref": ref, "inputs": inputs}
            )
            return status == 204
```
This creates an inconsistency for non-204 non-error response codes when `throw == True`, but this is a general issue of this project and can be addressed separately.

EnricoMi: Let's simplify the requester calls:
```suggestion
        url = f"{self.url}/dispatches"
        input = {"ref": ref, "inputs": inputs}

        if throw:
            self._requester.requestJsonAndCheck("POST", url, input=input)
            return True
        else:
            status, _, _ = self._requester.requestJson("POST", url, input=input)
            return status == 204
```

github/Workflow.py
EnricoMi: The `status == 204` is for consistency with other methods.

dblanchette:
> Happy to approve, please consider the suggestions.

@EnricoMi I applied the suggestions, is it all good?

EnricoMi (approved):
LGTM!

PR #2973, opened 2024-05-19
add documentation for BadCredentialsException

Lslightly:
add documentations for easier root cause localization.

### DIFF
--- a/github/GithubException.py
+++ b/github/GithubException.py
@@ -106,6 +106,57 @@ class GithubException(Exception):
 class BadCredentialsException(GithubException):
     """
     Exception raised in case of bad credentials (when Github API replies with a 401 or 403 HTML status)
+
+    ### Possible Causes
+
+    This exception is raised possibly because of:
+    1. Insufficient scope of token
+    2. An expired token. In an application with **cache** mechanism, the cached PyGithub-like objects might
+    contain an expired token unexpectedly. ATTENTION: The PyGithub-like objects can not update the token
+    automatically.
+
+    ### One way to find the cause
+    
+    One way to find the root cause is as follows.
+    The steps are abstracted from https://github.com/PyGithub/PyGithub/issues/1753#issuecomment-1039533679
+    
+    1. In the source code directory of PyGithub package downloaded by
+    pip(often `lib/python3.x/site-packages/` in virtual env or `~/.local/lib/python3.x/site-packages/`), 
+    edit `github/Requester.py`.
+    
+    replace in `def __log`:
+    
+    ```python
+        elif requestHeaders["Authorization"].startswith("token"):
+        headersForRequest["Authorization"] = "token (oauth token removed)"
+    ```
+    
+    with
+    
+    ```python
+        elif requestHeaders["Authorization"].startswith("token"):
+            import hashlib
+            token = requestHeaders["Authorization"][5:]
+            token = hashlib.md5(token.encode('utf-8')).hexdigest()
+            headersForRequest["Authorization"] = f"token ({token})"
+    ```
+
+    2. Insert `github.enable_console_debug_logging()` at the entry of your program. You would see the token
+    used when you run your program. The output format as follows. **CHECK whether the output token is expected**. 
+    
+    ```
+    GET https://api.github.com/repos/totycro/stacs/branches/main {'Authorization': 'token (xxxx) 'User-Agent': ...
+    ```
+
+    ### Ways to fix
+
+    If you find the token is unexpected, you can choose to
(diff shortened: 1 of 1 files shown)


Lslightly:
I find bad looks in [readthedocs](https://pygithub--2973.org.readthedocs.build/en/2973/utilities.html#github.GithubException.BadCredentialsException) but I am not familiar with readthedocs.

But the headings work well in vscode python extension when hovering over `BadCredentialsException`.

image

github/GithubException.py
```diff
@@ -106,6 +106,57 @@ def __str__(self) -> str:
...
     Exception raised in case of bad credentials (when Github API replies with a 401 or 403 HTML status)
+
+    ### Possible Causes
+
+    This exception is raised possibly because of:
+    1. Insufficient scope of token
+    2. An expired token. In an application with **cache** mechanism, the cached PyGithub-like objects might
+    contain an expired token unexpectedly. ATTENTION: The PyGithub-like objects can not update the token
+    automatically.
+
```
EnricoMi, line 117:
```suggestion
3. A leaked token. Committing a token into your repo and pushing to github automatically invalidates the token.

```

github/GithubException.py
```diff
@@ -106,6 +106,57 @@ def __str__(self) -> str:
...
+    ### One way to find the cause
+    
+    One way to find the root cause is as follows.
+    The steps are abstracted from https://github.com/PyGithub/PyGithub/issues/1753#issuecomment-1039533679
```
EnricoMi, line 121:
```suggestion
    ### Expose the token in the logs

    Attention: This is for debugging purposes only, as this leaks the token into log files, which is a security risk.

    The best way to investigate the root cause is to show the actual token used to authenticate in the logs.
    Usually, the token is obfuscated for security reasons. Here we deliberately write the plain token to the logs,
    so that we can compare it with the expected token.
```

github/GithubException.py
```diff
@@ -106,6 +106,57 @@ def __str__(self) -> str:
...
+    
+    1. In the source code directory of PyGithub package downloaded by
+    pip(often `lib/python3.x/site-packages/` in virtual env or `~/.local/lib/python3.x/site-packages/`), 
+    edit `github/Requester.py`.
+    
+    replace in `def __log`:
```
EnricoMi, line 127:
```suggestion
    1. Identify the location of the github source code installed via pip or virtual env:
    
    ```bash
    python -c "from github import Requester; print(Requester.__file__)"
    ```

    2. Edit this `github/Requester.py` file and replace in `def __log`:
```

github/GithubException.py
```diff
@@ -106,6 +106,57 @@ def __str__(self) -> str:
...
+            token = hashlib.md5(token.encode('utf-8')).hexdigest()
+            headersForRequest["Authorization"] = f"token ({token})"
+    ```
+
+    2. Insert `github.enable_console_debug_logging()` at the entry of your program. You would see the token
+    used when you run your program. The output format as follows. **CHECK whether the output token is expected**. 
+    
+    ```
+    GET https://api.github.com/repos/totycro/stacs/branches/main {'Authorization': 'token (xxxx) 'User-Agent': ...
+    ```
```
EnricoMi, line 149:
```suggestion
    2. Insert `github.enable_console_debug_logging()` at the start of your program. You will see the token
    used for each call of the GitHub REST API. The output format as follows: 
    
    ```
    GET https://api.github.com/repos/totycro/stacs/branches/main {'Authorization': 'token (xxxx)', 'User-Agent': ...
    ```

    **CHECK whether the output token is expected**.
```

github/GithubException.py
```diff
@@ -106,6 +106,57 @@ def __str__(self) -> str:
...
+
+    ### Ways to fix
+
+    If you find the token is unexpected, you can choose to
+    1) Get the PyGithub objects again. (easier and recommended)
+    2) Renew the token in PyGithub objects by replacing the token with the newer one.
+        For `Issue` and `Repository` got through github token,
+        use `o._requester.auth._token = <new token str>` to renew objects.
```
EnricoMi, line 157:
```suggestion
    1) Recreate the Github instance with a new token: `g = Github(auth=Auth.Token("access_token"))`
        and fetch the PyGithub objects again. (recommended)
    2) Renew the token in PyGithub objects by replacing the token with the newer one.
        For `Issue` and `Repository` got through github token,
        use `o._requester.auth._token = <new token str>` to renew objects.  (not recommended)
```

EnricoMi (commented):
I'd suggest the following improvements in these instructions.

PR #2975, opened 2024-05-20
Complete dropping Python 3.7

khneal:
Python 3.12 has been GA for 6 months. Python 3.7 is EOL.

Based on #2434 and #2332


tests/WorkflowJob.py
```diff
@@ -56,7 +56,7 @@ def testAttributes(self):
         self.assertEqual(self.job.started_at, started_at)
         completed_at = datetime(2023, 2, 17, 16, 4, 52, tzinfo=timezone.utc)
         self.assertEqual(self.job.completed_at, completed_at)
-        self.assertEqual(self.job.name, "test (Python 3.7)")
+        self.assertEqual(self.job.name, "test (Python 3.8)")
```
khneal, line 59: This change breaks a unit test, so I will revert.

github/Requester.py
```diff
@@ -454,7 +454,7 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
         self.__custom_connections = deque()
 
     @staticmethod
-    # replace with str.removesuffix once support for Python 3.7 is dropped
+    # replace with str.removesuffix once support for Python 3.8 is dropped
     def remove_suffix(string: str, suffix: str) -> str:
```
EnricoMi, line 458: Can you replace the usage of `remove_suffix` with `str.removesuffix` and remove this method?

EnricoMi (changes requested):
Fine with me, please consider the proposed change.

github/Requester.py
khneal: It would break in Python 3.8. The comment was wrong, but it's fixed now.
`str.removesuffix` was added in 3.9, not 3.8.
See [PEP 616](https://peps.python.org/pep-0616/) as well as [Python Docs](https://docs.python.org/3.9/library/stdtypes.html#str.removesuffix):

raboof (approved):
can confirm that 3.12 seems to work fine, so making that 'official' seems reasonable

github/Requester.py
EnricoMi: You are right, my bad!

EnricoMi:
Fixed in #3008, but there seem to be some bits missing.

EnricoMi (approved):

github/Requester.py
khneal: No worries, and thank you!

PR #2983, opened 2024-06-03
Implement support for re-running only failed workflow jobs.

chrisgavin:
This implements support for [re-running failed Actions workflows](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#re-run-failed-jobs-from-a-workflow-run).

Unlike the regular `rerun` method, the `rerun_failed_jobs` method will only re-run the failed jobs rather than all the jobs in a run.


tests/ReplayData/WorkflowRun.test_rerun_failed_jobs.txt
```diff
@@ -0,0 +1,21 @@
...
+{"message":"Must have admin rights to Repository.","documentation_url":"https://docs.github.com/rest/actions/workflow-runs#re-run-failed-jobs-from-a-workflow-run"}
```
EnricoMi, line 21: Looks like your authentication does not allow for rerunning jobs. Please try to get a successful response here.

EnricoMi (changes requested):
Looks all good, except that the replay data do not record a successful request.

tests/ReplayData/WorkflowRun.test_rerun_failed_jobs.txt
chrisgavin: Thanks @EnricoMi. I'm very happy to try to record a successful request. I copied the approach from the regular rerun method [which also has a failure](https://github.com/PyGithub/PyGithub/blob/main/tests/ReplayData/WorkflowRun.test_rerun.txt)

The problem I was having with getting a successful request is that the test references the PyGithub repository, which I don't have access to. I could attempt to mock up the recorded data by recording it against my own repository, and then doing a find-and-replace to make the repository name and workflow run ID match?

chrisgavin: I've pushed a change which should make the recording look mostly like what it should for a successful request. Hopefully that is okay.

I'm happy to change the test though, if you think it would be better to provide a fully authentic recording for a repository that I have admin access to though.

EnricoMi (approved):
LGTM!

PR #2992, opened 2024-06-19
Fix  API path of `Repository.get_git_ref`

csalerno-asml:
Rationale:

> 1. Note that PyGithub queries "git/refs/..." in the API.
> 2. After back and forth with Github support, we found out that "git/refs/...." is internally pointing to the "List matching references" API call. - https://docs.github.com/en/enterprise-server@3.11/rest/git/refs?apiVersion=2022-11-28#list-matching-references - Whereas the "Get a ref" call is strictly limited to 'git/ref/...' (Note the missing S).
However, PyGitHub uses it in the get_git_ref method.

This has caused a bug in my application where calling `Repository.get_git_ref("foo")` would return me any reference matching `foo` - probably as prefix - , rather the one and only reference `foo`.


EnricoMi:
Good catch! Should we make the old request available as `get_git_refs` so that users have a chance to migrate if they rely on this "feature"?

EnricoMi:
@csalerno-asml ping

csalerno-asml:
hey @EnricoMi , sorry for the late answer but I seldom access github.com :) 
I will rework this, but cannot promise you any ETA

EnricoMi (approved):
LGTM!

PR #3005, opened 2024-07-16
Add minimize and unminimize functions for IssueComment class

arash77:
This pull request adds two new functions, `minimize` and `unminimize`, to the `IssueComment` class. These functions allow users to minimize and unminimize comments using the GitHub GraphQL API. The `minimize` function takes an optional reason parameter and sends a mutation to minimize the comment. The `unminimize` function sends a mutation to unminimize the comment. These functions provide more control and flexibility for managing comments in GitHub issues.


EnricoMi (approved):
LGTM! Looks like we need GraphQL testing capabilities in this project.

EnricoMi (changes requested):
Looks like there is testing available. Can you add some tests for those two new IssueComment methods?

arash77:
> Looks like there is testing available. Can you add some tests for those two new IssueComment methods?

I added the tests. Are they enough?

EnricoMi:
I have added missing test data.

arash77:
> I have added missing test data.

Thank you so much!

EnricoMi (approved):
LGTM!

PR #3006, opened 2024-07-17
Create GithubRetry.DEFAULT_* constants

### LINKED ISSUE #2970: Implement GithubRetry DEFAULT_ALLOWED_METHODS and DEFAULT_STATUS_FORCELIST
`GithubRetry`'s default values  for `status_forcelist` and `allowed_methods` are computed in `__init__`. See [here](https://github.com/PyGithub/PyGithub/blob/7266e812ed2976ea36a4303edecfe5d75522343f/github/GithubRetry.py#L79-L80). We had a use case where we wanted extend (not override) the default values, but we resorted to copy-pasting. It would be better to have class members for it, similar to `Retry.DEFAULT_ALLOWED_METHODS`. 

Instead of:
```python
GithubRetry(
    # GithubRetry default value + 401
    status_forcelist=list(range(500, 600)) + [403, 401],
    # GithubRetry default value + "PATCH". 
    # GithubRetry.DEFAULT_ALLOWED_METHODS currently inherited from urllib3.Retry
    allowed_methods=GithubRetry.DEFAULT_ALLOWED_METHODS.union(
        {"GET", "POST", "PATCH"}
    ),
)
```

it would be better to have:
```python
GithubRetry(
    status_forcelist=GithubRetry.DEFAULT_STATUS_FORCELIST + [401],
    allowed_methods=GithubRetry.DEFAULT_ALLOWED_METHODS.union({"PATCH"}),
)
```

jodelasur:
Resolves #2970.

### DIFF
--- a/github/GithubRetry.py
+++ b/github/GithubRetry.py
@@ -67,6 +67,12 @@ class GithubRetry(Retry):
     # references the class, not the module (due to re-exporting in github/__init__.py)
     __datetime = datetime
 
+    #: Default values to be used for ``status_forcelist``
+    DEFAULT_STATUS_FORCELIST = list(range(500, 600))
+
+    #: Default methods to be used for ``allowed_methods``
+    DEFAULT_ALLOWED_METHODS = Retry.DEFAULT_ALLOWED_METHODS.union({"GET", "POST"})
+
     def __init__(self, secondary_rate_wait: float = DEFAULT_SECONDARY_RATE_WAIT, **kwargs: Any) -> None:
         """
         :param secondary_rate_wait: seconds to wait before retrying secondary rate limit errors
@@ -76,8 +82,8 @@ class GithubRetry(Retry):
         # 403 is too broad to be retried, but GitHub API signals rate limits via 403
         # we retry 403 and look into the response header via Retry.increment
         # to determine if we really retry that 403
-        kwargs["status_forcelist"] = kwargs.get("status_forcelist", list(range(500, 600))) + [403]
-        kwargs["allowed_methods"] = kwargs.get("allowed_methods", Retry.DEFAULT_ALLOWED_METHODS.union({"GET", "POST"}))
+        kwargs["status_forcelist"] = kwargs.get("status_forcelist", self.DEFAULT_STATUS_FORCELIST) + [403]
+        kwargs["allowed_methods"] = kwargs.get("allowed_methods", self.DEFAULT_ALLOWED_METHODS)
         super().__init__(**kwargs)
 
     def new(self, **kw: Any) -> Self:


github/GithubRetry.py
```diff
@@ -67,6 +67,12 @@ class GithubRetry(Retry):
     # references the class, not the module (due to re-exporting in github/__init__.py)
     __datetime = datetime
 
+    #: Default values to be used for ``status_forcelist``
+    DEFAULT_STATUS_FORCELIST = list(range(500, 600))
+
+    #: Default methods to be used for ``allowed_methods``
+    DEFAULT_ALLOWED_METHODS = Retry.DEFAULT_ALLOWED_METHODS.union({"GET", "POST"})
```
EnricoMi, line 74: Can we move these next to the existing `DEFAULT_SECONDARY_RATE_WAIT` above?

jodelasur: This would result to `GithubRetry.DEFAULT_ALLOWED_METHODS` inheriting `urllib3.Retry.DEFAULT_ALLOWED_METHODS`, which might cause confusion. Is that OK? Or should I have it both as a module and class constant?

EnricoMi: I don't see a problem with a module constant being based on `urllib3.Retry.DEFAULT_ALLOWED_METHODS`.

github/GithubRetry.py
```diff
@@ -46,6 +46,8 @@
 from github.Requester import Requester
 
 DEFAULT_SECONDARY_RATE_WAIT: int = 60
+DEFAULT_STATUS_FORCELIST = list(range(500, 600))
```
EnricoMi, line 49: I think we should also add `403` to the default.
```suggestion
DEFAULT_STATUS_FORCELIST = list(range(500, 600)) + [403]
```

github/GithubRetry.py
```diff
@@ -78,8 +80,8 @@ def __init__(self, secondary_rate_wait: float = DEFAULT_SECONDARY_RATE_WAIT, **k
         # 403 is too broad to be retried, but GitHub API signals rate limits via 403
         # we retry 403 and look into the response header via Retry.increment
         # to determine if we really retry that 403
-        kwargs["status_forcelist"] = kwargs.get("status_forcelist", list(range(500, 600))) + [403]
-        kwargs["allowed_methods"] = kwargs.get("allowed_methods", Retry.DEFAULT_ALLOWED_METHODS.union({"GET", "POST"}))
+        kwargs["status_forcelist"] = kwargs.get("status_forcelist", DEFAULT_STATUS_FORCELIST) + [403]
```
EnricoMi, line 83:
```suggestion
        kwargs["status_forcelist"] = kwargs.get("status_forcelist", DEFAULT_STATUS_FORCELIST)
```

EnricoMi (changes requested):
almost there

github/GithubRetry.py
EnricoMi: @jodelasur Any objections? Otherwise I would apply these changes.

github/GithubRetry.py
jodelasur: Given a user provides a custom `status_forcelist` value without 403, the current logic will still add 403 on init. If we apply this change, this won't happen anymore and 403 won't be in the list. This could affect the custom 403 logic applied for GithubRetry. What do you think?

If you feel this is a non-issue, please go ahead and apply. I'll be offline for a week or so, and won't be able to check on this.

github/GithubRetry.py
jodelasur: I guess this is OK. See my [comment below](https://github.com/PyGithub/PyGithub/pull/3006#discussion_r1818275896) though. May we could apply this and not the other one?

PR #3025, opened 2024-08-21
Update more `SecurityAndAnalysis` attributes

squatched:
Recently, `/orgs/{org}/repos` started returning something that doesn't match the schema given in [the docs](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-organization-repositories). This updates SecurityAndAnalysis to reflect what is actually being returned by this endpoint.

I am reaching out to GitHub to determine what's going on, if this is a permanent change, then I will push this.


EnricoMi (approved):
LGTM!

PR #3028, opened 2024-08-28
Add `GitCommitVerification` class

timgates42:
Turns the return value of `GitCommit.verification` from `dict` into a proper class.

https://docs.github.com/en/rest/commits/commits


EnricoMi (changes requested):
This looks great, thanks for the contribution! Can you a test asserting `commit.verification` for a commit that has `CommitVerification` attributes set?

timgates42:
Thanks @EnricoMi have added

timgates42:
@EnricoMi let me know if anything more required - Thanks!

timgates42:
this was added it appears also via https://github.com/PyGithub/PyGithub/commit/6276e20f27da94d3fcef8e6b2635097320cac782 which is now merged so will close this pr

EnricoMi:
It is always better to have proper objects than dicts. I have also added `verified_at`.

EnricoMi:
Renamed to `GitCommitVerification`.

EnricoMi (approved):
LGTM!

PR #3040, opened 2024-09-09
Adds List organization memberships for the authenticated user

eduramirezh:
This endpoint described in [the docs](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#list-organization-memberships-for-the-authenticated-user)
was missing. This adds it as a method in the AuthenticatedUser class which returns a list of Memeberships.


EnricoMi (approved):
LGTM!

PR #3045, opened 2024-09-16
Make tests pass some more years

bmwiedemann:
Without this patch, tests fail after 2024-11-25.

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +16 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our python-PyGithub package build.
See https://reproducible-builds.org/ for why this matters.


bmwiedemann:
Yes, that is how I found this.
```
 =================================== FAILURES ===================================
 _____________ Authentication.testAppInstallationAuthAuthentication _____________
 
 self = <tests.Authentication.Authentication testMethod=testAppInstallationAuthAuthentication>
 
     def testAppInstallationAuthAuthentication(self):
         # test data copied from testAppAuthentication to test parity
         installation_auth = github.Auth.AppInstallationAuth(self.app_auth, 29782936)
         g = github.Github(auth=installation_auth)
     
         # test token expiry
         # token expires 2024-11-25 01:00:02
         token = installation_auth.token
 >       self.assertFalse(installation_auth._is_expired)
 E       AssertionError: True is not false
 
 tests/Authentication.py:182: AssertionError
 =========================== short test summary info ============================
 FAILED tests/Authentication.py::Authentication::testAppInstallationAuthAuthentication
 ======================== 1 failed, 936 passed in 31.00s ========================
```

EnricoMi:
Then we should mock the clock used by `_is_expired`:

```patch
diff --git a/tests/Authentication.py b/tests/Authentication.py
index ffd419c6..071a2147 100644
--- a/tests/Authentication.py
+++ b/tests/Authentication.py
@@ -176,16 +176,15 @@ class Authentication(Framework.BasicTestCase):
         installation_auth = github.Auth.AppInstallationAuth(self.app_auth, 29782936)
         g = github.Github(auth=installation_auth)
 
-        # test token expiry
         # token expires 2024-11-25 01:00:02
         token = installation_auth.token
-        self.assertFalse(installation_auth._is_expired)
         self.assertEqual(
             installation_auth._AppInstallationAuth__installation_authorization.expires_at,
             datetime(2024, 11, 25, 1, 0, 2, tzinfo=timezone.utc),
         )
 
-        # forward the clock so token expires
+        # test token expiry
+        # control the current time used by _is_expired
         with mock.patch("github.Auth.datetime") as dt:
             # just before expiry
             dt.now = mock.Mock(return_value=datetime(2024, 11, 25, 0, 59, 3, tzinfo=timezone.utc))
```

bmwiedemann:
OK. I can test your patch.

bmwiedemann:
I tried that patch and got a different error/trace:
https://rb.zq1.de/other/python-PyGithub/pr3045-log1.txt

bmwiedemann:
There is around one month left to solve this. What are the plans?

EnricoMi:
> I tried that patch and got a different error/trace: https://rb.zq1.de/other/python-PyGithub/pr3045-log1.txt

Probably because you did not revert your changes of the replay data.

Reverting your commit and applying the proposed patch fixes the issue.

bmwiedemann:
You can have a look at
https://rb.zq1.de/other/python-PyGithub/build-log-0c41ff94d.txt
in line 85 you see that it runs 13 months from now with `-rtc base=2025-11-26`
in line 656, you can see that only 0c41ff94d56c5962a53fe611a380c90d7460adca.patch gets applied
and in line 3552 you can see the failed assertion.

bmwiedemann:
Looking more closely at the test log, the problem seems to be that it tries to connect a live server, but our builds run without access to the internet. So maybe something wrong with the mocking?

EnricoMi:
Of course, using the token has to be guarded against real time as well: 6898f52

How is this now?

bmwiedemann:
Yes, that makes it pass.

EnricoMi (approved):
LGTM!

bmwiedemann:
Can we please get a new release well ahead of 2024-11-25 ?

EnricoMi:
Why is that?

bmwiedemann:
So that we (openSUSE and all other distributions) don't need to cherry-pick this patch onto our 2.4.0 package.

EnricoMi:
Do you execute unit tests in your pipeline?

bmwiedemann:
Yes: https://src.opensuse.org/rpm/python-PyGithub/src/branch/factory/python-PyGithub.spec#L73

EnricoMi:
Released as v2.5.0.

PR #3046, opened 2024-09-18
Rework GraphQL mutations

EnricoMi:
This contains the following breaking changes:
- use mutation name as given in Github GraphQL documentation
- remove capitalized mutation name from mutation request
- rename `variables` argument to `mutation_input`
- rename `output` argument to `output_schema`
- return response data according to output schema (drop outer data.mutationName path)


tiangolo (approved):
I took the liberty of reviewing this. I clonned the branch locally and followed the code line by line, and it all makes sense to me. LGTM! :rocket:

Issue #2718, opened 2023-08-14
Feature Request: Github Branch Rulesets


aclarknexient:
Github recently released a new feature called rulesets:

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets

Rulesets are available in the REST and GraphQL APIs:
* https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-all-repository-rulesets
* https://docs.github.com/en/graphql/reference/mutations#createrepositoryruleset

It's still a new feature but at some point in the future I'd like to add it to our automation scripts.

I haven't yet looked at the source of PyGithub to see how difficult it would be for me to add this. I'm not an expert Python coder but I can copy and paste REST API calls and edit names, so I _may_ be able to add this if no one else is able to. I didn't see any PRs or other Issues that mention rulesets. I hope it's OK that I created this ticket.

sfc-gh-hpathak:
Yeah. This is GA now. Would be great to have this, I might look into it when free.

mgagnon-wrk:
+1 for this. We require rulesets for specific functionality that the old branch protection rules cannot provide. Now we are forced to look at alternative libraries.

timor-raiman:
+1 This would support enabling and configuring mergequeue

EnricoMi:
That should be straightforward to implement. Happy to review this! Any volunteers?

wyardley:
I will see if I've got time; in the meantime, now that `requester` objects are public, does anyone have an example of using that to work with these?

wyardley:
Ended up with something like this
```python
                url = repo.url + f"/rules/branches/{DEFAULT_BRANCH}"
                response = branch.requester.requestJsonAndCheck("GET", url)

                pr_params = [x["parameters"] for x in response[1] if x["type"] == "pull_request"][0]
                if not pr_params:
                    [...]
                else:
                    num_required = pr_params["required_approving_review_count"]
                    if int(num_required) > 0:
                        [...]
```

sparsick:
It seems that already a PR to cover this feature exists: https://github.com/PyGithub/PyGithub/pull/3188

PR #3056, opened 2024-10-09
Make requester a public attribute

### LINKED ISSUE #2071: Provide a way to make a "raw" request
There will always take a while from when new endpoint or parameter is available in the GitHub API until it is supported by a released version of PyGithub.

Something as simple as providing a "public" accessor for __requester that is guaranteed not to change it's name would suffice.
Another possibility would be that the Requester object is, optionally, created by the user and then used to create the GitHub object.

minrk:
There will always be new APIs not yet supported by PyGitHub (e.g. #2718). This adds an escape hatch for users to send requests directly without having to start from scratch or use private APIs while waiting for PyGitHub to add support.

Changes: 

- private `object._requester` `.__requester` is now available via public `.requester` attribute on MainClass, GitHubObject, etc.
- add Requester APIs to docs as public methods, indicating status as stable public APIs

Fixes #2071 which received only positive feedback, but was closed for inactivity.


minrk:
docformatter hook can't run with pre-commit 4 due to https://github.com/PyCQA/docformatter/pull/287

pre-commit action cannot specify pre-commit version.
pre-commit action is also readily replaced with a single `run`:

```yaml
run: |
  pip install 'pre-commit<4'
  pre-commit run --all-files --show-diff-on-failure --color=always
```

(plus cache, if you want it)

EnricoMi (commented):
You could also test the requester property on one GithubObject implementation, e.g. Repository.

EnricoMi:
This looks reasonable to do.

EnricoMi (approved):
LGTM!

EnricoMi:
Thanks for the contribution!

flatherskevin:
When is this scheduled for release? (Disclaimer I couldn't find release practices elsewhere in the repo, so I figured I would comment directly here)

wyardley:
Seems like this was released in [2.5.0](https://github.com/PyGithub/PyGithub/releases/tag/v2.5.0)

Is this usable on the repo or org objects? I see `repo._requester` but not `repo.requester`?

Seems like a great feature - would be great if there were one or two examples in the docs. Is there a good example of this in use somewhere?

PR #3059, opened 2024-10-09
avoid pre-commit action in order to pin pre-commit

### LINKED ISSUE #3058: Fix pre-commit action
docformatter hook can't run with pre-commit 4 due to https://github.com/PyCQA/docformatter/pull/287

pre-commit action cannot specify pre-commit version.
pre-commit action is also readily replaced with a single `run`:

```yaml
run: |
  pip install 'pre-commit<4'
  pre-commit run --all-files --show-diff-on-failure --color=always
```

(plus cache, if you want it)

_Originally posted by @minrk in https://github.com/PyGithub/PyGithub/issues/3056#issuecomment-2401725893_

minrk:
mainly to pin pre-commit for now

pre-commit action doesn't save much, and relinquishes some control

fixes #3058

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -33,7 +33,17 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.x"
-      - uses: pre-commit/action@v3.0.0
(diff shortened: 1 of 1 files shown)


EnricoMi (approved):
LGTM!

PR #3060, opened 2024-10-10
Support downloading a Release Asset

### LINKED ISSUE #1541: Add ability to download binary content of release asset
I know #525 introduced release asset capabilities, but unless I've missed the right doc page I don't think it's possible to download the binary content of a release asset - just the information about an asset.  See https://developer.github.com/v3/repos/releases/#get-a-single-release-asset and the note about downloading the asset's binary content.

### LINKED ISSUE #2445: Download Release Asset?

### LINKED ISSUE #2918: Download release assets
The documentation says that I can download a release assets with:
- https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#get-a-release-asset

```
To download the asset's binary content, set the Accept header of the request to [application/octet-stream](https://docs.github.com/rest/overview/media-types). The API will either redirect the client to the location, or stream it directly if possible. API clients should handle both a 200 or 302 response.
```

Can I do it with this library?

neel-m:
Provide support for  downloading release assets.

Fixes #1541
Fixes #2445
Fixes #2918

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -1128,9 +1128,7 @@ class Requester:
             if (
                 status == 302
                 and "location" in responseHeaders
-                and (
-                    isinstance(original_cnx, (HTTPSRequestsConnectionClass, HTTPRequestsConnectionClass))
-                )
+                and (isinstance(original_cnx, (HTTPSRequestsConnectionClass, HTTPRequestsConnectionClass)))
             ):
                 location = responseHeaders["location"]
                 o = urllib.parse.urlparse(location)
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -857,6 +857,26 @@ class Requester:
                     raise
                 return {"data": data}
 
+    def requestFile(
+        self,
+        verb: str,
+        url: str,
+        headers: Optional[Dict[str, Any]] = {},
+        cnx=None, path=None):
+        """Request a file from the server and save it to the given path"""
+        headers["Accept"] = "application/octet-stream"
+        headers["User-Agent"] = self.__userAgent
+        if self.__auth is not None:
+            self.__auth.authentication(headers)
+
+        cnx = self.__createConnection()
+        cnx.request(verb, url, input, headers)
+        response = cnx.session.get(url, stream=True, headers=headers)
+        with open(path, "wb") as f:
+            for chunk in response.iter_content(chunk_size=1024 * 1024):
+                if chunk:
+                    f.write(chunk)
+
     def requestJson(
         self,
         verb: str,
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -98,6 +98,7 @@ from typing import (
     Dict,
     Generic,
     ItemsView,
+    Iterator,
     List,
     Optional,
     Tuple,
@@ -134,12 +135,19 @@ class RequestsResponse:
         self.status = r.status_code
         self.headers = r.headers
         self.text = r.text
+        self.response = r
 
     def getheaders(self) -> ItemsView[str, str]:
         return self.headers.items()
 
     def read(self) -> str:
         return self.text
+
+    def iter_content(self, chunk_size: Union[int, None] = 1) -> Iterator:
+        return self.response.iter_content(chunk_size=chunk_size)
+
+    def raise_for_status(self) -> None:
+        self.response.raise_for_status()
 
 
 class HTTPSRequestsConnectionClass:
@@ -190,11 +198,13 @@ class HTTPSRequestsConnectionClass:
         url: str,
         input: Optional[Union[str, io.BufferedReader]],
         headers: Dict[str, str],
+        stream: bool = False,
     ) -> None:
         self.verb = verb
         self.url = url
         self.input = input
         self.headers = headers
+        self.stream = stream
 
     def getresponse(self) -> RequestsResponse:
         verb = getattr(self.session, self.verb.lower())
@@ -253,11 +263,12 @@ class HTTPRequestsConnectionClass:
         )
         self.session.mount("http://", self.adapter)
 
-    def request(self, verb: str, url: str, input: None, headers: Dict[str, str]) -> None:
+    def request(self, verb: str, url: str, input: None, headers: Dict[str, str], stream: bool = False) -> None:
         self.verb = verb
         self.url = url
         self.input = input
         self.headers = headers
+        self.stream = stream
 
     def getresponse(self) -> RequestsResponse:
         verb = getattr(self.session, self.verb.lower())
@@ -868,24 +879,23 @@ class Requester:
         """
         GET a file from the server and save it to the given path.
         """
-        _, _, response = self.getStream(url, parameters, headers, cnx)
-        response.raise_for_status()
+        _, _, stream_chunk_iterator = self.getResponseStream(url, parameters, headers, cnx)
         with open(path, "wb") as f:
-            for chunk in response.iter_content():
+            for chunk in stream_chunk_iterator():
                 if chunk:
                     f.write(chunk)
 
-    def getStream(
+    def getResponseStream(
         self,
         url: str,
         parameters: Optional[Dict[str, Any]] = None,
         headers: Optional[Dict[str, str]] = None,
         cnx: Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = None,
-    ) -> Tuple[int, Dict[str, Any], requests.Response]:
+    ) -> Tuple[int, Dict[str, Any], Callable]:
         """
         GET a stream from the server.
 
-        :returns:``(status, headers, response)``
+        :returns:``(status, headers, stream_chunk_iterator)``
 
         """
         if headers is None:
@@ -898,9 +908,10 @@ class Requester:
         status, responseHeaders, output = self.__requestEncode(
             cnx, "GET", url, parameters, headers, None, encode, stream=True
         )
-        if isinstance(output, requests.Response):
-            return status, responseHeaders, output
-        raise ValueError("getStream() Expected a requests.Response object, should never happen")
+        if isinstance(output, RequestsResponse):
+            output.raise_for_status()
+            return status, responseHeaders, output.iter_content
+        raise ValueError("getStream() Expected a RequestsResponse object, should never happen")
 
     def requestJson(
         self,
@@ -1038,7 +1049,7 @@ class Requester:
         input: Optional[T],
         encode: Callable[[T], Tuple[str, Any]],
         stream: bool = False,
-    ) -> Tuple[int, Dict[str, Any], Union[str, requests.Response]]:
+    ) -> Tuple[int, Dict[str, Any], Union[str, RequestsResponse]]:
         assert verb in ["HEAD", "GET", "POST", "PATCH", "PUT", "DELETE"]
         if parameters is None:
             parameters = {}
@@ -1087,23 +1098,18 @@ class Requester:
         requestHeaders: Dict[str, str],
         input: Optional[Any],
         stream: bool = False,
-    ) -> Tuple[int, Dict[str, Any], Union[str, requests.Response]]:
+    ) -> Tuple[int, Dict[str, Any], Union[str, RequestsResponse]]:
         self.__deferRequest(verb)
 
         try:
             original_cnx = cnx
             if cnx is None:
                 cnx = self.__createConnection()
-            if stream:
-                response = cnx.session.get(url, stream=True, headers=requestHeaders)
-                status = response.status_code
-                responseHeaders = {k.lower(): v for k, v in response.headers.items()}
-            else:
-                cnx.request(verb, url, input, requestHeaders)
-                resp = cnx.getresponse()
-                output = resp.read()
-                status = resp.status
-                responseHeaders = {k.lower(): v for k, v in resp.getheaders()}
+            _ = cnx.request(verb, url, input, requestHeaders, stream)
+            resp = cnx.getresponse()
+            output = resp.read() if not stream else "stream"
+            status = resp.status
+            responseHeaders = {k.lower(): v for k, v in resp.headers.items()}
 
             if input:
                 if isinstance(input, IOBase):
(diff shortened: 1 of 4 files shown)


EnricoMi (changes requested):
I'd like to make the following design suggestions:
- Current `requestFile` bypasses existing request infrastructure, which contains a lot of features (error code handling, limit rate response headers, retry logic, ...). Please base `requestFile` on `__requestEncode`.
- Naturally, given you retrive the file content as a stream, you would like to provide the user with the same option. Please add a `requestStream` method, that then is used by `requestFile`.
- Please use a `chunk_size=None`, so that chunks are as sent over the wire, user code can buffer chunks if needed.
- I think `__requestEncode` and `__requestRaw` need an optional `stream: bool = False` parameter that will make return the content as a stream object rather than a `str`.
- We probably need to extend the test framework to record the HTTP chunks and replay them during testing methods that call into `Requester.requestFile`.
- Parameter `path` in method `download_asset` might be optional. With no `path` given it could return the stream.
- I would not return a `bool` in `download_asset` when `path` is given as this is of little use. If there is no exception, the file can be considered downloaded successfully.

neel-m:
> I'd like to make the following design suggestions:
> 
> * Current `requestFile` bypasses existing request infrastructure, which contains a lot of features (error code handling, limit rate response headers, retry logic, ...). Please base `requestFile` on `__requestEncode`.
> * Naturally, given you retrive the file content as a stream, you would like to provide the user with the same option. Please add a `requestStream` method, that then is used by `requestFile`.
> * Please use a `chunk_size=None`, so that chunks are as sent over the wire, user code can buffer chunks if needed.
> * I think `__requestEncode` and `__requestRaw` need an optional `stream: bool = False` parameter that will make return the content as a stream object rather than a `str`.
> * We probably need to extend the test framework to record the HTTP chunks and replay them during testing methods that call into `Requester.requestFile`.
> * Parameter `path` in method `download_asset` might be optional. With no `path` given it could return the stream.
> * I would return a `bool` in `download_asset` when `path` is given as this is of little use. If there is no exception, the file can be considered downloaded successfully.

I remembered why I made my original choice to bypass the existing request infrastructure.  The existing code expects to get something that looks like text, and it would have been messy to update it.  The existing code is also confusing.  `__requestEncode()` and `__requestRaw()` have no docstrings so the intent is not so clear.  It looks like the only thing that should be calling `__requestRaw()` is __requestEncode()`.  That appears to be true but the Authentication tests have a "mock" version that I also had to update.  It looks like the magic is handled in the combination of both the Raw and Encode versions.  The Encode version has an incomplete name (encode is for the data that is sent to the server and not what is retrieved, it seems).  This is legacy code and there are likely reasons for all of these choices, which probably made sense to someone and might still be good choices now (I don't claim to have enough knowledge to know for sure).

I made the decision to rename the new user-facing functions to: `getFile()` and `getStream()`.  As should be obvious from the names, `getFile()` requires a path to the file that it will create and `getStream()` returns a Response (that allows access to the stream, this is a little less obvious, but is probably the better answer).

Since the underlying `__requestEncode()` and `__requestRaw()` used to only return a str for the data, I modified them to return that or the new `requests.Response`.  That required updates to the typing hints (and some code) to convince it what I was doing was okay.

I am not sure what you mean/want for the test framework.

I didn't make the path optional for `download_asset()`, but that should be easy to add if you think it is still a good idea.

I will remove the bool return for `download_asset()`, since it is of no value with the current code.

EnricoMi:
> I assume you forgot a not in the sentence about returning a bool in download_asset()

right, updated my comment...

github/Requester.py
```diff
@@ -832,6 +832,51 @@ def __structuredFromJson(self, data: str) -> Any:
...
+    def getFile(
+        self,
+        url: str,
+        path: str,
+        parameters: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
+        cnx: Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = None,
+    ) -> None:
+        """
+        GET a file from the server and save it to the given path.
```
EnricoMi, line 844: Sounds like path is a directory, but path includes the file name.

github/Requester.py
```diff
@@ -832,6 +832,51 @@ def __structuredFromJson(self, data: str) -> Any:
...
+                if chunk:
+                    f.write(chunk)
+
+    def getStream(
+        self,
+        url: str,
+        parameters: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
+        cnx: Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = None,
+    ) -> Tuple[int, Dict[str, Any], requests.Response]:
```
EnricoMi, line 859: Can we return the `response.iter_content()` instead of `response`?

github/Requester.py
```diff
@@ -999,19 +1061,24 @@ def __requestRaw(
...
             if cnx is None:
                 cnx = self.__createConnection()
-            cnx.request(verb, url, input, requestHeaders)
-            response = cnx.getresponse()
-
-            status = response.status
-            responseHeaders = {k.lower(): v for k, v in response.getheaders()}
-            output = response.read()
+            if stream:
+                response = cnx.session.get(url, stream=True, headers=requestHeaders)
```
EnricoMi, line 1073: The `True` branch assumes `verb` to be `get`.

github/Requester.py
```diff
@@ -999,19 +1061,24 @@ def __requestRaw(
...
+                status = response.status_code
+                responseHeaders = {k.lower(): v for k, v in response.headers.items()}
+            else:
+                cnx.request(verb, url, input, requestHeaders)
+                resp = cnx.getresponse()
+                output = resp.read()
+                status = resp.status
+                responseHeaders = {k.lower(): v for k, v in resp.getheaders()}
```
EnricoMi, line 1081: I think we can simplify this:
```suggestion
            response = cnx.request(verb, url, input, requestHeaders, stream=stream)
            resp = cnx.getresponse()
            output = resp.iter_content() if stream else output = resp.read()
            status = resp.status_code
            responseHeaders = {k.lower(): v for k, v in resp.getheaders()}
```

This requires some modification to `HTTPRequestsConnectionClass` / `HTTPSRequestsConnectionClass` in https://github.com/PyGithub/PyGithub/blob/main/github/Requester.py.

EnricoMi:
> I am not sure what you mean/want for the test framework.

No worries, I'll be happy if this is manually tested and just works. I will add tests separately then.

github/Requester.py
neel-m: This is the definition of path (that I am using): https://en.wikipedia.org/wiki/Path_(computing)
path can mean the full path to a file (which is the usage I have in mind)
You are correct that path can also mean just the base path of a file, i.e., the parent directory.  Are you suggesting a different word for the parameter?

github/Requester.py
neel-m: Sure, that would mean that `getStream()` should also do `response.raise_for_status()` on the response it gets.

github/Requester.py
neel-m: Is that a question, or a statement?
Since the code in question is in `__requestRaw()` I suppose it could be reasonable to assume that it might be useful for some other verb.  But, all the other code in this library assumes that the returned content is text, so it would be a much bigger task to review it all and figure out what other changes would be necessary to make it work with other verbs and other encoding.

I also looked at the GitHub API and did not find anything other than Release Assets that can return a binary stream.  Updating a Release Asset does require using the `POST` verb, but that does not return a stream.

This is the reason I hard-coded the verb.  Release Assets only need `GET` and no additional data needs to be provided.

github/Requester.py
neel-m: Okay, I think I see what you are saying.  It would likely be more consistent between stream and not stream, so maybe easier to understand.

github/Requester.py
neel-m: This is now no longer hard-coding GET at this point.

github/Requester.py
EnricoMi: making this more clear in the doc string is sufficient

github/Requester.py
```diff
@@ -833,6 +844,51 @@ def __structuredFromJson(self, data: str) -> Any:
...
+                if chunk:
+                    f.write(chunk)
+
+    def getResponseStream(
+        self,
+        url: str,
+        parameters: Optional[Dict[str, Any]] = None,
+        headers: Optional[Dict[str, str]] = None,
+        cnx: Optional[Union[HTTPRequestsConnectionClass, HTTPSRequestsConnectionClass]] = None,
+    ) -> Tuple[int, Dict[str, Any], Callable]:
```
EnricoMi, line 870: I think we can be more specific here:
```suggestion
    ) -> Tuple[int, Dict[str, Any], Callable[Union[int, None], Iterator]:
```

github/Requester.py
```diff
@@ -833,6 +844,51 @@ def __structuredFromJson(self, data: str) -> Any:
...
+        """
+        GET a stream from the server.
+
+        :returns:``(status, headers, stream_chunk_iterator)``
```
EnricoMi, line 874: Here, `stream_chunk_iterator` is technically not true, it returns a function that provides the iterator.

If we'd wire `chunk_size` through (after the `cnx` argument) all the way from `download_asset`, we could get away with returning the iterator directly here. I think this would simplify the API. Plus, `chunk_size` is also meaningful for `getFile`. and `download_asset(path="...", chunk_size=...)`. Then, `stream_chunk_iterator` can stay here.

EnricoMi (changes requested):
Almost there!

I think we can replace returning the `Callback` with returning the `Iterator` when `chunk_size` is being wired through from the user API (`download_asset`). That way, the API is leaner for implementing methods like `download_asset`, while user get (optional) control over the chunk size.

github/Requester.py
```diff
@@ -833,6 +844,51 @@ def __structuredFromJson(self, data: str) -> Any:
...
+        """
+        GET a file from the server and save it to the given path.
+        """
+        _, _, stream_chunk_iterator = self.getResponseStream(url, parameters, headers, cnx)
+        with open(path, "wb") as f:
+            for chunk in stream_chunk_iterator():
+                if chunk:
+                    f.write(chunk)
+
+    def getResponseStream(
```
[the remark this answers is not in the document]
neel-m: okay, makes sense

github/Requester.py
```diff
@@ -1000,19 +1073,19 @@ def __requestRaw(
...
-            response = cnx.getresponse()
-
-            status = response.status
-            responseHeaders = {k.lower(): v for k, v in response.getheaders()}
-            output = response.read()
+            _ = cnx.request(verb, url, input, requestHeaders, stream)
+            resp = cnx.getresponse()
+            output = resp.read() if not stream else "stream"
+            status = resp.status
+            responseHeaders = {k.lower(): v for k, v in resp.headers.items()}
```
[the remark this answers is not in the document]
neel-m: I don't recall what I was thinking when I made the change, probably matched what I had in an earlier version, but no good reason for the shorter name.

Update __log() to handle this correctly

github/Requester.py
```diff
@@ -1049,9 +1122,8 @@ def __requestRaw(
                     )
                 if self._logger.isEnabledFor(logging.INFO):
                     self._logger.info(f"Following Github server redirection from {url} to {o.path}")
-                return self.__requestRaw(original_cnx, verb, o.path, requestHeaders, input)
-
-            return status, responseHeaders, output
+                return self.__requestRaw(original_cnx, verb, o.path, requestHeaders, input, stream=stream)
+            return status, responseHeaders, output if not stream else resp
```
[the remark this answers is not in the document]
neel-m: okay

github/Requester.py
neel-m: that didn't work, but I implemented the spirit of the comment

github/Requester.py
neel-m: this now works as you suggested

neel-m:
@EnricoMi please take a look at the changes.  It was challenging getting it to work with the unit tests, since they return 302 and that ran the changes I made (which are not compatible with the mocks).  I am not sure the way I handled it is the best way.

github/Requester.py
neel-m: okay

neel-m:
@EnricoMi when do you think you might have some time to review this PR?

github/Requester.py
```diff
@@ -1048,8 +1123,24 @@ def __requestRaw(
...
                 time.sleep(Consts.PROCESSING_202_WAIT_TIME)
-                return self.__requestRaw(original_cnx, verb, url, requestHeaders, input)
+                return self.__requestRaw(original_cnx, verb, url, requestHeaders, input, stream=stream)
 
+            if (
+                status == 302
+                and "location" in responseHeaders
+                and (
+                    isinstance(original_cnx, HTTPSRequestsConnectionClass)
+                    or isinstance(original_cnx, HTTPRequestsConnectionClass)
```
EnricoMi, line 1133:
```suggestion
                    isinstance(original_cnx, (HTTPSRequestsConnectionClass, HTTPRequestsConnectionClass))
```

EnricoMi (approved):
LGTM! Thanks for the streaming and download support!

PR #3065, opened 2024-10-21
Allow for private_key / sign function in AppAuth

### LINKED ISSUE #3064: Support "generic" JWT creation instead of explicit private key for AppAuth
Most of the APIs used by a GitHub app are accessible through the `GitHubIntegration` class ([example](https://pygithub.readthedocs.io/en/stable/examples/Authentication.html#app-authentication)) and the class only allows an `AppAuth` object for authentication ([reference](https://pygithub.readthedocs.io/en/stable/github_integration.html)) which in turn only allows an explicit private key as an input ([reference](https://pygithub.readthedocs.io/en/stable/utilities.html#github.Auth.AppAuth)).

However, the official GitHub documentation recommends the following:

> Consider storing your GitHub App's private key in a key vault [...] and making it sign-only.

Source: [Best practices for creating a GitHub App (Private Keys)](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/best-practices-for-creating-a-github-app#private-keys).

This is not possible with the current interfaces which all require an explicit private key.
Would it be possible considering an approach where the user passes a function as an authentication method for the `AppAuth` class as an alternative to the `private_key` parameter. This method is expected to return a JWT which could then be used for any subsequent processes (e.g. as alternative path of the `create_jwt` implementation, see [here](https://github.com/PyGithub/PyGithub/blob/c44ec523b1830c2d747fcf417b478562854664e8/github/Auth.py#L244))

EnricoMi:
Allows to use keys stored in key vaults providing sign-only access to keys.

Removes `AppAuth.private_key` and  `AppInstallationAuth.private_key`.

Fixes #3064.


github/Auth.py
```diff
@@ -190,36 +193,49 @@ class AppAuth(JWT):
...
         self._jwt_expiry = jwt_expiry
         self._jwt_issued_at = jwt_issued_at
-        self._jwt_algorithm = jwt_algorithm
 
     @property
     def app_id(self) -> Union[int, str]:
         return self._app_id
 
-    @property
-    def private_key(self) -> str:
```
webknjaz, line 220: This turned out to be a breaking change, but it's only documented in the features section of the change log…

PR #3067, opened 2024-10-24
Add support for make-latest to create_git_release and create_git_tag_and_release

SebastienSyd:
### The problem:

def "create_git_release" and "create_git_tag_and_release" can not create releases that are not the "latest", as by default they are set as "latest"

### How solving the problem:

adding optional parameter "make_latest", from original api doc: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release

similar to: https://github.com/PyGithub/PyGithub/pull/2888


github/Repository.py
```diff
@@ -1264,6 +1264,7 @@ def create_git_tag_and_release(
         draft: bool = False,
         prerelease: bool = False,
         generate_release_notes: bool = False,
+        make_latest: str = "true",
```
EnricoMi, line 1267: This should be a `bool`:
```suggestion
        make_latest: bool = True,
```

EnricoMi, line 1267: What is the current behaviour of Github with `make_latest` omitted?

tests/Repository.py
```diff
@@ -432,6 +432,7 @@ def testCreateGitReleaseWithAllArguments(self):
             True,
             False,
             "da9a285fd8b782461e56cba39ae8d2fa41ca7cdc",
+            "true",
         )
         self.assertEqual(release.tag_name, "vX.Y.Z-by-PyGithub-acctest2")
```
EnricoMi, line 437: Does `Release` have an attribute like `latest`?

EnricoMi (changes requested):

github/Repository.py
SebastienSyd: according to the official documentation, the default is "true"

tests/Repository.py
SebastienSyd: I haven't found any `latest` attribute sadly

github/Repository.py
EnricoMi: alright

EnricoMi (approved):
LGTM!

SebastienSyd:
@EnricoMi any news?

github/Repository.py
bebach: Why should make_latest be a string and not a bool?

bebach:
> @EnricoMi any news?

Good point. This would be mighty helpful for us :-)

PR #3078, opened 2024-11-06
remove COMMENT as default event for create_review

eddie-santos:
Fixes a discrepancy between PyGithub for creating a pull request review, compared to the Github API. When using the REST API, not specifying an `event` defaults to using `PENDING` mode, which creates a review and automatically adds the user as a reviewer. In the current PyGithub implementation, if no `event` parameter is given, it defaults to `COMMENT`, which does not add the user as a reviewer. It is not possible to use `event="PENDING"` via the API, it's just the default Github creates when `event` is sent empty.

https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request

### DIFF
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -491,7 +491,6 @@ class PullRequest(CompletableGithubObject):
         assert is_optional(event, str), event
         assert is_optional_list(comments, dict), comments
         post_parameters: dict[str, Any] = NotSet.remove_unset_items({"body": body})
-        post_parameters["event"] = "COMMENT" if is_undefined(event) else event
         if is_defined(commit):
             post_parameters["commit_id"] = commit.sha
         if is_defined(comments):


github/PullRequest.py
```diff
@@ -491,7 +491,6 @@ def create_review(
         assert is_optional(event, str), event
         assert is_optional_list(comments, dict), comments
         post_parameters: dict[str, Any] = NotSet.remove_unset_items({"body": body})
-        post_parameters["event"] = "COMMENT" if is_undefined(event) else event
```
EnricoMi, line 494: This will remove the `event` argument all together.

eddie-santos: Gotcha, so by adding it to `NotSet.remove_unset_items` it'll handle not passing `event` for us?

eddie-santos: That is, provided the caller doesn't use the `event` argument

EnricoMi: Yes, if `event` is not given by caller, `NotSet.remove_unset_items` will remove it from the dict.

github/PullRequest.py
```diff
@@ -490,8 +490,12 @@ def create_review(
         assert is_optional(body, str), body
         assert is_optional(event, str), event
         assert is_optional_list(comments, dict), comments
-        post_parameters: dict[str, Any] = NotSet.remove_unset_items({"body": body})
-        post_parameters["event"] = "COMMENT" if is_undefined(event) else event
+        post_parameters: dict[str, Any] = NotSet.remove_unset_items({
+            "body": body,
+            "event": event,
+            "commit_id": commit.sha if is_defined(commit) else None,
```
EnricoMi, line 496: I think this should read `NotSet` to work:
```suggestion
            "commit_id": commit.sha if is_defined(commit) else NotSet,
```

github/PullRequest.py
```diff
@@ -490,8 +490,12 @@ def create_review(
...
+            "comments": comments if is_defined(comments) else []
+        })
         if is_defined(commit):
```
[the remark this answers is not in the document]
eddie-santos: Done.

EnricoMi (approved):
LGTM!

EnricoMi:
I presume these tests fail because their test data in `tests/ReplayData/` contain the default `COMMENT` value in the request. Can you please edit those files so the tests pass?

PR #3138, opened 2025-01-08
Make Pickle test use recorded data

EnricoMi:
Otherwise fetching repo data might be defered by rate limit and CI is blocked.


EnricoMi:
This has been merged via #3137.

PR #3171, opened 2025-01-10
Make `conclusion` nullable in `WorkflowJob.py`

Apakottur:
The value of `conclusion` in the Workflow Job object can be NULL:
conclusion_nullable

GitHub doc reference:
https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28


EnricoMi (changes requested):
You need to change return type of

    def conclusion(self) -> str:

as well.

EnricoMi (approved):
LGTM!

Issue #3179, opened 2025-01-23
Change to Github API message returned with 404 leads to `UnknownObjectException` not being raised


coltonb-arcadia:
**Expected Behavior**
- When retrieving a file that is not found, I expect a `UnknownObjectException` to be raised.

**Actual Behavior**
- A GithubException is raised instead, because the error message is being explicitly checked here:

https://github.com/PyGithub/PyGithub/blob/f7d522495ee89ef00a5ed788479beaee277612bf/github/Requester.py#L818

The message is now verbose and references the missing object, instead of "not found".

This caused us to fail to catch an acceptable 404 and broke some things for us.

**Workaround**

```python
# Before
except UnknownObjectException:
    # handle error

# After
except GithubException as e:
    if e.status != 404:
        raise e
    # handle error
```

skinitimski:
I prepared a [PR](https://github.com/PyGithub/PyGithub/pull/3181) with a potential fix, but here is another workaround for anyone who wants to keep their existing exception handling. A good place to put this might be your top-level `__init__.py`:

```python
from github import Requester, UnknownObjectException

_createException = Requester.Requester.createException

def createException(cls, status, headers, output):
    message = output.get("message", "").lower() if output is not None else ""
    if status == 404 and "no object found" in message:
        return UnknownObjectException(status, output, headers)
    else:
        return _createException(status, headers, output)

Requester.Requester.createException = createException
```

GunalanD10:
string check seems costly to me : "no object found" in message:

@skinitimski  , can we only have 404 check ?

PR #3180, opened 2025-01-23
remove explicit 404 message check from createException

coltonb-arcadia:
Addresses #3179 

Not sure if test updates are required for this or where to make them.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -815,7 +815,7 @@ class Requester:
             exc = GithubException.BadUserAgentException
         elif status == 403 and cls.isRateLimitError(message):
             exc = GithubException.RateLimitExceededException
-        elif status == 404 and message == "not found":
+        elif status == 404:
             exc = GithubException.UnknownObjectException
 
         return exc(status, output, headers)


alejandro-angulo:
I'm not a maintainer myself but I did notice there is a test that specifies a message. Might be worth updating to better match what the API responds with now.

https://github.com/pygithub/PyGithub/blob/f23da453917a36c8bd48ab8d99e5fa7221884342/tests/Requester.py#L364-L373

endersonmenezes:
I believe that you can't simply remove the message, because the library tries to intercept and help the developer with various types of 404. I think it's important to include a treatment of: "No object found for the path <path>". If you do this directly in requests it is bad, we need this change in get contents.

skinitimski:
> I believe that you can't simply remove the message, because the library tries to intercept and help the developer with various types of 404. I think it's important to include a treatment of: "No object found for the path ".

:+1:

> If you do this directly in requests it is bad, we need this change in get contents.

This might be a better design as it keeps the handling for the specific exception closer to the code handling the response for that api endpoint. But, right now requester captures all 4XX responses in its global handling, and unwinding that seems like a lot.

I opened [another PR](https://github.com/PyGithub/PyGithub/pull/3181) that keeps the old check, and adds a new clause for the new message. Thus we preserve old behaviors without rearchitecting the 4XX handling. Also updated unit tests.

EnricoMi:
Merged #3181 in favour of this one.

PR #3175, opened 2025-01-21
Add `Repository.merge_upstream` method

Felixoid:
Allows to  merge a branch to upstream repository.

PR #3185, opened 2025-02-03
Use message of response in GithubException

EnricoMi:
The `message` extracted from response data could easily be used for the `message` parameter of the `GithubException`.


Felixoid (approved):
Nice, although it would be a race between this PR and https://github.com/PyGithub/PyGithub/pull/3175.

The new tests should be fixed there in advance, or fixed in this PR

Issue #3186, opened 2025-02-04
GitHub.get_organization wrong signature hint due to misnamed variable


blyedev:
I use signature hint and thus see the vars for each function i use and the login var was pretty confusing on this function given it needs the name of the org.

https://github.com/PyGithub/PyGithub/blob/d90323fac7e6004b097dd91f09668f285ce3f4d0/github/MainClass.py#L426-L433

PR #3187, opened 2025-02-04
Fix `Github.get_organization` signature hint #3186

blyedev:
related to issue #3186

### DIFF
--- a/github/MainClass.py
+++ b/github/MainClass.py
@@ -423,13 +423,13 @@ class Github:
             url_parameters["since"] = since
         return PaginatedList(github.NamedUser.NamedUser, self.__requester, "/users", url_parameters)
 
-    def get_organization(self, login: str) -> Organization:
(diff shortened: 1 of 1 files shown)


Felixoid (approved):

EnricoMi:
You are right, it is called `org` in the docs: https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#get-an-organization

EnricoMi (approved):
LGTM!

blyedev:
@EnricoMi it seems the docs build has failed, you might want to rerun it before merging

cyberious:
IMHO: This was a breaking change and, as such, should have been a major version bump. While calling without arguments is not an issue, we have a wrapper class built to have retry logic with a generic function being passed and using the *kwargs, which this change ended up breaking. However, I am happy to see consistency between API docs and code.

PR #3248, opened 2025-02-28
Allow SHAs when creating PR comments

tuchfarber:
Existing code requires a `Commit` to be passed into `PullRequest.create_review_comment` even though it only uses the SHA. Given an existing `PullRequest`, to pass a `Commit`, one would need to make another round trip via `Repository.get_commit(sha=sha)` just to pass in the larger object. 

This may need additional tests / work, but I was hoping to gauge interest in the idea before completing the rest. Thanks :)

### DIFF
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -448,9 +448,7 @@ class PullRequest(CompletableGithubObject):
         headers, data = self._requester.requestJsonAndCheck("GET", self.issue_url)
         return github.Issue.Issue(self._requester, headers, data, completed=True)
 
-    def create_comment(
-        self, body: str, commit: github.Commit.Commit | str, path: str, position: int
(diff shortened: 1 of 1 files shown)


github/PullRequest.py
```diff
@@ -467,12 +467,39 @@ def create_review_comment(
...
+            start_side=start_side,
+            in_reply_to=in_reply_to,
+            subject_type=subject_type,
+            as_suggestion=as_suggestion,
+        )
+
+    def create_review_comment_with_primatives(
+        self,
+        body: str,
+        commit: str,
```
EnricoMi, line 487: I'd rather go for one method with `commit: Commit | str`.

tuchfarber: Updated :)

EnricoMi (approved):
LGTM!

PR #3254, opened 2025-03-03
Use warnings.deprecated() decorator with Python 3.13+

lazka:
PyGitHub is using the "deprecated" Python package for deprecations until now, but was only using the provided deprecated() decorator from it, which in Python 3.13 has a stdlib equivalent with [warnings.deprecated()](https://docs.python.org/3/library/warnings.html#warnings.deprecated).

The stdlib variant has multiple advantages:

* Removes the dependency on the "deprecated" package and the transitive "wrapt" package (which is a C extension)
* Mypy can alert users of deprecated APIs since mypy 1.14 (opt-in for now), see https://mypy-lang.blogspot.com/2024/12/mypy-114-released.html
* Pylance in VSCode gives visual hints and warnings when using a deprecated function/class

The make everything still work with older Python we provide a wrapper that implements the new stdlib interface, and we only require the "deprecated" package with older Python. The wrapper requires the "extra_stacklevel" kwarg added in 1.2.15, so make that the minimum version. The wrapper exposed a bug in types-deprecated, which is tracked here: https://github.com/python/typeshed/pull/13573 which is why type checking is skipped in parts there.

For classes the stdlib variants just warns with the provided message while the package converts it to "Call to deprecated class Foo (\<msg\>)". Adjust the tests to use regex matching there, to cover both cases.

### DIFF
--- a/github/GithubObject.py
+++ b/github/GithubObject.py
@@ -53,6 +53,7 @@ from __future__ import annotations
 
 import email.utils
 import re
+import sys
 import typing
 from abc import ABC
 from datetime import datetime, timezone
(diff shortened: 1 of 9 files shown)


lazka:
The added complexity seemed worth it to me, since eventually the extra code+dependencies can be removed in some years, but I can understand if that's considered too much. I was playing around with the new 3.13 features and this seemed worth a shot.

Since this changes code paths for 3.13 I created #3253 for running 3.13 in CI, which should ideally be merged before this.

Here is how the integration looks like in VSCode:

Screenshot From 2025-03-02 08-10-11

github/GithubObject.py
```diff
@@ -74,6 +75,26 @@
 T_gh = typing.TypeVar("T_gh", bound="GithubObject")
 
 
+if sys.version_info < (3, 13):
```
EnricoMi, line 78: Maybe a better place for this would be `__init__`.

lazka:
I'm opening a new PR, since too many things have changed.

lazka:
Simpler variant -> #3255

PR #3255, opened 2025-03-03
Replace `deprecated.deprecated()` with `typing_extensions.deprecated()`

lazka:
PyGitHub is using the `deprecated` Python package for deprecations until now, but was only using the provided `deprecated()` decorator from it, which since Python 3.13 has a stdlib equivalent with `warnings.deprecated()`. See https://peps.python.org/pep-0702/ for details.

Luckily `typing_extensions`, which we already use, provide the same decorator for old Python versions (since 4.5), and just passes through the stdlib one when using Python 3.13+.

The new decorator has multiple advantages:

* Removes the dependency on the `deprecated` package and the transitive `wrapt` package (which is a C extension)
* Mypy can alert users of deprecated APIs since mypy 1.14 (opt-in), see https://mypy-lang.blogspot.com/2024/12/mypy-114-released.html
* Pylance in VSCode gives visual hints and warnings when using a deprecated function/class

There are some small differences between the decorators, which in our case only affect one test case, where the old one would prefix a class deprecation with "Call to deprecated class Foo" while the new one just uses the message as is. Also with the new one a message is required, so adjust the examples in the contribution guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,17 +55,17 @@ $ python scripts/add_attribute.py Commit url string
 
 ## Deprecation warning
 
-Before removing attributes/methods, consider adding deprecation warnings instead. The [Deprecated](https://github.com/tantale/deprecated) packages provides a handy decorator to add deprecation warnings with an optional reason.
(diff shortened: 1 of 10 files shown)


EnricoMi (approved):
LGTM! Thanks for looking into this, getting rid of that dependency is great.

lazka:
thanks!

PR #3256, opened 2025-03-04
Throw `UnknownObjectException` for missing branch

### LINKED ISSUE #3011: Wrong exception type raised when requesting non-existent branch.
If requesting a branch that does not exist, a generic `GithubException` is raised, however the proper exception to be raised is `UnknownObjectException`. Per the docstring:

```
class UnknownObjectException(GithubException):
    """
    Exception raised when a non-existing object is requested (when Github API replies with a 404 HTML status)
    """
```

The reason this happens is because the code that generates the exception based on the HTTP error code is not only looking at the status, but is also looking at the message it encapsulates. It only builds the `UnknownObjectException` if the message is "not found" but the 404 HTTP response when requesting an unknown branch has the message "Branch not found".
```
      elif status == 404 and message == "not found":
            exc = GithubException.UnknownObjectException
```

Is there any reason that we are checking the contents of the message and not just the HTTP status code? If there is, would it at least be better to chang `message == "not found"` to `"not found" in message`?

Minimal example:
```
>>> import github
>>> gh = github.Github("ACCESS_TOKEN_REDACTED")
>>> org = gh.get_organization("ORG_NAME_REDACTED")
>>> repo = org.get_repo("REPO_NAME_REDACTED")
>>> branch = repo.get_branch("foo")
  File "C:\Users\zhighleggett\AppData\Local\Programs\Python\Python312\Lib\site-packages\github\Repository.py", line 2018, in get_branch
    headers, data = self._requester.requestJsonAndCheck("GET", f"{self.url}/branches/{branch}")
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\zhighleggett\AppData\Local\Programs\Python\Python312\Lib\site-packages\github\Requester.py", line 548, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, self.__customConnection(url)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\zhighleggett\AppData\Local\Programs\Python\Python312\Lib\site-packages\github\Requester.py", line 609, in __check
    raise self.createException(status, responseHeaders, data)
github.GithubException.GithubException: 404 {"message": "Branch not found", "documentation_url": "https://docs.github.com/rest/branches/branches#get-a-branch", "status": "404"}
```

jaseemabid:
Fixes #3011

Ideally this code should just check for status code and not touch the
message at all, but I'm sticking with the pattern here.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -834,7 +834,7 @@ class Requester:
             exc = GithubException.BadUserAgentException
         elif status == 403 and cls.isRateLimitError(lc_message):
             exc = GithubException.RateLimitExceededException
-        elif status == 404 and (lc_message == "not found" or "no object found" in lc_message):
+        elif status == 404 and ("not found" in lc_message or "no object found" in lc_message):
             exc = GithubException.UnknownObjectException
         else:
             # for general GithubException, provide the actual message
--- a/tests/PullRequest.py
+++ b/tests/PullRequest.py
@@ -532,9 +532,8 @@ class PullRequest(Framework.TestCase):
         status = self.delete_restore_pull.merge(delete_branch=True)
         self.assertTrue(status.merged)
         self.assertTrue(self.delete_restore_pull.is_merged())
-        with self.assertRaises(github.GithubException) as raisedexp:
+        with self.assertRaises(github.UnknownObjectException) as raisedexp:
             self.delete_restore_repo.get_branch(self.delete_restore_pull.head.ref)
-        self.assertEqual(raisedexp.exception.message, "Branch not found")
         self.assertEqual(
             raisedexp.exception.data,
             {
@@ -544,9 +543,8 @@ class PullRequest(Framework.TestCase):
         )
 
     def testRestoreBranch(self):
-        with self.assertRaises(github.GithubException) as raisedexp:
+        with self.assertRaises(github.UnknownObjectException) as raisedexp:
             self.delete_restore_repo.get_branch(self.delete_restore_pull.head.ref)
-        self.assertEqual(raisedexp.exception.message, "Branch not found")
         self.assertEqual(raisedexp.exception.status, 404)
         self.assertEqual(
             raisedexp.exception.data,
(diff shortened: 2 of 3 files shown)


github/Requester.py
```diff
@@ -834,7 +834,11 @@ def createException(
             exc = GithubException.BadUserAgentException
         elif status == 403 and cls.isRateLimitError(lc_message):
             exc = GithubException.RateLimitExceededException
-        elif status == 404 and (lc_message == "not found" or "no object found" in lc_message):
+        elif status == 404 and (
+            lc_message == "not found"  #
+            or lc_message == "branch not found"
```
EnricoMi, line 839: what about
```suggestion
            or " not found" in lc_message
```

EnricoMi (commented):
Can you add a test case that covers the particular error message so we have real record data?

github/Requester.py
jaseemabid: @EnricoMi Should I remove the check for message completely? This seems unnecessary if status code is 404. 

Can add tests 👍

jaseemabid:
@EnricoMi I've updated the PR and tests to reflect the new exception:

1. Replaced base `github.GithubException` with the more specific `github.UnknownObjectException`

2. Removed message check `self.assertEqual(raisedexp.exception.message, "Branch not found")` in tests

`createException()` doesn't pass a message to any specific exceptions, so this change is in line with other errors like `BadCredentialsException` or `RateLimitExceededException` which also don't have a message field set.

jaseemabid:
@EnricoMi Is this alright or do you need any additional changes?

jaseemabid:
@EnricoMi  Bump? 👋

EnricoMi:
Fixed in #3299.

PR #3259, opened 2025-03-09
Add `PullRequest.get_issue_timeline` method

kukarkinmm:
Currently there is already `get_issue_events` method, but `get_issue_timeline` is missing. This means users have to use two requests instead of one: `as_issue().get_timeline()`. Seems to be suboptimal.

This pull request adds `PullRequest.get_issue_timeline()` in the most straightforward way.


EnricoMi (approved):
LGTM!

PR #3260, opened 2025-03-10
Support built-in `reversed()` on `PaginatedList`

mfocko:
Since `PaginatedList` can already be used as in iterator and there is an interface defined for reversing the list, it makes sense to also implement the `__reversed__` “dunder” method that is utilized by `reversed()` built-in function for reversing iterators.


mfocko:
> This is a great addition!
> 
> However, before this can be merged, we need to add unit tests to https://github.com/PyGithub/PyGithub/blob/main/tests/PaginatedList.py to guarantee that this works now and will keep working in the future.

I will have a look and add the tests… though as long as it depends on the `.reversed` property, which should be already tested, it is fine

mfocko:
I think the text file with HTTP requests could've been symlinked to the previous test as it is supposed to be the same (should have 1:1 requests, just different way of achieving from the user-perspective), though… it would mean that changing that one test, sometime in future, could lead to unexpected failure of the new test for built-in `reversed()`…

Don't think it should be a problem, since git will store it only once.

ruancomelli (approved):
Looking great now - thanks for the extra tests!

> I think the text file with HTTP requests could've been symlinked to the previous test as it is supposed to be the same (should have 1:1 requests, just different way of achieving from the user-perspective), though… it would mean that changing that one test, sometime in future, could lead to unexpected failure of the new test for built-in reversed()…
> 
> Don't think it should be a problem, since git will store it only once.

Yeah, I don't think that should be a problem; and we can always revisit this later if necessary.

EnricoMi:
> I think the text file with HTTP requests could've been symlinked to the previous test

Symlinks don't work well with Windows, and we are running these tests on Windows.

I have extended the `Framework` to allow tests reference a different test data file explicitly.

I think this will be useful in other places.

EnricoMi (approved):
LGTM!

tests/Framework.py
```diff
@@ -418,14 +419,26 @@ def ignoreWarning(self, category=Warning, module=""):
...
+        fileName = None
+        if self.__customFilename:
+            fileName = self.__customFilename
+        else:
+            for _, _, functionName, _ in traceback.extract_stack():
+                if functionName.startswith("test") or functionName == "setUp" or functionName == "tearDown":
+                    # because in class Hook(Framework.TestCase), method testTest calls Hook.test
+                    if functionName != "test":
+                        fileName = f"{self.__class__.__name__}.{functionName}.txt"
+        fileName = os.path.join(self.replayDataFolder, fileName) if fileName else None
```
Prateikx, line 441: >_possible bug_

**✨ AI Suggestion:** The code attempts to join `self.replayDataFolder` with `fileName` even when `fileName` might be `None`. If the loop in lines 436-440 does not assign a value to `fileName`, it remains `None` (from line 432). Passing `None` to `os.path.join` will raise a `TypeError`. The conditional assignment `if fileName else None` does not prevent this, as `os.path.join` is called regardless within the conditional expression's 'true' branch.
```suggestion
                                        fileName = f"{self.__class__.__name__}.{functionName}.txt"
                        fileName = os.path.join(self.replayDataFolder, fileName) if fileName is not None else None
```

Prateikx, line 441: >_possible bug_

**✨ AI Suggestion:** The variable `fileName` might be `None` if `self.__customFilename` is not set and the loop over `traceback.extract_stack()` does not find a suitable function name. Calling `os.path.join(self.replayDataFolder, None)` will raise a `TypeError`. Add a check to ensure `fileName` is not `None` before joining the path.
```suggestion
                        fileName = f"{self.__class__.__name__}.{functionName}.txt"
        fileName = os.path.join(self.replayDataFolder, fileName) if fileName is not None else None
```

Prateikx, line 441: >_possible bug_

**✨ AI Suggestion:** The variable `fileName` might be `None` if the logic in lines 436-440 does not find a suitable function name in the call stack. Calling `os.path.join` with `None` as the second argument will raise a `TypeError`. Add a check to ensure `fileName` is not `None` before joining the path.
```suggestion
                                            fileName = f"{self.__class__.__name__}.{functionName}.txt"
                        fileName = os.path.join(self.replayDataFolder, fileName) if fileName is not None else None
```

tests/PaginatedList.py
```diff
@@ -97,6 +98,14 @@ def testReversedIterationSupportsIterator(self):
             return
         self.fail("empty iterator")
 
+    def testReversedIterationSupportsBuiltinReversed(self):
+        # reuse test data of testReversedIterationSupportsIterator
+        with self.replayData("PaginatedList.testReversedIterationSupportsIterator.txt"):
+            for i in reversed(self.list):
+                self.assertEqual(i.id, 4286936)
+                return
+            self.fail("empty iterator")
```
Prateikx, line 107: >_possible bug_
**✨ AI Suggestion:** The test `testReversedIterationSupportsBuiltinReversed` uses the same replay data as `testReversedIterationSupportsIterator`, but asserts on a specific value. This makes the test brittle and dependent on the data used in the other test. It's better to either use separate replay data or assert on the behavior of the reversed iterator, not the specific values returned.

Prateikx, line 107: >_possible bug_

**✨ AI Suggestion:** The test `testReversedIterationSupportsBuiltinReversed` uses the same replay data as `testReversedIterationSupportsIterator`, but asserts on a specific value. This makes the test brittle and dependent on the data used in the other test. It's better to either use separate replay data or assert on the behavior of the reversed iterator, not the specific values.

Prateikx, line 107: >_possible bug_

**✨ AI Suggestion:** The test `testReversedIterationSupportsBuiltinReversed` uses the same replay data as `testReversedIterationSupportsIterator`, but asserts on a specific value. This makes the test brittle and dependent on the data used in the other test. It's better to either use separate replay data or assert on the behavior of the reversed iterator, not specific values.
```suggestion
    def testReversedIterationSupportsBuiltinReversed(self):
            with self.replayData("PaginatedList.testReversedIterationSupportsBuiltinReversed.txt"):
                reversed_list = list(reversed(self.list))
                self.assertEqual(len(reversed_list), len(self.list))
                for i, reversed_i in zip(self.list, reversed(reversed_list)):
                    self.assertEqual(i.id, reversed_i.id)
```

Prateikx, line 107: >_possible bug_

**✨ AI Suggestion:** The test `testReversedIterationSupportsBuiltinReversed` uses the same replay data as `testReversedIterationSupportsIterator`, but asserts on a specific value. This makes the test brittle and dependent on the data used in the other test. It's better to either use separate replay data or assert on the behavior of the reversed iterator, not specific values.
```suggestion
    def testReversedIterationSupportsBuiltinReversed(self):
            with self.replayData("PaginatedList.testReversedIterationSupportsBuiltinReversed.txt"):
                reversed_list = list(reversed(self.list))
                self.assertEqual(reversed_list, list(self.list.reversed))
                self.assertNotEqual(reversed_list, list(self.list))
```

Issue #3214, opened 2025-02-18
[BUG] Illegal `iss` integer type in generated JWTs


webknjaz:
I was passing an `int` to `Auth` as per its typing annotations, which I didn't initially realize was cascading into a problem with the JWT RFC violation.

This was when I was writing some tests that supply a legit RSA keypair and exercise [`create_jwt()`](https://github.com/PyGithub/PyGithub/blob/038624c/github/Auth.py#L240-L242).

I then attempt to decode the payload with signature verification. I attempted `jwt.decode(..., issuer=123)` / `jwt.decode(..., issuer='123')` and hit errors in both cases. So I troubleshooted the problem and discovered that PyGitHub uses App ID as an integer value when constructing JWT payload. But the JWT spec is pretty clear about the expectation that it must be a string:
> [4.1.1](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1).  "iss" (Issuer) Claim
>
>   The "iss" (issuer) claim identifies the principal that issued the
>   JWT.  The processing of this claim is generally application specific.
>   The "iss" value is **a case-sensitive string** containing a StringOrURI
>   value.  Use of this claim is OPTIONAL.

*(emphasis mine)*

So the bug in PyGitHub is that it doesn't turn the number into a string, and the bug in PyJWT is that it allows it in the first place. I also filed the issue over there too: https://github.com/jpadilla/pyjwt/issues/1039.

GitHub's own docs correctly show it as being a string as well: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app#example-using-python-to-generate-a-jwt.

Implementing this will also allow supporting Client ID issuer values: https://github.com/PyGithub/PyGithub/issues/3213.

webknjaz:
This should be easily fixable by wrapping https://github.com/PyGithub/PyGithub/blob/038624c/github/Auth.py#L276C20-L276C32 with `str()`.

And in addition to that, `jwt.decode()` invocations in tests at https://github.com/PyGithub/PyGithub/blob/038624c/tests/Authentication.py#L300-L305 and https://github.com/PyGithub/PyGithub/blob/038624c/tests/Authentication.py#L315-L320 can be improved with the built-in issuer verification that PyJWT provides.

webknjaz:
[...]

PR #3272, opened 2025-03-27
Normalize App ID to String & Enhance JWT Issuer Verification

x612skm:
This PR fixes PyGithub/PyGithub#3214 by ensuring self._app_id is always a string in AppAuth, preventing type mismatches in the JWT "iss" claim. It also strengthens JWT tests by adding issuer validation during decoding.

The AppAuth class previously stored self._app_id as-is (either int or str, per Union[int, str]), which could result in the JWT "iss" claim being an integer. Per the JWT specification ([RFC 7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)), the iss (issuer) claim should be a string, and GitHub’s authentication flow expects it as such. This type mismatch caused potential failures in downstream processing or validation (e.g., with PyJWT or GitHub’s API), as reported in #3214. Additionally, the tests lacked proactive issuer checks, relying only on payload comparison, which missed an opportunity to catch errors earlier.

**Changes**

- String Normalization in AppAuth (github/Auth.py): `Changed self._app_id = app_id to self._app_id = str(app_id)` in the initializer. This ensures the "iss" claim is consistently a string, regardless of whether the input app_id is an integer (common for GitHub App IDs) or a string.
- Test Improvements (tests/Authentication.py): Updated testCreateJWT and testCreateJWTWithExpiration to expect a string "iss" and added `issuer=str(APP_ID)` to jwt.decode(). This leverages PyJWT’s built-in issuer validation (introduced in PyJWT 1.0.0+), aligning with its security features and failing fast on issuer mismatches.

**Fixes**
✅ Fixes type mismatch in JWT "iss" claim.
✅ Enhances test robustness by failing early on incorrect issuers.

**Impact**
Fixes the root cause of type-related errors in JWT generation, ensuring compliance with RFC 7519 and GitHub’s expectations.
Strengthens tests by mirroring real-world validation (e.g., GitHub rejecting tokens with invalid iss), reducing the risk of regressions.

**Testing**
Verified with pytest tests/Authentication.py—both testCreateJWT and testCreateJWTWithExpiration pass with the updated expectations and issuer verification.

Fixes: [PyGithub/PyGithub#3214](https://github.com/PyGithub/PyGithub/issues/3214)


tests/Authentication.py
```diff
@@ -302,8 +302,9 @@ def testCreateJWT(self):
             key=PUBLIC_KEY,
             algorithms=["RS256"],
             options={"verify_exp": False},
+            issuer=str(APP_ID),
```
Prateikx, line 305: **Suggestion:** The tests now explicitly check for string issuer, but the assertion could be simplified. Instead of asserting the entire payload dictionary, focus solely on the "iss" claim to improve readability and maintainability.
```suggestion
self.assertEqual(payload["iss"], str(APP_ID))

    def testCreateJWTWithExpiration(self):
```

tests/Authentication.py
```diff
@@ -317,8 +318,9 @@ def testCreateJWTWithExpiration(self):
...
```
Prateikx, line 321: **Suggestion:** Similar to the previous suggestion, simplify the assertion to focus only on the "iss" claim for better clarity and maintainability.
```suggestion
self.assertEqual(payload["iss"], str(APP_ID))

    def testUserAgent(self):
```

x612skm:
Can someone take a look at this?

PR #3275, opened 2025-04-04
Point deprecation warnings to the caller code rather than inner class

xmo-odoo:
By default `warnings.warn` will point at the location of the `warn` call itself, and provides no stacktrace information, which can make it much harder to find the relevant application code.

By setting the stacklevel to 2, `warnings` will provide a link (file and line number) to the *caller* of `Github()`, which is the location to fix.


xmo-odoo:
Done, and rebased while at it.

EnricoMi (approved):
LGTM!

PR #3074, opened 2024-10-30
Adding very basic support for GitHub packages

Henkhogan:
I found that there is no feature available to query packages like docker images thus I tried to add the ability to query them using the same code patterns that I found in the repo.

This PR only covers a small fraction of the Package APIs: https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28

PR #3292, opened 2025-05-10
Add complete Package API implementation

hboyd2003:
I finished the Package API implementation that [@Henkhogan](https://github.com/Henkhogan) did in #3074.
Given that he stopped responding to the PR and has had no activity for the last few months, I decided to open up this PR.

All of @EnricoMi uggestions on the original PR have been addressed, and the implementation is now complete.
In the original implementation, [@Henkhogan](https://github.com/Henkhogan) used `StrEnum`, which I found useful and have continued to support as properties. However, since `StrEnum` was introduced in Python 3.11, we may want to consider removing it to maintain compatibility.


github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+        return github.Package.Package(self._requester, headers, data, completed=True)
+
+
+    def delete_package(self, package_type: str, package_name: str) -> None:
+        """
+        :calls: `DELETE /user/packages/{package_type}/{package_name} <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#delete-a-package-for-the-authenticated-user>`_
+        """
+        assert isinstance(package_type, str), package_type
+        assert isinstance(package_name, str), package_name
+        self._requester.requestJsonAndCheck("DELETE", f"{self.url}/packages/{package_type}/{package_name}")
```
EnricoMi: Please move to `Package.delete`.

github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+
+    def list_package_versions(self, package_type: str, package_name: str) -> PaginatedList[PackageVersion]:
+        """
+        :calls: `GET /user/packages/{package_type}/{package_name}/versions <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#list-package-versions-for-a-package-owned-by-the-authenticated-user`_
+        """
+        assert isinstance(package_type, str), package_type
+        assert isinstance(package_name, str), package_name
+        return PaginatedList(PackageVersion, self._requester, f"{self.url}/packages/{package_type}/{package_name}/versions", None)
+
+    def get_package_version(self, package_type: str, package_name: str, package_version_id: int) -> PackageVersion:
```
EnricoMi: Please move to `Package.get_version`.

github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+        """
+        :calls: `GET /user/packages/{package_type}/{package_name}/versions/{package_version_id} <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#get-a-package-version-for-the-authenticated-user>`_
+        """
+        assert isinstance(package_type, str), package_type
+        assert isinstance(package_name, str), package_name
+        assert isinstance(package_version_id, int), package_version_id
+        headers, data = self._requester.requestJsonAndCheck("GET", f"{self.url}/packages/{package_type}/{package_name}/versions/{package_version_id}")
+        return PackageVersion(self._requester, headers, data, completed=True)
+
+    def delete_package_version(self, package_type: str, package_name: str, package_version_id: int) -> None:
```
EnricoMi: Please move to `PackageVersion.delete`.

github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+        return PaginatedList(github.Package.Package, self._requester, f"{self.url}/packages", url_parameters)
+
+    def get_package(self, package_type: github.Package.PackageType, package_name: str) -> Package:
+        """
+        :calls: `GET /user/packages/{package_type}/{package_type} <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#get-a-package-for-the-authenticated-user>`_
+        """
+        assert isinstance(package_type , github.Package.PackageType), package_type
+        assert isinstance(package_name, str), package_name
+        headers, data = self._requester.requestJsonAndCheck("GET", f"{self.url}/packages/{package_type.value}/{package_name}")
+        return github.Package.Package(self._requester, headers, data, completed=True)
```
EnricoMi: With
```suggestion
        url = f"{self.url}/packages/{package_type.value}/{package_name}"
        return github.Package.Package(self._requester, {}, {"url": url}, completed=False)
```
we can make this return a lazy `Package` instance which will execute the `GET` when the first attribute is accessed. This allows for `AuthenticatedUser.get_package(...).delete()` without having a redundant `GET`.

github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+
+    def delete_package(self, package_type: str, package_name: str) -> None:
+        """
+        :calls: `DELETE /user/packages/{package_type}/{package_name} <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#delete-a-package-for-the-authenticated-user>`_
+        """
+        assert isinstance(package_type, str), package_type
+        assert isinstance(package_name, str), package_name
+        self._requester.requestJsonAndCheck("DELETE", f"{self.url}/packages/{package_type}/{package_name}")
+
+    def restore_package(self, package_type: str, package_name: str) -> None:
```
EnricoMi: Could be moved to `Package.restore`. This requires lazy-ness of `get_package` above.

github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+    def delete_package_version(self, package_type: str, package_name: str, package_version_id: int) -> None:
+        """
+        :calls: `DELETE /user/packages/{package_type}/{package_name}/versions/{package_version_id} <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#delete-a-package-version-for-the-authenticated-user>`_
+        """
+        assert isinstance(package_type, str), package_type
+        assert isinstance(package_name, str), package_name
+        assert isinstance(package_version_id, int), package_version_id
+        self._requester.requestJsonAndCheck("DELETE", f"{self.url}/packages/{package_type}/{package_name}/versions/{package_version_id}")
+
+    def restore_package_version(self, package_type: str, package_name: str, package_version_id: int) -> None:
```
EnricoMi: Could be moved to `PackageVersion.restore`. This requires lazy-ness of `Package.get_version`.

github/AuthenticatedUser.py
```diff
@@ -856,6 +861,85 @@ def get_orgs(self) -> PaginatedList[Organization]:
...
+
+    def get_package_version(self, package_type: str, package_name: str, package_version_id: int) -> PackageVersion:
+        """
+        :calls: `GET /user/packages/{package_type}/{package_name}/versions/{package_version_id} <https://docs.github.com/en/rest/packages/packages?apiVersion=latest#get-a-package-version-for-the-authenticated-user>`_
+        """
+        assert isinstance(package_type, str), package_type
+        assert isinstance(package_name, str), package_name
+        assert isinstance(package_version_id, int), package_version_id
+        headers, data = self._requester.requestJsonAndCheck("GET", f"{self.url}/packages/{package_type}/{package_name}/versions/{package_version_id}")
+        return PackageVersion(self._requester, headers, data, completed=True)
```
EnricoMi:
```suggestion
        url = f"{self.url}/packages/{package_type}/{package_name}/versions/{package_version_id}"
        return PackageVersion(self._requester, {}, {"url": url}, completed=False)
```

github/AuthenticatedUser.py
hboyd2003: Package already includes a delete method. I added this here in the event that a user wants to delete a package without a package object. Should it just be removed?

This is true for all the methods you want me to move.

hboyd2003: I've read you other comments now. I believe they can just be removed in that case.

hboyd2003:
@EnricoMi I have made the changes you have requested aswell as rebased to the current upstream please review.

PR #3295, opened 2025-05-22
Get collaborator role name

jmgate:
This commits adds a duplicate of `get_collaborator_permission()`, which, instead of returning `data["permission"]`, returns `data["role_name"]`. The former only returns "admin", "write", "read", or "none"; the latter now also tells us whether the collaborator has the "maintain" or "triage" role.

I'm using PyGitHub to monitor ~1K repositories under my company's GitHub organization, to ensure they stay in compliance with our policies, e.g., keeping track of the roles of outside collaborators, and making sure they stay at "triage" or below.  It's possible to achieve what we need to using `get_collaborators()` with some combination of `affiliation` and `permission`, but the addition of this new function means making significantly fewer API calls and speeding up our monitoring CI jobs by a factor of 2-3.

I realize this is a brute-force addition.  If you're happy to accept it as-is, great.  If we need to clean things up such that there's less duplication across the two functions, or clearer docstring to differentiate the two, etc., just let me know.


EnricoMi:
Ideally, `get_collaborator_permission` should return the entire `data` as some `GithubObject`, so calling site has access to `permission` and `role_name` attributes. But I can't think of a way that doesn't break `get_collaborator_permission` return type. This is the right name for the method returning the object instead of a string.

jmgate:
I'm afraid there's not an ideal solution, given the GitHub API.  `get_collaborator_permission` is hitting the `permission` endpoint, but that returns a data structure, which then has a `permission` attribute that this method should return for it to make sense from a user perspective.  E.g., you wouldn't want the user to have to do something like

```python
permission = repo.get_collaborator_permission("user")["permission"]
#                                  ^^^^^^^^^^          ^^^^^^^^^^
# Why do I need to ask for the permission twice?
```

That said, we could create a `_get_collaborator_permission_data()` method to hit the endpoint, and then do something along the lines of

```python
def get_collaborator_permission(...):
    return _get_collaborator_permission_data(...)["permission"]
def get_collaborator_role_name(...):
    return _get_collaborator_permission_data(...)["role_name"]
```

jmgate:
@EnricoMi, would you like me to change up this PR such that there's just a `get_collaborator_permission_data()` method, and then the user can grab whatever details they like from the resulting data structure?

EnricoMi:
You are right, this is difficult to resolve, given Githubs API response. This can be fixed separately, lets keep this simple for your additional use case.

EnricoMi:
Can you please add a unit test, similar to `Repository.testCollaboratorPermission`.

You can copy existing replay data `tests/ReplayData/Repository.testCollaboratorPermission.txt` and add the expected `role_name` attribute to the test data.

jmgate:
@EnricoMi, is the test and corresponding replay data in 6e0913f sufficient?

EnricoMi (approved):
LGTM!

github/Repository.py
```diff
@@ -1301,6 +1301,23 @@ def get_collaborator_permission(self, collaborator: str | NamedUser) -> str:
...
         return data["permission"]
 
+    def get_collaborator_role_name(self, collaborator: str | NamedUser) -> str:
+        """
+        :calls: `GET /repos/{owner}/{repo}/collaborators/{username}/permission <https://docs.github.com/en/rest/reference/repos#collaborators>`_
+        :param collaborator: string or :class:`github.NamedUser.NamedUser`
+        :rtype: string
+        """
+        assert isinstance(collaborator, NamedUser) or isinstance(collaborator, str), collaborator
+        if isinstance(collaborator, NamedUser):
```
EnricoMi:
```suggestion
        assert isinstance(collaborator, (github.NamedUser.NamedUser, str)), collaborator
        if isinstance(collaborator, github.NamedUser.NamedUser):
```

EnricoMi (changes requested):

jmgate:
Thanks so much for your help getting these changes in, @EnricoMi 😀

PR #3246, opened 2025-02-24
Adding `prevent_self_review` property to `Repository.createEnvironment`

gopidesupavan:
Currently create environment api supports `prevent_self_review`, adding missing property to support in PyGithub.

https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment

as per response from the endpoint looks like this, `prevent_self_review` falls under protection_rules, updated accordingly to the `EnvironmentProtectionRule`

```
{
    "id": 5687891617,
    "node_id": "EN_kwDON9Ex-M8AAAABUwZWoQ",
    "name": "test_two",
    "url": "https://api.github.com/repos/gopidesupavan/test-dependaboat-configs/environments/test_two",
    "html_url": "https://github.com/gopidesupavan/test-dependaboat-configs/deployments/activity_log?environments_filter=test_two",
    "created_at": "2025-02-24T20:40:56Z",
    "updated_at": "2025-02-24T20:40:56Z",
    "can_admins_bypass": true,
    "protection_rules": [
        {
            "id": 29234176,
            "node_id": "GA_kwDON9Ex-M4BvhQA",
            "type": "required_reviewers",
            "prevent_self_review": true,
            "reviewers": [
                {
                    "type": "User",
                    "reviewer": {
                        "login": "gopidesupavan",
                        "id": 31437079,
                        "node_id": "MDQ6VXNlcjMxNDM3MDc5",
                        "avatar_url": "https://avatars.githubusercontent.com/u/31437079?v=4",
                        "gravatar_id": "",
                        "url": "https://api.github.com/users/gopidesupavan",
                        "html_url": "https://github.com/gopidesupavan",
                        "followers_url": "https://api.github.com/users/gopidesupavan/followers",
                        "following_url": "https://api.github.com/users/gopidesupavan/following{/other_user}",
                        "gists_url": "https://api.github.com/users/gopidesupavan/gists{/gist_id}",
                        "starred_url": "https://api.github.com/users/gopidesupavan/starred{/owner}{/repo}",
                        "subscriptions_url": "https://api.github.com/users/gopidesupavan/subscriptions",
                        "organizations_url": "https://api.github.com/users/gopidesupavan/orgs",
                        "repos_url": "https://api.github.com/users/gopidesupavan/repos",
                        "events_url": "https://api.github.com/users/gopidesupavan/events{/privacy}",
                        "received_events_url": "https://api.github.com/users/gopidesupavan/received_events",
                        "type": "User",
[...]

PR #3312, opened 2025-07-08
added support for specifying `prevent_self_review` when creating environments

icebreaker:
This has been previously requested/discussed in the following discussion:

https://github.com/PyGithub/PyGithub/discussions/2889


EnricoMi:
Fixed in #3246.

PR #3313, opened 2025-07-09
Fix side-effect when removing Authorization key from headers

### LINKED ISSUE #3302: Asset download from a public repo fails without auth
The Release Asset download feature added in #3060 is currently broken for public repositories.

Steps to reproduce:

```python
from github import Github

github = Github()
public_repo = github.get_repo("stellarcarbon/sorocarbon")
latest_release = public_repo.get_latest_release()
asset = latest_release.assets[0]
asset.download_asset(f"/tmp/{asset.name}")
```

It fails on:
https://github.com/PyGithub/PyGithub/blob/b4092b5d5255610fa998f66c720ab5700fddabeb/github/Requester.py#L1179

```text
  File "/home/alex/.pyenv/versions/stellarcarbon-3.12/lib/python3.12/site-packages/github/GitReleaseAsset.py", line 163, in download_asset
    self._requester.getFile(self.url, path=path, chunk_size=chunk_size)
  File "/home/alex/.pyenv/versions/stellarcarbon-3.12/lib/python3.12/site-packages/github/Requester.py", line 901, in getFile
    _, _, stream_chunk_iterator = self.getStream(url, parameters, headers, cnx, chunk_size=chunk_size)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/.pyenv/versions/stellarcarbon-3.12/lib/python3.12/site-packages/github/Requester.py", line 928, in getStream
    status, responseHeaders, output = self.__requestEncode(
                                      ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/.pyenv/versions/stellarcarbon-3.12/lib/python3.12/site-packages/github/Requester.py", line 1098, in __requestEncode
    status, responseHeaders, output = self.__requestRaw(
                                      ^^^^^^^^^^^^^^^^^^
  File "/home/alex/.pyenv/versions/stellarcarbon-3.12/lib/python3.12/site-packages/github/Requester.py", line 1186, in __requestRaw
    del requestHeaders["Authorization"]
        ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'Authorization'
```

Apparently, the assumption that an Authorization header will be present is incorrect. I don't know which (if any) other features are impacted by this assumption. Could we simply pop the key, or check for its presence before deletion?

alecglen:
Resolves #3302


github/Requester.py
```diff
@@ -1176,7 +1176,7 @@ def __requestRaw(
                     self._logger.debug(f"Following Github server redirection (302) from {url} to {o.path}")
                 # remove auth to not leak authentication to redirection location
                 if o.hostname != self.__hostname:
-                    del requestHeaders["Authorization"]
+                    requestHeaders = {k: v for k, v in requestHeaders.items() if k != "Authorization"}
```
EnricoMi, line 1179: Good point! Does the `del` pattern appear any where else in this code base?

alecglen: @EnricoMi The only other `del` statements in this project are in the [__getstate__ method](https://github.com/PyGithub/PyGithub/blob/2d4785dd08374147a8eab866eeb704508558336a/github/Requester.py#L459) earlier in the same class. That usage is safe as it 1) operates on a local copy of the dict, and 2) deletes keys that the class guarantees will exist.

PR #3377, opened 2025-09-16
Consider per-page settings when iterating paginated properties

EnricoMi:
The GitHub API has the "feature" of paginated properties. Some objects returned by the API have a property that allows for pagination. Fetching subsequent pages of that property means fetching the entire object (with all other properties) and the specified page of the paginated property. Iterating over the paginated property means fetching all other properties multiple times. Fortunately, the allowed size of each page (`per_page` is usually 300, in contrast to the "usual" `per_page` maximum of 100).

Objects with paginated properties:
- `Commit.files`
- `Comparison.commits`
- `EnterpriseConsumedLicenses.users`

This PR makes iterating those paginated properties use the configured `per_page` setting.

It further allows to specify an individual `per_page` when either retrieving such objects, or fetching paginated properties.

Methods returning objects with paginated properties:
- `Repository.compare`
- `Repository.get_commit`
- `Enterprise.get_consumed_licenses` 

Accessing paginated properties with individual `per_page` value:
- `Commit.get_files`
- `Comparison.get_commits`
- `EnterpriseConsumedLicenses.get_users`

Note: setting the `Github(per_page=…)` to a value other than the default of 30 will cause paginated properties to use that setting, unless a different value is specified when retrieving such objects or their paginated properties via methods listed above. For instance, `Commit.files` fetches paged with that number of files. Before, 300 files were fetched with each request.


github/GithubObject.py
```diff
@@ -639,3 +639,61 @@ def update(self, additional_headers: dict[str, Any] | None = None) -> bool:
...
+        url: str | None = None,
+        accept: str | None = None,
+        per_page: int | None = None,
+    ):
+        if per_page is None and requester.per_page != Consts.DEFAULT_PER_PAGE:
+            per_page = requester.per_page
+
+        # add per_page to URL if instance is incomplete
+        # we only modify the URL if this instance is incomplete
+        if completed is None or completed is False:
```
iarspider:
```suggestion
        if not completed:
```

iarspider: looks cleaner imo

github/Enterprise.py
```diff
@@ -139,16 +139,16 @@ def website_url(self) -> str:
...
-        headers, data = self._requester.requestJsonAndCheck("GET", self.url + "/consumed-licenses")
-        if "url" not in data:
-            data["url"] = self.url + "/consumed-licenses"
-
-        return github.EnterpriseConsumedLicenses.EnterpriseConsumedLicenses(
-            self._requester, headers, data, completed=True
+        assert (
+            licence_users_per_page is None or isinstance(licence_users_per_page, int) and licence_users_per_page > 0
+        ), licence_users_per_page
+        return EnterpriseConsumedLicenses(
```
iarspider:
```suggestion
        return github.EnterpriseConsumedLicenses.EnterpriseConsumedLicenses(
```

github/GithubObject.py
```diff
@@ -639,3 +639,61 @@ def update(self, additional_headers: dict[str, Any] | None = None) -> bool:
...
+            if per_page is None:
+                # we use the default per_page (not Consts.DEFAULT_PER_PAGE as this URL might have a different default)
+                # we set page=1 to get pagination links, PaginatedList can work from there
+                url = self.set_if_not_set(attributes, url, page=1)
+            else:
+                # we set the given per_page
+                url = self.set_if_not_set(attributes, url, per_page=per_page)
```
iarspider:
```suggestion
                url = self.set_if_not_set(attributes, url, per_page=per_page, page=1)
```

otherwise github will return pagination links without `per_page`

iarspider: Actually, github will always (?) return object without `per_page`:

```bash
$ curl -s "https://api.github.com/repos/PyGithub/PyGithub/compare/main...remove-deprecations?per_page=3&page=1" | jq .url
"https://api.github.com/repos/PyGithub/PyGithub/compare/main...remove-deprecations"
```

tests/GithubObject.py
```diff
@@ -242,6 +242,209 @@ def testMakeTimetsampAttributeBadValues(self):
...
+
+    def testRepoComparisonCommitsFilesReversed(self):
+        # tests paginated property of Comparison.commits and Commit.files
+        self.g.per_page = 2
+        repo = self.g.get_repo("PyGithub/PyGithub", lazy=True)
+        comparison = repo.compare("main", "remove-deprecations")
+        # commits is a PaginatedList, which should respect per_page
+        commits = list(reversed(comparison.commits))
+        commit = commits[-1]
+        # files is a PaginatedList, which should respect per_page
```
iarspider: This is not true. Even for explicitly paginated requests (`page=1&per_page=N`), the `url` attribute is returned without pagination information, thus overwriting url used in construction of object(s).

iarspider:
Before https://github.com/PyGithub/PyGithub/blob/9ca6b4b31274b62ff017e80a728da5c51fdb2e1a/github/PaginatedList.py#L217, I suggest to add

```
            # firstUrl, taken from url property of returned object, doesn't contain per_page
            if self._reversed:
                if self.__lastUrl:
                    self.url = self.__lastUrl
            else:
                if self.__firstUrl:
                    self.url = self.__firstUrl
```

tests/GithubObject.py
iarspider: So maybe we need to change all `_useAttributes` methods to not overwrite `_url` if it's set?

iarspider:
Unit Tests are green, yay!

iarspider:
@EnricoMi a kind reminder of this PR. We would really like to get this merged and not maintain patches.

EnricoMi:
@iarspider I have to go over this once more as this is a significant change. But it should be done properly. I hope the notion of paginated properties and when to provide a `per_page` does affect what is properly documented and understandable.

EnricoMi:
@iarspider this is finally complete, please have a look and test with your use-case.

https://pygithub--3377.org.readthedocs.build/en/3377/utilities.html#classes-with-paginated-properties

iarspider:
> @iarspider this is finally complete, please have a look and test with your use-case.
> 
> https://pygithub--3377.org.readthedocs.build/en/3377/utilities.html#classes-with-paginated-properties

Thanks! I'll try to check this ASAP.

EnricoMi:
@iarspider have you had a chance to take a look / test on your side?

EnricoMi:
@iarspider thanks for raising this issue!

iarspider:
@EnricoMi thanks! Sorry, I couldn't test it - I am no longer working on that project.

PR #3380, opened 2025-09-18
Deprecate paginated properties in Commit and Comparison

EnricoMi:
Deprecates paginated properties `Commit.files` and `Comparison.commits`. Use respective getter methods instead.


EnricoMi:
There is a use-case for having these paginated properties (reuse first page retrieved together with the object). This reduces number of requests.

Superseded by #3377.

PR #3381, opened 2025-09-19
Add support for `type` parameter to get_issues

nrysk:
## 1. Describe the problem the Pull Request attempts to solve

This pull request adds support for the [`type` query parameter](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues). a feature recently added to the GitHub Issues. Currenty PyGithub's `get_issues` method does not have a way to filter issues by their type.

## 2. Explain how the you went about solving the problem

I have updated the `get_issues` method in `github/Repository.py` to accept a new optional `type` parameter. When this parameter is provided, it is included in the URL parameter for API request.

## 3. Provide a test that exemplifies the expected behaviour

I have added two test methods, `testGetIssuesWithTypeArgument` and `testGetIssuesWithTypeWildcard`.

---
*As my native language is not English, I apologize for any errors in my writing.*


github/Repository.py
```diff
@@ -3166,6 +3166,7 @@ def get_issues(
         milestone: Milestone | Opt[str] = NotSet,
         state: Opt[str] = NotSet,
         assignee: NamedUser | Opt[str] = NotSet,
+        type: Opt[str] = NotSet,
```
EnricoMi: Given this is a positional argument, we have to add this at the end of the argument list, unfortunately. Otherwise this might be breaking for some users.

nrysk:
Thanks! I've applied the change.

EnricoMi (approved):
LGTM!

PR #3396, opened 2025-10-29
Add `incomplete_results` property to `PaginatedList`

### LINKED ISSUE #2166: Non deterministic behavior when iterating over paginatedList after calling search_code
Hello community,
I am using the `search_code` function to find repositories which used a specific term in a file. When iterating over the result within a for loop and just printing the current counter to see how far the loop goes, the program does **not** reach `result.totalCount` as expected. It stops more or less randomly somewhere before. 

```
query = "\"known_third_party = \"kedro\"\" in:file language:toml"
counter = 0
result = github.search_code(query)
print(str(result.totalCount))
for code in result:
    counter += 1
    print(counter, code.repository.clone_url)
    time.sleep(1)
```

Executing the above code three times in a row, the totalCount was always 340, but the program stopped at 280, 311 and 318.

Can you please help me with that? Why does the loop not reach the last item?

Thanks a lot for your help!

With best regards,
Alex

### LINKED ISSUE #2253: `incomplete_results` is missing as part of the search-results
I noticed this when using PyGitHub. I saw this comment as well in another issue:

---
Reading the Github API docs about search, I also notice that `incomplete_results` is missing as part of the search-results processin in PyGithub. Probably including that value might also already help out with detecting if search results might be (in)complete.

_Originally posted by @mfonville in https://github.com/PyGithub/PyGithub/issues/824#issuecomment-399421378_
---

I would love to have this property.

### LINKED ISSUE #2812: Inconsistent results when running search_repositories and sorting by descending stars
I was running the search_repositories function a few times (just 10 or so minutes apart) like this:

```python
g = Github(auth=auth)
query = "language:python"
result = g.search_repositories(query, sort="stars", order="desc")
repo_list = list(result)
```

So I'm searching Python repositories in descending order of stars... but I noticed I got inconsistent results on consecutive runs. For example sometimes the results returned Flask and FastAPI in the top ten results (which is expected) but other times those repositories were completely absent from the results.

Searching on descending order of stars should give very consistent results for the top repos (that have wide margins between their star counts) but its dropping repos that I know should be in the results.

Felixoid:
The patch to fix #2166, fix #2253, and fix #2812

According to https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#timeouts-and-incomplete-results, the incomplete results are expected and should be identified by the `incomplete_results` flag.

This patch adds the flag and tests for it.


Felixoid:
Dear @EnricoMi, I am not sure regarding the `totalCount` implementation here.

The `PaginatedList` class requests URLs here and there, and it feels like either the `incomplete_results` should be set everywhere or there should be a helper method.

Felixoid:
Hey-hey dear @EnricoMi, is there a way I can help to get this change into a released version?

We're very badly affected by the lack of `incomplete_results` flag in PaginatedList objects. There's no way to recognize if there aren't any issues/PRs to process, or an internal search timeout reached currently.

We wouldn't push so hard, but I can't find a local workaround for this issues as well.

Felixoid:
Hi-hi!

I'd like to get the first review round, please :pray:

github/PaginatedList.py
```diff
@@ -222,6 +223,14 @@ def __init__(
     def is_rest(self) -> bool:
         return self.__is_rest
 
+    @property
+    def incomplete_results(self) -> bool:
```
EnricoMi: There should be a way to distinguish `True` and `False` from "not known":

```suggestion
    def incomplete_results(self) -> bool | None:
```

EnricoMi (commented):
Thanks for raising this.

I think `incomplete_results` is unknown until we start iterating or touch `totalCount` (i.e. fetch the first page), right?

github/PaginatedList.py
```diff
@@ -203,6 +203,7 @@ def __init__(
         if self.__requester.per_page != 30:
             self.__nextParams["per_page"] = self.__requester.per_page
         self._reversed = False
+        self._incomplete_results = False
```
Felixoid:
```suggestion
        self._incomplete_results = None
```

Felixoid:
> I think `incomplete_results` is unknown until we start iterating or touch `totalCount` (i.e. fetch the first page), right?

Right, and I couldn't find a way where the correct place to catch `incomplete_results` while getting the `totalCount` during the first request https://github.com/PyGithub/PyGithub/pull/3396/files#diff-b6cef6ca688edac9c6e3644e79651d704760a8de87dbe41bb0c756cef6e6c3f7R21

Maybe you can suggest it?

Felixoid:
> Maybe you can suggest it?

I think I nailed it

Felixoid:
Rebased it onto the current master. Hopefully, it will address failing schemas workflow jobs.

EnricoMi:
I have aligned `incomplete_results` with behaviour of `totalCount` so it triggers a request when called first. Method `PaginatedList.get_page` now also updates `incomplete_results`. I have also reworked tests.

And I have added documentation: https://pygithub--3396.org.readthedocs.build/en/3396/utilities.html#pagination

Felixoid:
Looks pretty solid! Not sure why the [codecov/patch](https://github.com/PyGithub/PyGithub/pull/3396/checks?check_run_id=59677894776) complains, including the places that aren't touched in the PR.

Looking forward to having it merged :pray:

EnricoMi (approved):
LGTM!

Felixoid:
Is there something else before the merge?

Felixoid:
Thanks for merging!

Do you plan to spin out a new release?

dorz-dev:
when will this fix be released?

Issue #3398, opened 2025-10-30
o.hostname and self.__hostname don't align for private repos


ViktorWelbers:
Hi everyone

There is an issue with private domains in the requester:

The following setup, will fail because the way the domains for private repos are matched is currently flawed, when setting a base_url like **https://api.github.YOUR-ORG.com** in the GitHub class (Library Version 2.8.1)

```python
    ...
    # Create GitHub App integration
    auth = Auth.AppAuth(app_id, private_key).get_installation_auth(
        installation_id
    )

    # Create GitHub client with access token
    github = Github(base_url="https://api.github.ORG.com", auth=auth)

    repo = github.get_repo(f"{repo_owner}/{repo_name}")  # <---- Will work without any problems
    readme = repo.get_readme(ref=branch)  # <---- Will explode because of assertion error
    ...
```
Will have this error:

```bash
Traceback (most recent call last):
  File "/home/code/main.py", line 70, in <module>
    main()
  File "/home/code/main.py", line 66, in main
    raise e
  File "/home/code/main.py", line 50, in main
    readme_content = fetch_github_readme(
                     ^^^^^^^^^^^^^^^^^^^^
  File "/home/code/main.py", line 37, in fetch_github_readme
    return repo.get_readme(ref=branch).decoded_content.decode("utf-8")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/code/.venv/lib/python3.12/site-packages/github/Repository.py", line 3513, in get_readme
    headers, data = self._requester.requestJsonAndCheck("GET", f"{self.url}/readme", parameters=url_parameters)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/code/.venv/lib/python3.12/site-packages/github/Requester.py", line 630, in requestJsonAndCheck
    self.__customConnection(url),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/code/.venv/lib/python3.12/site-packages/github/Requester.py", line 882, in __customConnection
    self.__assertUrlAllowed(url)
  File "/home/code/.venv/lib/python3.12/site-packages/github/Requester.py", line 876, in __assertUrlAllowed
    assert self.__hostnameHasDomain(o.hostname, self.__hostname), o.hostname
[...]

PR #3399, opened 2025-10-30
fix: Improve domain matching in Requester.py

ViktorWelbers:
Enhance domain matching logic to support reverse checks.

Related to issues described in here: https://github.com/PyGithub/PyGithub/issues/3398

This PR enables the use of base_urls like "api.github.ORG.com" which would previously have failed to work.

Look into [issue ](https://github.com/PyGithub/PyGithub/issues/3398)for a more detailed report.


EnricoMi:
Superseded by #3419.

PR #3426, opened 2025-12-19
Fix Workflow.create_dispatch

haxer-max:
Currently` Workflow.create_dispatch` is broken. It returns false even after successful workflow dispatch. This is because status code of underlying API is now 200 instead of 204. This is fixed here

issue for the same: https://github.com/PyGithub/PyGithub/issues/3427


EnricoMi:
See #3427 for why this is not needed atm.

PR #3430, opened 2026-01-08
Clarify that organization members do not include role or permissions

### LINKED ISSUE #1477: unable to read or set the owners for an organization
I am trying to make a script that is both capable of determining the current owners of an org and setting them. I started by trying to just read who the current owners are but am having issues with that.

I've tried to do

```
git = Github(access_token)
self.org = self.git.get_organization('ORG')
for user in self.org.get_members(role="admin"):
   print(f"user = {user}, user.permissions = {user.permissions}, user.role = {user.role}, user.site_admin = {user.site_admin}, user.type = {user.type}")
```

But when it comes across a member that is an admin, it prints out 
```
2020-04-20 21:14:43 = INFO - user = NamedUser(login="userX"), user.permissions = None, user.role = None, user.site_admin = False, user.type = User
```

What am I doing wrong?

Avani1626:
Fixes #1477

The GitHub API does not include role or permission details for
organization members returned by Organization.get_members().

PyGithub currently exposes user.role and user.permissions for these
users, which can be misleading as they are always None.

This change adds documentation clarifying this API limitation so users
understand the expected behavior.

### DIFF
--- a/github/Organization.py
+++ b/github/Organization.py
@@ -1251,6 +1251,12 @@ class Organization(CompletableGithubObject):
     ) -> PaginatedList[NamedUser]:
         """
         :calls: `GET /orgs/{org}/members <https://docs.github.com/en/rest/reference/orgs#members>`_
+        Note:
+    The GitHub API does not include role or permission details
+    for organization members returned by this method.
+    As a result, attributes such as `user.role` and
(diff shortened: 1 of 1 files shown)


EnricoMi:
Thanks! Can you add a link to the GitHub API docs, please?

Issue #3434, opened 2026-01-14
Reaction.delete no longer works


iarspider:
Title says it all: calling `Reaction.delete` just returns an error:

```
Requester.py:586 [PyGithub] DELETE https://api.github.com/reactions/323843145
process_pr_v2.py:4441 [WARNING] Failed to remove old reaction: 404 {"message": "Not Found", "documentation_url": "https://docs.github.com/rest", "status": "404"}
```

The [URL in docstring](https://docs.github.com/en/rest/reference/reactions#delete-a-reaction-legacy) is gone, the right way to delete a reaction now is to call `IssueComment.delete_reaction(...)` or similar method. 

I can open a PR removing this method if that's the right approach.

EnricoMi:
Let's deprecate the method first. There might be old versions of Github Enterprise out there where this still exists.

Please create a pull request to deprecate it, and another one to remove this. Thanks!

PR #3435, opened 2026-01-14
Deprecate `Reaction.delete`

iarspider:
As discussed in https://github.com/PyGithub/PyGithub/issues/3434 . The endpoint no longer exists in public GitHub.


EnricoMi (approved):
LGTM!

PR #3436, opened 2026-01-14
Drop Reaction.delete method

iarspider:
As discussed in https://github.com/PyGithub/PyGithub/issues/3434 . The endpoint no longer exists in public GitHub.


EnricoMi:
I have turned this into a draft, to postpone the merge for a while (at least after the 2.9.0 or 3.0.0 release).

PR #3437, opened 2026-01-16
Add Python 3.14 to pyproject

Repiteo:
- Related: #3429

Officially bumps support of the tool itself to 3.14. ~~Includes an explicit bump for pre-commit's "docformatter," as one of their 1.7.7 dependencies not supporting 3.14 and was fixed upstream, but hasn't gotten an official release in that time.~~ **EDIT:** pre-commit changes proved to be more involved, will handle elsewhere

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ classifiers = [
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
   "Topic :: Software Development",
 ]
 description = "Use the full Github API v3"


EnricoMi (approved):
LGTM!

EnricoMi:
Thanks for this!

Repiteo:
Shoot, it looks like docformatter updated beyond just the dependency fix, so a bunch of files say they need changes. Gonna look into this now

Repiteo:
Yeah, there's unfortunately not an "easy" fix here. The dependency fix came *after* a few other changes, so I can't isolate it upstream. A shame too, because this update brought a *lot* of false-positives with it.

I could change this PR's scope to just updating `pyproject.toml`, as that dependency is isolated to the pre-commit hook, so it's not a blocker for 3.14 support

PR #3438, opened 2026-01-17
Replace `black`, `pyupgrade`, and `docformatter` with `ruff`

Repiteo:
- Related: #3437

Originally the above PR updated `docformatter` to support Python 3.14, but that ended up having a lot of unintended side effects, so I started looking for alternatives. I realized that this tool was somewhat redundant, as we already had `ruff`, which has the capability to handle docstring logic on its own. Turns out, this went beyond that, as we don't really have a need for `black` nor `pyupgrade` either; the former was what `ruff` was initially designed as a drop-in replacement for, and the latter has a dedicated ruleset already built-in

This PR removes the pre-commit hooks and settings for `black`, `pyupgrade`, and `docformatter` entirely. They've been replaced with `ruff` linting and formatting equivalents. Docstrings could undergo further changes, but I held off in this PR as that level of noise warrants isolation. This has the bonus of ensuring `pre-commit` will function as expected when using 3.14 and later

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -12,3 +12,6 @@ e113e37de1ec687c68337d777f3629251b35ab28
 
 # Run pre-commit autoupdate and run all hooks
 4c24c73d34b54a061b0c800eeea64dc64c230a63
+
+# Apply ruff to whole codebase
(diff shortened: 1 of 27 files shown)


EnricoMi (approved):
LGTM!

PR #3446, opened 2026-02-09
Add support for GitHub API versioning (X-GitHub-Api-Version header)

### LINKED ISSUE #3442: GitHub now has versioned APIs
To avoid breaking changes I think this client library should support passing in an API version?

https://docs.github.com/en/rest/about-the-rest-api/api-versions?apiVersion=2022-11-28

Perhaps have this library pinned to a specific version of the GitHub API?

daniel7an:
## Summary

Adds support for GitHub's versioned REST API by sending the `X-GitHub-Api-Version` header with all requests.

## Changes

- Added `DEFAULT_API_VERSION = "2022-11-28"` constant in `Consts.py`
- Added `api_version` parameter to `Github`, `GithubIntegration`, and `Requester` constructors
- The `X-GitHub-Api-Version` header is included in all API requests when `api_version` is set
- Defaults to `"2022-11-28"` (current stable version per [GitHub docs](https://docs.github.com/en/rest/about-the-rest-api/api-versions))
- Can be set to `None` to omit the header (preserving backward compatibility)

## Usage

```python
# Default: uses '2022-11-28'
g = github.Github(auth=auth)

# Specific version
g = github.Github(auth=auth, api_version="2022-11-28")

# Opt out of sending the header
g = github.Github(auth=auth, api_version=None)
```

Fixes #3442

### DIFF
--- a/github/GithubIntegration.py
+++ b/github/GithubIntegration.py
@@ -85,6 +85,7 @@ class GithubIntegration:
         jwt_expiry: int = Consts.DEFAULT_JWT_EXPIRY,
         jwt_issued_at: int = Consts.DEFAULT_JWT_ISSUED_AT,
         jwt_algorithm: str = Consts.DEFAULT_JWT_ALGORITHM,
+        api_version: str | None = Consts.DEFAULT_API_VERSION,
         auth: AppAuth | None = None,
         # v3: set lazy = True as the default
         lazy: bool = False,
@@ -101,6 +102,8 @@ class GithubIntegration:
         :param pool_size: int
         :param seconds_between_requests: float
         :param seconds_between_writes: float
+        :param api_version: string, GitHub API version to use (sent as X-GitHub-Api-Version header),
+                            defaults to "2022-11-28", set to None to not send the header
         :param jwt_expiry: int deprecated, use auth=github.Auth.AppAuth(...) instead
         :param jwt_issued_at: int deprecated, use auth=github.Auth.AppAuth(...) instead
         :param jwt_algorithm: string deprecated, use auth=github.Auth.AppAuth(...) instead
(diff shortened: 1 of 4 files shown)


EnricoMi:
Thanks for getting this started.

I think as a first step, the default should be `None`, so nothing changes for users. Then, this PR at least gives users a chance to fix the Github API version if they want or need to in the future.

With a default version, we would have to touch all of our `ReplayData` as well.

github/GithubIntegration.py
```diff
@@ -85,6 +85,7 @@ def __init__(
         jwt_expiry: int = Consts.DEFAULT_JWT_EXPIRY,
         jwt_issued_at: int = Consts.DEFAULT_JWT_ISSUED_AT,
         jwt_algorithm: str = Consts.DEFAULT_JWT_ALGORITHM,
+        api_version: str | None = Consts.DEFAULT_API_VERSION,
```
EnricoMi, line 88: I think it is fair to move this up right after `base_url`, but as the first non-positional argument.

github/GithubIntegration.py
```diff
@@ -101,6 +102,8 @@ def __init__(
         :param pool_size: int
         :param seconds_between_requests: float
         :param seconds_between_writes: float
+        :param api_version: string, GitHub API version to use (sent as X-GitHub-Api-Version header),
+                            defaults to "2022-11-28", set to None to not send the header
```
EnricoMi, line 106:
```suggestion
        :param api_version: string, GitHub API version to use (see https://docs.github.com/en/rest/about-the-rest-api/api-versions),
                            uses GitHub's default version if None, defaults to None
```

github/GithubIntegration.py
EnricoMi, line 119: Please assert type of `api_version` here as well.

PR #3450, opened 2026-02-14
fix totalCount KeyError when last link has no page param

### LINKED ISSUE #1006: PaginatedList.totalCount KeyError: 'page' with large results
When using get_repo().totalCount on a small result set (only my own repositories) totalCount returns the repository count correctly. When querying my enterprise Github for a complete list of repositories, the following error occurs:

```bash
  File "./test-repo-list.py", line 9, in <module>
    print("Found %s repositories." % full_repo_list.totalCount)
  File "/usr/local/lib/python3.7/dist-packages/github/PaginatedList.py", line 175, in totalCount
    self.__totalCount = int(parse_qs(lastUrl)['page'][0])
KeyError: 'page'
````

I am using Python 3.7.1 and the pip-installed version 1.43.4 of PyGithub with a Github Enterprise version 2.14.7.

This is the code I have used to test this:

```python
#!/usr/bin/env python3.7
from github import Github
import os

api_token = os.getenv('GITHUB_TOKEN', None)
g = Github(base_url="https://mygithubhost.lads", login_or_token=api_token)

full_repo_list = g.get_repos()
print("Found %s repositories." % full_repo_list.totalCount)
```

I did notice, after some skimming of PaginatedList.py, that the Link headers differ between the two tests. When collecting just my own repositories, I see the "last" link returned in the link header, but when querying for all repositories I only see "next" and "first" links returned.

g.get_user().get_repos():
```json
'link':'<https://mygithubhost.lads/api/v3/user/repos?per_page=1&page=2>; rel="next", <https://mygithubhost.lads/api/v3/user/repos?per_page=1& 
page=10>; rel="last"'`
```

g.get_repos():
```json
'link':'<https://mygithubhost.lads/api/v3/repositories?per_page=1&since=112>; rel="next", <https://mygithubhost.lads/api/v3/repositories{?since}>; rel="first"'
```

I don't know what's causing this difference in behavior, though.

If I don't try to get totalCount from the object, it works as expected, but totalCount specifically seems to fail.

Mr-Neutr0n:
`PaginatedList.totalCount` crashes with `KeyError: 'page'` when the GitHub API returns a `rel="last"` link that uses `since`-based pagination instead of `page`-based pagination. This happens with endpoints like list-all-public-repositories on GitHub Enterprise.

The root cause is two things:
1. `parse_qs()` was being called on the full URL instead of just the query string — it worked by accident for `page` since it appeared after `&`, but `per_page` (the first param) was being parsed with the full URL path as part of its key.
2. No check for whether `page` actually exists in the parsed params before accessing it.

Fixed by using `urlparse().query` to extract the query string properly, and checking for `page` before accessing it. When `page` isn't present, the code falls through to the existing data-based counting logic.

Fixes #1006

### DIFF
--- a/tests/PaginatedList.py
+++ b/tests/PaginatedList.py
@@ -323,6 +323,13 @@ class PaginatedList(Framework.TestCase):
         self.assertEqual(review_requests[0].totalCount, 0)
         self.assertEqual(review_requests[1].totalCount, 0)
 
+    def testTotalCountWithNoPageParam(self):
+        # When the "last" link uses since-based pagination instead of page-based,
+        # totalCount should fall back to counting via data instead of raising KeyError.
+        # See https://github.com/PyGithub/PyGithub/issues/1006
+        repos = self.g.get_repos()
+        self.assertEqual(repos.totalCount, 1)
+
     def testTotalCountWithDeprecationLink(self):
         # Test the original reported scenario: search_issues with commit SHA
         issues = self.g.search_issues("commit:example_sha")
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -56,7 +56,7 @@ from __future__ import annotations
 
 from collections.abc import Callable, Iterator
 from typing import Any, Generic, TypeVar, overload
-from urllib.parse import parse_qs
+from urllib.parse import parse_qs, urlparse
 
 from github import Consts
 from github.GithubObject import GithubObject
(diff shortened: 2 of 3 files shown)


tests/PaginatedList.py
```diff
@@ -323,6 +323,13 @@ def testTotalCountWithDictionary(self):
         self.assertEqual(review_requests[0].totalCount, 0)
         self.assertEqual(review_requests[1].totalCount, 0)
 
+    def testTotalCountWithNoPageParam(self):
+        # When the "last" link uses since-based pagination instead of page-based,
+        # totalCount should fall back to counting via data instead of raising KeyError.
+        # See https://github.com/PyGithub/PyGithub/issues/1006
+        repos = self.g.get_repos()
+        self.assertEqual(repos.totalCount, 1)
```
EnricoMi, line 331: Thanks for looking into this issue!

The last URL says per_page=1&since=99999, doesn't this indicate that totalCount should be 100000?

EnricoMi: Looks like `since` refers to the id of the last returned instance (here repository id), so there is no way to know the total count of this paginated list.

The `totalCount` refers to the number of elements the entire paginated list contains, not the size of one page. So can only tell after all pages have been retrived.

Mr-Neutr0n:
@EnricoMi — good catch, and your follow-up analysis is exactly right: `since=99999` is the id of the last repository, not a count, so we genuinely can't infer the total from it. The fix deliberately falls back to "count what's in `self._data`" when the `last` link has no `page` query parameter, which is what the test asserts.

So the test's `assertEqual(repos.totalCount, 1)` is the *current page* count, not the *full* list count — it's just verifying the fallback path no longer raises and produces the same number as `len(repos)`. I'll add a one-line comment to the test to make that intent explicit, and rebased the branch (this PR is currently `mergeable: false`, looks like it's drifted from `main`).

EnricoMi:
I think the semantis of `totalCount` is the size of the paginated list, so the number of items you can expect when fully iterating the list. Then the assertion that you are referring to is wrong.

I think this new situation introduced by `since` needs more rework / rethinking: #3451, #3452

Mr-Neutr0n:
That makes sense — agreed that my fallback-to-`len(data)` patch is not preserving the existing `totalCount` semantics, it is only avoiding the crash. With `since`-based pagination, the right fix is broader than this narrow PR and likely needs the API/design work you sketched in #3451 and #3452 (optional/unknown `totalCount`, or a separate `retrievableCount`/length-style concept).

I am going to close this PR rather than push a semantically-wrong partial fix onto the branch. If it helps later, I am happy to come back with a new PR against the direction you settle on in those follow-up issues.

PR #3464, opened 2026-03-19
Async support

### LINKED ISSUE #1538: Async support
What do you think about PyGithub async support? If I do async support for PyGithub - do you accept it?

Ousret:
This PR brings native asyncio support to PyGithub by generating the code from the sync code.

In order to achieve such thing we switched Requests for Niquests, which is a drop-in replacement for Requests.
It keeps everything, including the "responses" mocking utility.

The main changes here are:

- Switching Requests to Niquests
- Creating the "generate_async.py" script included in pre-commit
- conftest.py update to bring async support with responses

The rest is purely generated code. `github.asyncio` and `tests.asyncio`.

The `tests.asyncio` is generated by directly copying the sync part +/- replacing some key part:

- A concrete example is the lazy `my_gh_object[0]` that depends on `__getitem__` which cannot be made awaitable. So we introduced a bare `getitem` coroutine method instead.

And in order to avoid unnecessary complexity, we've built a SyncProxy to avoid doing heavy inference/transformation in tests. We assessed the usage of doing everything under AsyncioIsolatedTestCase but it was too difficult to bring to life.

The async package is fully derived from the sync source. The generator handles import rewriting, `async def`/`await` insertion, property promotion, and type annotation fixups. No manual async code to maintain. And as long as the project APIs/interfaces are stable, the script will continue to work as-is.

I am the maintainer of Niquests. I'm transparent about this because it cuts both ways: it means I have direct commit access to fix any compatibility issue that arises, and it means I'm personally committed to long-term support for this integration.

There's a ton of opportunity with this, like running this natively in Pyodide, true HTTP/2 support, OS trust store by default, and much more.

I can maintain the `generate_async.py` script as long as I can. 

Close https://github.com/PyGithub/PyGithub/issues/1538

Regards,


Ousret:
Everything should be in order. I fixed the last remaining issues (typing mostly) and brought coverage to 96.3% overall.
Also improved the performance of the async generator.

I would appreciate any early comments / tests about this.

Regards,

doc/async.rst
```diff
@@ -0,0 +1,190 @@
...
+What stays the same
+-------------------
+
+- **Authentication** -- ``Auth.Token``, ``Auth.AppAuth``, ``Auth.Login`` etc. are
+  the same sync objects used identically in both modes.
+- **Data classes** -- ``InputGitTreeElement``, ``InputGitAuthor``, and similar input
+  objects are plain data holders and are not async.
+- **Exceptions** -- ``GithubException``, ``UnknownObjectException``,
+  ``BadCredentialsException``, etc. are raised the same way.
+- **``__repr__``/``__eq__``/``__hash__``** -- These dunder methods stay synchronous;
```
EnricoMi, line 151: This doesn't render nicely, maybe we need some whitespaces here:

```suggestion
- **``__repr__`` / ``__eq__`` / ``__hash__``** -- These dunder methods stay synchronous;
```

EnricoMi (commented):
Rather than having two variants (sync, async) for each example code could we have the `async.rst` show how to turn sync code into async code and provide some async examples there and have all other (existing) examples be sync?

I think users that want (and know async) code can easily turn sync example code into async code.

EnricoMi:
Given `No manual async code to maintain`, there is no need to check in the async code, right? It could be generated during release process. Is there any advantage during code review to have the async code in git?

EnricoMi:
Would it make sense to release separate sync and async PyGithub package? The former depending on requests, the latter on niquest?

EnricoMi:
Is the `scripts/generate_async.py` script (except for some PyGithub config) specific to this project or generic to any project using niquest? Could this be installed from a PyPi package during async release generation?

Ousret:
> Rather than having two variants (sync, async) for each example code could we have the async.rst show how to turn sync code into async code and provide some async examples there and have all other (existing) examples be sync?

That's fair. I'll remove the dual tabs from the existing docs and consolidate async guidance into a dedicated `async.rst` with a clear "sync-to-async" migration section and a few representative examples. We can always revisit inline tabs later if user demand warrants it. I did it initially because I wrote Niquests docs with that strategy.

> Given `No manual async code to maintain`, there is no need to check in the async code, right? It could be generated during release process. Is there any advantage during code review to have the async code in git?

There is a trade-off, but I think checking it in is the better default for a few reasons:

- Maintainers can inspect what the generator actually produces on every PR without running it locally. This is especially valuable during the initial stabilization period.
- Contributors and users can browse the async API directly in the repository. Anyone cloning the project can run the async tests immediately without an extra generation step.
- The async test suite runs against the checked-in code on every PR. If generation were deferred to release time, issues could slip through undetected during the development cycle.
I'm aware the main downside is larger diffs. To mitigate that, I can add a `linguist-generated=true` attribute in `.gitattributes` for `github/asyncio/` and `tests/asyncio/` so GitHub collapses those files by default in PR reviews. That way the generated code is in git but doesn't clutter the review UI.

> Would it make sense to release separate sync and async PyGithub package? The former depending on requests, the latter on niquest?

I should be transparent: as the Niquests maintainer I have an obvious interest here, so I'll stick to the technical merits and you can weigh them accordingly.

First, to address the underlying concern: Niquests is a drop-in replacement for Requests at the API level. Existing sync code, including the `responses` mock library, works without modification. The migration in this PR touches imports, not call sites. So the impact on sync-only users is a dependency swap, not a behavioral change.
That said, maintaining two separate packages would introduce concrete problems:

-Both packages would need the same models, pagination logic, and error handling. Keeping them in sync across two release cycles is a real maintenance burden. 
- Two packages means two sets of type annotations that need to stay consistent. This tends to drift over time.
- Anyone who needs both sync and async in the same project (not uncommon) would have to manage two dependencies with potentially conflicting versions.
- Two full test matrices, two release pipelines, two sets of changelogs.

A single package with the async path as a subpackage (`github.asyncio`) keeps the surface area manageable while giving async users first-class support. A split would also fragment the issue tracker and documentation, which hurts users searching for help.

> Is the `scripts/generate_async.py` script (except for some PyGithub config) specific to this project or generic to any project using niquest? Could this be installed from a PyPi package during async release generation?

The script is specific to PyGithub. I explored building a generic sync-to-async transformer but the heuristics end up too project-specific (import rewriting rules, property-to-coroutine promotion, test fixture handling, etc.). In its current form, the script is where it belongs. If the approach proves successful here and patterns emerge, extracting a reusable tool could be worth revisiting, but it's premature today.

Hope that clarifies. Let me know which of these you'd like me to action first, I'll start with the docs consolidation.
Regards,

Ousret:
Documentation updated.

regards,

doc/examples/Issue.rst
```diff
@@ -8,7 +8,7 @@ Get issue
```
EnricoMi, line 8: unrelated changes, please revert

doc/examples/MainClass.rst
```diff
@@ -44,8 +44,8 @@ Get enterprise consumed licenses by name
```
EnricoMi, line 44: unrelated changes, please revert

doc/examples/Repository.rst
```diff
@@ -57,7 +57,7 @@ Get list of code scanning alerts
         Use of a broken or weak cryptographic hashing algorithm on sensitive data
```
EnricoMi, line 57: unrelated changes, please revert

EnricoMi (changes requested):

EnricoMi:
Can we make `github.asyncio.CLASS` derive from `github.CLASS` to reduce code duplication (`_initAttributes`, `_useAttributes`, all properties)?

doc/examples/Issue.rst
Ousret: reverted

doc/examples/MainClass.rst
Ousret: reverted

doc/examples/Repository.rst
Ousret: reverted

Ousret:
> Can we make `github.asyncio.CLASSS` derive from `github.CLASS` to reduce code duplication (`_initAttributes`, `_useAttributes`, all properties)?

I explored this early on and hit fundamental blockers. The sync to async transformation is viral; it cascades through the entire class hierarchy:

1. `_completeIfNotSet` is `def` in sync and `async def` in async. Every property on a `CompletableGithubObject` subclass calls it, so every property must also become async. If `AsyncRepository` inherited from `Repository`, it would inherit ~80 sync properties and need to re-declare all of them to add `await` -- no duplication saved.
2. `_useAttributes` looks identical but resolves class references (e.g., `NamedUser`, `Organization`) to different modules via imports. Inheriting it from sync would construct sync sub-objects inside async objects.
3. `CompletableGithubObject.__init__` calls `self.complete()` synchronously in sync mode. In async, we can't `await` in `__init__`, so a flag is stored instead. Inheriting the sync `__init__` would trigger a synchronous completion call in an async context.

This is a well-known limitation in the Python ecosystem, projects like SQLAlchemy, httpx, and others that offer both sync and async APIs use code generation or fully separate codepaths rather than inheritance for the same reasons.

The generator handles all of this automatically. The duplicated code is never manually written or maintained, so in practice the duplication is cosmetic (in the repository) rather than a maintenance burden.

github/asyncio/MainClass.py
```diff
@@ -0,0 +1,1098 @@
...
+        app_auth: AppAuthentication | None = None,
+        base_url: str = Consts.DEFAULT_BASE_URL,
+        timeout: int = Consts.DEFAULT_TIMEOUT,
+        user_agent: str = Consts.DEFAULT_USER_AGENT,
+        per_page: int = Consts.DEFAULT_PER_PAGE,
+        verify: bool | str = True,
+        retry: int | niquests.RetryConfiguration | None = default_retry,
+        pool_size: int | None = None,
+        seconds_between_requests: float | None = Consts.DEFAULT_SECONDS_BETWEEN_REQUESTS,
+        seconds_between_writes: float | None = Consts.DEFAULT_SECONDS_BETWEEN_WRITES,
```
EnricoMi, line 185: We are making great effort to not trigger [Github's primary and secondary rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10). I think this throttling does not make sense in the presence of asyncio. These arguments should be removed and set `None` internally, I think.

EnricoMi: On a similar note: I guess it is not a good idea to target the public api,github.com with asyncio. This only seems useful for an enterprise instance where you pay for the scalability of your Github instance, right?

github/Auth.py
```diff
@@ -327,14 +327,13 @@ def __init__(
...
         token_permissions: dict[str, str] | None = None,
-        requester: Requester | None = None,
+        requester: Any = None,
     ):
         super().__init__()
 
         assert isinstance(app_auth, AppAuth), app_auth
         assert isinstance(installation_id, int), installation_id
         assert token_permissions is None or isinstance(token_permissions, dict), token_permissions
-        assert requester is None or isinstance(requester, Requester), requester
```
EnricoMi, line 337: Can you elaborate why we cant assert the type of the requester any more? Shouldn't this be either `Requester` or `asyncio.Requester`?

github/Requester.py
```diff
@@ -135,17 +134,17 @@ def getheaders(self) -> ItemsView[str, str]:
         return self.headers.items()
 
     def read(self) -> str:
-        return self.response.text
+        return self.response.text or ""
 
-    def iter_content(self, chunk_size: int | None = 1) -> Iterator:
-        return self.response.iter_content(chunk_size=chunk_size)
+    def iter_content(self, chunk_size: int | None = -1) -> Iterator:
```
EnricoMi, line 139: What does chuck size of `-1` mean? Not documented in https://github.com/jawah/niquests/blob/main/src/niquests/models.py#L1172.

github/Requester.py
```diff
@@ -1327,7 +1318,7 @@ def __requestRaw(
                 return self.__requestRaw(
                     cnx, verb, path, requestHeaders, input, stream=stream, follow_302_redirect=True
                 )
-            return status, responseHeaders, output
+            return status or 0, responseHeaders, output
```
EnricoMi, line 1326: Not sure treating `None` status code as `0` is sensible here.

What did `requests` used to do with `None` status code to provide `Respone.status_code` being an `int` (and not `int | None`)?

github/asyncio/MainClass.py
Ousret: > I think this throttling does not make sense in the presence of asyncio.

It is still very relevant to that part. I agree that it is not ideal in it's current form.
But it work "correctly" as far as I can tell.
It still prevent pushing too many requests at once to github servers.

> On a similar note: I guess it is not a good idea to target the public api,github.com

Most people are going to use the regular public exposed api (I do) and we don't need too much request pushed at once, most of the time we need to comply to asyncio I/O scheduler (i.e. not blocking the event loop)

github/Requester.py
Ousret: You missed it by a couple of lines. https://github.com/jawah/niquests/blob/bc042f340490c0992f33eaffa1592163db9b5769/src/niquests/models.py#L1190

"-1" is a new feature introduced by Niquests. It mean truly give me what comes out of the socket x http state machine at once (thus, don't buffer)

github/Requester.py
Ousret: it's a bizarre aspect of Requests and the baggage that comes with it.

At first Requests author made it so that every attributes in Response are optional, leaving extreme latitude to 3rd parties. And on my responsibilities, I have to keep the compat with that constraint. Some plugin (adapters) fetch "file://" which have zero concept of "status_code" and it's set as None.

In the current context, there is zero chance of having status == None. It's just to make mypy happy there.
We could also add "assert status is not None" as a dev reader barrier.

github/asyncio/MainClass.py
Ousret: In another PR we could introduce "TokenBucketLimiter" ready-to-go-hook as a real/proper alternative

github/Auth.py
Ousret: Indeed, I forgot to replace it with a proper definition.

We can't directly mention "Requester" or "asyncio.Requester" due to cyclic dependency and the reusable aspect of this code.
The compromise here is as follow: I defined a Protocol type checkable "Requester" contract that cover intended usage within Auth.py

regards,

Ousret:
I refreshed this branch against latest main. Everything is still fine locally.
If I can do anything to help push this forward let me know.

regards,

Ousret:
I fixed the codeql error, and the schema generator issue.
As for the target diff coverage, -0.7% is fairly acceptable given the patch diff. I am letting you decide whether it's okay as is or if we need to push it higher.

regards,

EnricoMi:
Hi @sfdye @s-t-e-v-e-n-k , I need your thoughts on this. I am particularly concerned about the following implications:

**Moving from `requests` to `niquests`**

This looks reasonable given `requests` is on eternal feature-freeze. Some migration will have to happen anyways. This definitively deserves a major version release.

**Auto-generated code**

The auto-generated async code is managed by a 5k+ lines of code script: https://github.com/PyGithub/PyGithub/pull/3464/changes#diff-d76dcd6019e3842fac4f2164b2184b581b0a1ce6816090b2773294630bbd4e47

Maintaining this script is vital for the fitness of the `github.asyncio` module.

**Duplicating code**

This PR adds 70k lines of code. It duplicates almost any class and test in this repo. This materializes async code in the repository and changes in the `github` module result in equivalent changes in the `github.asyncio` module. Both are reviewed in pull-requests.

Alternatively, the auto-generated async code is not materialized in the repo and only generated in CI and on release. A Github workflow can generate the async code and provide diffs.

TBH, non-async changes in `github` should result in trivial changes in `github.asyncio`, so reviewing them should not be necessary. Changes to the async auto-generation requires thorough review of the entire implied changes in async code.

**One or two packages**

Either the sync or the async code is used by users. Why always installing both? I'd prefer two packages: `PyGithub` and `PyGithub-asyncio`, both fully self-contained. Users only install what they want.

github/asyncio/MainClass.py
EnricoMi: You say `comply to asyncio I/O scheduler` is more likely the use-case of this asyncio code than unthrottled, high-performance access to te Github API (enterprise only)?

Ousret: If you use PyGithub right now (sync only) inside an ASGI for example, the event loop can't move on, it literally block for x ms until response fully resolved. Can't yield control to another task. Making PyGithub effectively unusable in an ASGI (or any async based concurrent program).

And yes, the capability to "do faster or more efficient" github request must go through async. the threading solution is often underperforming due to the GIL.

regards,

Ousret:
After considerations, I found that the script for async generation isn't as good as I wanted.
I deemed it not great for a merge. The PR itself colapse with it. I don't have nearly as much time as I wanted to dedicate for it. Maybe another time.

Regards,

PR #3468, opened 2026-03-23
Fix: Handle /releases/tags/{tag} URL conversion to canonical form

SebastienSyd:
## Description

Fixes #3467 

When a release is fetched by tag name using `repo.get_release("v1.0")`, the URL was incorrectly set to `/releases/tags/TAG_NAME` instead of `/releases/ID`, which caused subsequent API calls like `get_assets()` to fail with a 404 error.

## Root Cause

The issue was introduced in v2.9.0 with lazy loading for releases (PR #3403). The `__postProcess()` method in `Requester.py` was setting the object's URL to the GET request URL when the response didn't contain a `url` field. For tag-based lookups (`/releases/tags/{tag}`), this resulted in a non-canonical URL being stored.

## Solution

Updated `__postProcess()` in `Requester.py` to:
1. Detect when the response contains an `id` field AND the URL contains `/tags/` (GitHub's name-based lookup pattern)
2. Reconstruct the URL with the canonical ID-based form: `/releases/{id}`

This approach handles GitHub's API pattern for name-based lookups generically.

## Changes

- **github/Requester.py**: Added logic to reconstruct URLs containing `/tags/` with the canonical ID-based form
- **tests/GitRelease.py**: Added `testGetAssetsByTag()` to verify releases fetched by tag can retrieve assets without 404 errors

## Testing

The new test `testGetAssetsByTag()` validates that:
- A release can be fetched by tag name
- Assets can be retrieved from that release without 404 errors
- The correct asset data is returned

## Impact

- Fixes the regression introduced in v2.9.0
- Maintains compatibility with lazy loading
- Doesn't affect other endpoints or non-tag-based lookups


github/Requester.py
```diff
@@ -880,7 +880,15 @@ def __postProcess(
                     if href:
                         data["url"] = href
             else:
-                data["url"] = url
+                # GitHub API provides name-based lookup shortcuts (e.g., /releases/tags/{tag})
```
EnricoMi, line 883: This is very class-specific logic in a very general place.

tests/GitRelease.py
```diff
@@ -228,6 +228,20 @@ def testGetAssets(self):
         self.assertTrue(asset is not None)
         self.assertEqual(asset.id, asset_id)
 
+    def testGetAssetsByTag(self):
```
EnricoMi, line 231: This test does not reproduce the issue. It already passes without your code changes in `Requester.py`.

EnricoMi (commented):
I think the problem is not `Requester` reusing the URL here, but `Release` using the release-by-tag URL as the lazy URL in the first place.

EnricoMi:
Closing in favour of #3469.

PR #3472, opened 2026-03-25
Fix duplicated URL in `OrganizationVariable.edit()` and `OrganizationSecret.edit()`

### LINKED ISSUE #3266: Editing Organization Variable uses incorrect PATCH url
```
    gh_org = gh.get_organization("ORGNAME")
    tagvar = gh_org.get_variable(variable_name="SOME_VAR_NAME")
    tagvar.edit(value="foo")
```
Returns false (and a 404) due to appending `/actions/variables/{self.name}` twice, once at construction of the OrganizationVariable from the Organization.get_variable() and once again in the .edit() function.

Locally, I've patched it to
```
         status, _, _ = self._requester.requestJson(
             "PATCH",
-            f"{self.url}/actions/variables/{self.name}",
+            f"{self.url}",
             input=patch_parameters,
         )
```

Which works for me, but it's not 100% clear whether it's the construct or the edit that's "wrong" in terms of how it should be fixed - so I defer to you guys!

### LINKED ISSUE #3286: OrganizationSecret "edit" method always failing with 404 due malformed url
Test code:

```python
    from github import Auth, Github

    ...
    auth = Auth.Token(token)
    g = Github(auth=auth)
    org = g.get_organization(ORG)
    public_key = org.get_public_key()
    secret = org.get_secret(secret_name="ORG_SECRET")
    result = secret.edit(value=encrypted) # result is always False, due 404 error from Github api
 ```
Debug output:
```
DEBUG:urllib3.connectionpool:https://api.github.com:443 "PATCH /orgs/ORG/actions/secrets/ORG_SECRET/actions/secrets/ORG_SECRET HTTP/1.1" 404 None
```

Issue similar to this one https://github.com/PyGithub/PyGithub/issues/2785

Invalid code: 
https://github.com/PyGithub/PyGithub/blob/main/github/OrganizationSecret.py#L98-L102

Krishnachaitanyakc:
## Summary

- **Fix `OrganizationVariable.edit()`**: The `edit()` method constructed the PATCH URL as `f"{self.url}/actions/variables/{self.name}"`, but `self.url` already contains the full path (e.g., `/orgs/{org}/actions/variables/{name}`). This caused the URL to be doubled to `/orgs/{org}/actions/variables/{name}/actions/variables/{name}`, resulting in a 404 error on every call.
- **Fix `OrganizationSecret.edit()`**: Same issue — the method used `f"{self.url}/{secret_type}/secrets/{self.name}"` but `self.url` already includes the full path, causing a doubled URL like `/orgs/{org}/actions/secrets/{name}/actions/secrets/{name}`.
- **Fix `Organization.create_variable()`**: The returned `OrganizationVariable` object had its `url` attribute set to the organization URL (`self.url`, e.g., `/orgs/{org}`) instead of the full variable URL (`/orgs/{org}/actions/variables/{name}`).

All three fixes align with how the parent `Variable.edit()` and `Secret.delete()` methods already use `self.url` directly.

Fixes #3266
Fixes #3286

## Test plan

- [x] Added `testEditVariable` — verifies `OrganizationVariable.edit()` sends PATCH to the correct URL (`/orgs/{org}/actions/variables/{name}`)
- [x] Added `testEditSecret` — verifies `OrganizationSecret.edit()` sends PATCH to the correct URL (`/orgs/{org}/actions/secrets/{name}`)
- [x] Added `testCreateVariableUrl` — verifies `Organization.create_variable()` returns a variable with the correct full URL
- [x] All 86 existing Organization tests continue to pass


EnricoMi (approved):
Thanks for fixing this!

PR #3475, opened 2026-03-29
Fix OrganizationVariable value variable

layadang:
## Summary

The [schema for getting an organization variable](https://docs.github.com/en/rest/actions/variables?apiVersion=2026-03-10#get-an-organization-variable--code-samples) exposes the "value" field that is not reflected in this repo's schema.

Previous issue encountered:

```python
from github import Github
from dotenv import load_dotenv
import os

load_dotenv()

# Authentication is defined via github.Auth
from github import Auth

# Using an access token
auth = Auth.Token(os.getenv("GITHUB_PAT"))

# Public Web Github
g = Github(auth=auth)

org = g.get_organization("pd03-cs-622")
org_var = org.get_variable("TEST_VARIABLE")
print(org_var.value)
```

Throws this error:

```bash
(***) ➜ *** git:(main) ✗ uv run main.py
  File "/Users/***/***/***/main.py", line 18, in <module>
    print(org_var.value)
          ^^^^^^^^^^^^^
  File "/Users/***/***/***/.venv/lib/python3.14/site-packages/github/Variable.py", line 100, in value
    self._completeIfNotSet(self._value)
                           ^^^^^^^^^^^
AttributeError: 'OrganizationVariable' object has no attribute '_value'. Did you mean: 'value'?
```

Printing `org_var.raw_data` shows:

```python
{'name': 'TEST_VARIABLE', 'value': '123', 'created_at': '2026-03-28T20:22:30Z', 'updated_at': '2026-03-28T20:22:30Z', 'visibility': 'all', 'url': 'https://api.github.com/orgs/pd03-cs-622/actions/variables/TEST_VARIABLE'}
```

With this fix, previous script will print expected value:
```bash
(***) ➜ *** git:(main) ✗ uv run main.py
123
```

(Of course, this is for my organization -- feel free to test on one where you have access to viewing the Actions variables)

## Changes

- Added `self._value` that mirrors what is in [`github/Variable.py`](https://github.com/PyGithub/PyGithub/blob/main/github/Variable.py) for `github/OrganizationVariable.py`
- Updated unit tests

## Verification

All unit tests and pre-commit hook passes. Added additional assert for organization variable value.


EnricoMi:
Thanks for looking into this. The `value` property is already implemented by the base class `Variable`. The underlying issue is that this base class is not properly initialized / reused. This is fixed in #3507.

PR #3479, opened 2026-04-21
Add support for `head_repo` when creating pull request

### LINKED ISSUE #2942: `head_repo` missing from `create_pull`
A while ago, Github allowed [forking repositories within an organisation](https://github.blog/changelog/2022-06-27-improved-innersource-collaboration-and-enterprise-fork-policies/) (with both source and destination being the organisation).

When creating a PR, the `head` can be `{owner}:{branch}` which is not sufficient to distinguish between same-organisation forks. To handle this issue, they added a `head_repo` parameter, which is the name of the source repository for the PR.

pygithub does not currently support this parameter. It should probably be either a string or an organisation.

An other update which might be nice is to allow using a `Branch` as `head`, that way pygithub would have all the metadata for both sides of the PR creation and can infer all the bits and where to put them.

shannon:
I needed the ability to specify the head_repo during a create_pull call in order to create a pull request on a different repo that is in the same organization.

The change is relatively straightforward and just passes the value to the post_parameters.

Resolves: https://github.com/PyGithub/PyGithub/issues/2942


tests/Repository.py
```diff
@@ -1647,6 +1647,16 @@ def testCreatePull(self):
...
 
+    def testCreatePullWithHeadRepo(self):
+        pull = self.repo.create_pull(
+            title="Pull request with head_repo",
+            body="Testing head_repo parameter",
+            base="main",
+            head="patch-1",
+            head_repo="octocat/PyGithub",
+        )
+        self.assertEqual(pull.id, 1436216)
```
EnricoMi:
```suggestion
        self.assertEqual(pull.head.repo.full_name, "octocat/PyGithub")
```

EnricoMi (changes requested):
Lets assert the head repo instead.

EnricoMi (approved):
LGTM!

PR #3401, opened 2025-11-02
Add `WorkflowRun.get_attempt` and `Repository.get_workflow_job` functions

cecheta:
Add `get_workflow_run_attempt` and `get_workflow_job` functions to Repository.

Fixes #3448.
Fixes #2847.

PR #3488, opened 2026-05-19
Add Repository.get_workflow_job to wrap GET /actions/jobs/{job_id}

### LINKED ISSUE #3448: Missing API "Get a job for a workflow run"
Hi,
it looks like [this API][1] is missing from the repository? I tried searching for `actions/jobs`, but no lib result came up: [search link][3].

I would have expected a function in `Repository.py`, e.g. `def get_workflow_job(id_: int) -> WorkflowJob`, like we have for [`get_workflow_run`][2]

[1]: https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#get-a-job-for-a-workflow-run
[2]: https://github.com/PyGithub/PyGithub/blob/main/github/Repository.py#L3799
[3]: https://github.com/search?q=repo%3APyGithub%2FPyGithub%20actions%2Fjobs&type=code

ADiTyaRaj8969:
Fixes #3448.

Adds a `get_workflow_job(id_)` method on `Repository`, mirroring the existing `get_workflow_run(id_)` method one for one. Same signature shape, same one-line URL build, same lazy construction of the return type.

### What's new

```python
def get_workflow_job(self, id_: int) -> WorkflowJob:
    """
    :calls: `GET /repos/{owner}/{repo}/actions/jobs/{job_id} <https://docs.github.com/en/rest/actions/workflow-jobs#get-a-job-for-a-workflow-run>`_
    :param id_: int
    :rtype: :class:`github.WorkflowJob.WorkflowJob`
    """
    assert isinstance(id_, int)
    url = f"{self.url}/actions/jobs/{id_}"
    return github.WorkflowJob.WorkflowJob(self._requester, url=url)
```

Plus the matching imports for `WorkflowJob` (runtime and TYPE_CHECKING).

### Why

The `WorkflowJob` class already exists in PyGithub, but the only way to get one is via `repo.get_workflow_run(run_id).jobs()[0]`. The underlying REST endpoint (`GET /repos/{owner}/{repo}/actions/jobs/{job_id}`) takes a job id directly, so a one call wrapper is useful when you already have the id (for example, from a webhook payload).

### Tests

I followed the precedent of `get_workflow_run`, which also has no dedicated direct test in `tests/Repository.py` and is exercised through other tests. The `WorkflowJob` class itself is already covered by `tests/WorkflowJob.py`. Happy to add a dedicated test with a hand crafted or recorded ReplayData if you would like, just say the word.

### Verification

```
$ python -m pytest tests/Repository.py tests/Workflow.py tests/WorkflowRun.py tests/WorkflowJob.py -q
187 passed
```

Diff: +12 / -0, all pure addition.

### DIFF
--- a/github/Repository.py
+++ b/github/Repository.py
@@ -252,6 +252,7 @@ import github.Team
 import github.Variable
 import github.View
 import github.Workflow
+import github.WorkflowJob
 import github.WorkflowRun
 from github import Consts
 from github.GeneratedReleaseNotes import GeneratedReleaseNotes
@@ -336,6 +337,7 @@ if TYPE_CHECKING:
     from github.Team import Team
     from github.View import View
     from github.Workflow import Workflow
+    from github.WorkflowJob import WorkflowJob
     from github.WorkflowRun import WorkflowRun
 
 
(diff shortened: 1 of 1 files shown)


EnricoMi:
Closing in flavour of #3401.

PR #3494, opened 2026-05-31
Fix GithubException.__str__ decoding bytes data

### LINKED ISSUE #3492: GithubException.__str__ crashes with TypeError when data is bytes
## Description

`GithubException.__str__` raises `TypeError: Object of type bytes is not JSON serializable` when the exception is constructed with `bytes` data.

This happens because `__str__` calls `json.dumps(self.data)` without checking if `data` is `bytes`.

## Root Cause

In `GithubRetry.increment()`, when a 403 response has a non-JSON body (e.g. an HTML rate-limit page), the flow is:

1. `get_content()` returns `bytes` (it returns `response.content` which is always `bytes`)
2. `json.loads(content)` fails because the body is not valid JSON
3. In the `except` block, `GithubException(response.status, content, response.headers)` is raised — with `content` still being raw `bytes`
4. Any subsequent `str(exc)` or logging calls `__str__()` → `json.dumps(self.data)` → **TypeError**

## Minimal Reproducible Example

```python
from github.GithubException import GithubException

exc = GithubException(status=403, data=b"<html>Rate limit exceeded</html>", message="rate limit")
str(exc)
# TypeError: Object of type bytes is not JSON serializable
```

## Expected Behavior

`str(exc)` should return a string representation without crashing, e.g. by decoding bytes before serialization.

## Environment

- PyGithub 2.9.1
- Python 3.14

SAY-5:
`GithubException.__str__` calls `json.dumps(self.data)` directly, which raises `TypeError: Object of type bytes is not JSON serializable` when the exception is constructed from a non-JSON response body that `requests` returns as `bytes` (e.g. an HTML rate-limit page on a 403). Decode bytes to text before serialising so callers can log or stringify these exceptions without crashing.

Fixes #3492.

### DIFF
--- a/github/GithubException.py
+++ b/github/GithubException.py
@@ -102,7 +102,10 @@ class GithubException(Exception):
             msg = f"{self.status}"
 
         if self.data is not None:
-            msg += " " + json.dumps(self.data)
+            data = self.data
+            if isinstance(data, bytes):
+                data = data.decode("utf-8", errors="replace")
+            msg += " " + json.dumps(data)
 
         return msg
 
--- a/tests/Exceptions.py
+++ b/tests/Exceptions.py
@@ -115,6 +115,10 @@ class Exceptions(Framework.TestCase):
     def testExceptionPickling(self):
         pickle.loads(pickle.dumps(github.GithubException("foo", "bar", None)))
 
+    def testBytesData(self):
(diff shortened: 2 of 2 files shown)


EnricoMi (approved):
LGTM!

PR #3502, opened 2026-06-03
Use `@overload` to type `get_user`

xmo-odoo:
LSPs and type checkers complain about `Github(...).get_user().create_repo(...)` and other such calls which only work on the authenticated user.

Because whether the user is named or authenticated depends entirely on a `login` being provided or not, `overload` allows statically separating these two cases and makes typecheckers happier.

Demo on the mypy playground:
- [current (union)](https://mypy-play.net/?gist=6f5eb7865f4549ec041b8df3fee14805)
- [proposal (overload)](https://mypy-play.net/?gist=40490f9b0f58b40b3f60ca908583de24)


EnricoMi:
@xmo-odoo thanks for this fix! mypy is not happy about the main method not being typed:
https://github.com/PyGithub/PyGithub/actions/runs/27138995994/job/80099001604

I have reverted the typing of that main method. Does this work for you?

xmo-odoo:
Ah yes, I'd missed that pygithub is checked with `no-untyped-def` so didn't think to enable that on the playground, sorry about that. It's fine by me.

EnricoMi (approved):
LGTM!

PR #3510, opened 2026-06-10
Lazily complete issues obtained from PRs

xmo-odoo:
Currently when `PullRequest.as_issue` gets called it immediately completes the issue, which triggers an HTTP request which requires `issues: read`.

However this may not be necessary, if the goal is just to e.g. add comments to the PR, that goes through the issue endpoints but it can be accessed with just the issue number and `pull_requests: write` permission (https://docs.github.com/en/rest/issues/comments?apiVersion=2026-03-10#create-an-issue-comment).

By creating the issue uncompleted, `as_issue` can be called with just `pull_requests: read`, and `issue: read` is only necessary when trying to retrieve specific issue attributes.

### DIFF
--- a/github/PullRequest.py
+++ b/github/PullRequest.py
@@ -453,7 +453,7 @@ class PullRequest(CompletableGithubObject):
         """
         :calls: `GET /repos/{owner}/{repo}/issues/{issue_number} <https://docs.github.com/en/rest/reference/issues>`_
         """
-        return github.Issue.Issue(self._requester, url=self.issue_url)
+        return github.Issue.Issue(self._requester, url=self.issue_url, completed=False)
 
     def create_comment(self, body: str, commit: Commit, path: str, position: int) -> PullRequestComment:
         """


xmo-odoo:
Sorting and diffing a PR and its issue (e.g. https://api.github.com/repos/PyGithub/PyGithub/issues/3487 and https://api.github.com/repos/PyGithub/PyGithub/pulls/3487), there's a fair amount of overlap between the two so part of the issue could also be pre-generated from the PR (even for just fields which match exactly), but that's a lot more work and likely a lot less applicable / relevant.

xmo-odoo:
Also "create comment" is probably not the best example given it's reflected on PRs (`create_issue_comment`) but there are issue operations which are not reflected on PRs e.g. lock/unlock.

xmo-odoo:
And while the same could be done to `as_pull_request` on the other side, while a PR always has a matching issue (as far as I know) that's not the case the other way around, so `as_pull_request` can fail, and making it lazy would have the initial call succeed then fail as soon as we try to interact with the PR, which is probably bad.

EnricoMi:
Have you tried lazy mode?
```python
g = Github(auth=auth, lazy=True)
issue = g.get_repo(...).get_pull(...).as_issue()
```
This should not fire a single request.

xmo-odoo:
> Have you tried lazy mode?
> 
> ```python
> g = Github(auth=auth, lazy=True)
> issue = g.get_repo(...).get_pull(...).as_issue()
> ```
> 
> This should not fire a single request.

That does work. Though I still think there's no reason for `as_issue` to immediately resolve the issue even in eager mode.

PR #3516, opened 2026-06-15
Encode boolean query parameters as lowercase

### LINKED ISSUE #3304: Missing Conversion of Python boolean values (True, False) to their JSON-compatible lowercase counterparts (true, false)
## Problem Statement

We need a solution to ensure that the python boolean representations are correctly converted to JSON-compatible lowercase counterparts. Currently, when you pass **query parameters** to the GitHub REST API, such as in the [GET Notifications endpoint](https://docs.github.com/en/rest/activity/notifications#list-notifications-for-the-authenticated-user), there is a bug when you attempt to input true/false values as Python boolean variables. 

For example, in the following function:

https://github.com/PyGithub/PyGithub/blob/b4092b5d5255610fa998f66c720ab5700fddabeb/github/Repository.py#L3801

the query parameters *all* and *participating* are boolean and, in the Repository.py file, there are type assertions to enforce such. However, when the params are actually sent to the `self._requester`, they are not converted correctly to their JSON-compatible lowercase counterparts and are (as such) ignored by the GitHub API. 

-----
## Reproducible Example

The following code should show all user notifications, even those that were previously marked as read. However, the since the python boolean value `True` is not converted correctly, you will only see notifications that are not yet marked as read.

```python
from github import Github, Auth

auth = Auth.Token('<access_token>')
g = Github(auth=auth)
org = g.get_organization('<organization>')
repo = org.get_repo('<repository>')

notifications = repo.get_notifications(all=True)
for notification in notifications:
    print(notification.subject)
```

-----
## Potential Root Cause

The code `urllib.parse.urlencode(parameter_list)` in the `add_parameters_to_url()` function of **Requester.py** Line 494

https://github.com/PyGithub/PyGithub/blob/b4092b5d5255610fa998f66c720ab5700fddabeb/github/Requester.py#L494

does not correctly convert python boolean data types "True" and "False" into json formatted "true" and "false" values respectively. See this [Stack Overflow question](https://stackoverflow.com/questions/66904887/python-urllib-keeps-booleans-capitalized-when-using-urllib-parse-urlencode) on the same topic. Perhaps we can switch to another url parsing function or some custom code solution to solve this problem?

Bolt4243:
Closes #3304.

**Problem:** `add_parameters_to_url` passes Python booleans straight to `urllib.parse.urlencode`, which renders them capitalized. So `get_notifications(all=True)` sends `all=True` instead of `all=true`. GitHub expects lowercase booleans and ignores the capitalized value, so the parameter has no effect.

```python
>>> import urllib.parse
>>> urllib.parse.urlencode([("all", True)])
'all=True'
```

**Fix:** Convert boolean values to lowercase strings before encoding.

**Test:** Added a case to `testAddParametersToUrl` covering `all`/`participating`. It fails on the current code (`all=True&participating=False`) and passes with the fix.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -503,7 +503,12 @@ class Requester:
         url_params = urllib.parse.parse_qs(query)
         # union parameters in url with given parameters, the latter has precedence
         url_params.update(**{k: v if isinstance(v, list) else [v] for k, v in parameters.items()})
-        parameter_list = [(key, value) for key, values in url_params.items() for value in values]
+        # GitHub expects lowercase booleans (true/false) in the query string
+        parameter_list = [
+            (key, str(value).lower() if isinstance(value, bool) else value)
+            for key, values in url_params.items()
+            for value in values
+        ]
         # remove query from url
         url = urllib.parse.urlunparse((scheme, netloc, url, params, "", fragment))
 
--- a/tests/ReplayData/Github.testGetGlobalAdvisoriesManyFilters.txt
+++ b/tests/ReplayData/Github.testGetGlobalAdvisoriesManyFilters.txt
@@ -2,7 +2,7 @@ https
 GET
 api.github.com
 None
-/advisories?type=reviewed&ecosystem=npm&severity=medium&cwes=200%2C900&is_withdrawn=False&affects=directus%2Cmade_up&modified=%3E2023-07-01&sort=updated&direction=desc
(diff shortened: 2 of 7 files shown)


EnricoMi (approved):
LGTM!

EnricoMi:
Thanks for fixing this. You might have to adjust [some files](https://github.com/PyGithub/PyGithub/runs/81491177529) in `tests/ReplayData` that contain `True` or `False` in the requested URLs.

This should fix it: https://github.com/Noethix55555/PyGithub/pull/1

Noethix55555:
@EnricoMi fixed

PR #3518, opened 2026-06-15
Parse only the query string for `PaginatedList.totalCount`

### LINKED ISSUE #3520: PaginatedList.totalCount raises KeyError when the last link lists page first
`PaginatedList` computes `totalCount` with `parse_qs(lastUrl)["page"][0]`, passing the whole URL to `parse_qs` instead of just the query string. This only works because GitHub conventionally puts `page` last in the `rel="last"` link. When `page` comes first (e.g. `...?page=42&per_page=1`), the first parameter key is prefixed with the scheme/host/path and `["page"]` raises `KeyError`.

PR: #3518

Bolt4243:
`PaginatedList` computed `totalCount` with `parse_qs(lastUrl)["page"][0]`, passing the whole URL to `parse_qs`. `parse_qs` expects only a query string, so it splits the entire URL and the first parameter's key ends up prefixed with the scheme/host/path. This only worked because GitHub conventionally puts `page` last in the `rel="last"` link. When `page` comes first (e.g. `...?page=42&per_page=1`), `["page"]` raises `KeyError`.

The fix uses `Requester.get_parameters_of_url`, which extracts the query string before `parse_qs` and is already used by the sibling pagination code in the same file.

Added `testTotalCountWithLastPageAndPageFirst` with a replay fixture whose `rel="last"` link lists `page` first. It raises `KeyError` on the current code and passes with the change; the full `tests/PaginatedList.py` suite passes.

Fixes: #3520, #1006.

### DIFF
--- a/github/PaginatedList.py
+++ b/github/PaginatedList.py
@@ -59,7 +59,6 @@ from __future__ import annotations
 
 from collections.abc import Callable, Iterator
 from typing import Any, Generic, TypeVar, overload
-from urllib.parse import parse_qs
 
 from github import Consts
 from github.GithubObject import GithubObject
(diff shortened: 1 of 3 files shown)


EnricoMi (approved):
LGTM!

PR #3529, opened 2026-06-17
Fix `OrganizationSecret.edit()` to seal the value and use `PUT`

### LINKED ISSUE #3528: OrganizationSecret.edit() uses a non-existent PATCH endpoint and leaks the plaintext secret
## Description

`OrganizationSecret.edit()` is broken in two ways and cannot work against the GitHub API:

1. It issues `PATCH /orgs/{org}/actions/secrets/{name}`, but **there is no PATCH endpoint for organization secrets**. Create/update is `PUT /orgs/{org}/actions/secrets/{name}`.
2. It sends the **plaintext** secret as `value`. The API requires a LibSodium-sealed `encrypted_value` together with the `key_id` of the org public key (exactly like `Organization.create_secret`).
3. There is no way to pass `selected_repository_ids` when `visibility="selected"`.

## Current code (github/OrganizationSecret.py)

```python
patch_parameters = {"name": self.name, "value": value, "visibility": visibility}
status, _, _ = self._requester.requestJson("PATCH", self.url, input=patch_parameters)
```

## Expected

Mirror `Organization.create_secret`: fetch the org public key for the `secret_type`, `encrypted_value = public_key.encrypt(value)`, then `PUT {key_id, encrypted_value, visibility, selected_repository_ids}`.

## Version

main

Bolt4243:
Fixes #3528

## Problem

`OrganizationSecret.edit()` issued `PATCH /orgs/{org}/actions/secrets/{name}` (no such endpoint exists) and sent the **plaintext** secret as `value` instead of a LibSodium-sealed `encrypted_value` + `key_id`. It also offered no way to set `selected_repository_ids` for `visibility="selected"`.

## Fix

Mirror `Organization.create_secret`:
- Fetch the org public key for the `secret_type` (new `get_public_key` helper, derived from the secret URL).
- `encrypted_value = public_key.encrypt(value)`.
- `PUT` body `{key_id, encrypted_value, visibility}`, adding `selected_repository_ids` when `visibility="selected"` (ints for actions, strings for dependabot, matching create_secret).
- Add a `selected_repositories` parameter.
- Fix the `:calls:` docstrings to PUT.

Also fixes `selected_repositories` property to use the completing `selected_repositories_url` accessor.

## Tests

`testEditSecret` now mocks the public-key fetch + encryption and asserts (via replay data) the request is a PUT carrying `encrypted_value`+`key_id` with the plaintext `value` absent. Added `testEditSecretSelected` covering `selected_repository_ids`. Both fail on HEAD (PATCH/plaintext) and pass with the fix.


KeenHero:
Thanks for tackling this sealing the value and switching the update endpoint to `PUT` are the important parts of the fix.

I checked out `c473ac51` and ran the focused tests (`tests/Organization.py -k EditSecret`): both pass. Ruff on the changed Python files and mypy on `OrganizationSecret.py` are also clean.

One behavior still looks worth correcting before merge: the new code calls `requestJson(...)` and returns `status in (201, 204)`. That converts unexpected responses such as 401/403/422 into `False` instead of raising the normal `GithubException`, which is the same failure-hiding pattern being addressed in #2760/#3550. Could this use `requestJsonAndCheck("PUT", ...)` and return `True` after it succeeds (or otherwise explicitly validate 201/204 and raise for everything else)? A failure-path test would make that guarantee clear.

This PR also includes the lazy `selected_repositories` fix and selected-repository editing support that overlap with #3531. It may be worth coordinating/splitting that portion so the security/correctness fix here can land without the two PRs competing.

github/OrganizationSecret.py
EnricoMi, line 42:
```suggestion
    from github.PublicKey import PublicKey
    from github.Repository import Repository
```

github/OrganizationSecret.py
```diff
@@ -90,32 +91,57 @@ def edit(
...
-            "PATCH",
+            "PUT",
             self.url,
-            input=patch_parameters,
+            input=put_parameters,
         )
-        return status == 204
+        return status in (201, 204)
+
+    def get_public_key(self, secret_type: str = "actions") -> github.PublicKey.PublicKey:
```
EnricoMi:
```suggestion
    def get_public_key(self, secret_type: str = "actions") -> PublicKey:
```

EnricoMi (changes requested):
Very good contribution, only minor code style changes required.

EnricoMi (approved):
LGTM! Thanks!

PR #3537, opened 2026-06-17
Fix doubled slash in request path for `base_url` with trailing slash

### LINKED ISSUE #3536: Trailing slash in base_url produces a doubled slash in every request path
### Bug

A trailing slash in `base_url` is carried into the request path unnormalized, producing a doubled slash in every request.

`Requester.__init__` sets `self.__prefix = o.path` with no normalization, and `__makeAbsoluteUrl` does `url = f"{self.__prefix}{url}"` for leading-slash URLs. So:

| `base_url` | `__makeAbsoluteUrl("/repos/o/r")` |
|---|---|
| `https://api.github.com/` | `//repos/o/r` |
| `https://ghe.example.com/api/v3/` | `/api/v3//repos/o/r` |
| `https://api.enterprise.ghe.com/github-api/` | `/github-api//repos/o/r` |

github.com tolerates the `//`, but GitHub Enterprise and many reverse proxies treat `//` as a distinct path and return 404.

`Requester.get_graphql_prefix` has the same trailing-slash doubling for a custom prefix:

```python
>>> Requester.get_graphql_prefix("/foo/")
'/foo//graphql'
```

### Reproduce (offline, no network)

```python
import github
r = github.Github(base_url="https://ghe.example.com/api/v3/").requester
print(r._Requester__makeAbsoluteUrl("/request"))   # -> '/api/v3//request'  (expected '/api/v3/request')
```

### Note

The existing `tests/Requester.py::testMakeAbsoluteUrl` did not catch this because it used `assert "<literal>", requester.__makeAbsoluteUrl(...)`. The literal is always truthy and the real value becomes the assertion *message*, so the test asserted nothing.

### Fix

Normalize the prefix once in `__init__` (`o.path.rstrip("/")`) and strip trailing slashes in `get_graphql_prefix` before the final concat. Also convert the no-op test to real `assertEqual` comparisons. PR incoming.

Bolt4243:
Fixes #3536.

### Problem

A trailing slash in `base_url` was carried into `Requester.__prefix` unnormalized, so `__makeAbsoluteUrl` produced a doubled slash in every leading-slash request path:

- `https://api.github.com/` -> `//request`
- `https://ghe.example.com/api/v3/` -> `/api/v3//request`
- `https://api.enterprise.ghe.com/github-api/` -> `/github-api//request`

github.com tolerates `//`, but GitHub Enterprise and reverse proxies commonly 404 on it. `get_graphql_prefix` had the same doubling for a custom prefix (`/foo/` -> `/foo//graphql`).

### Fix

- Normalize the prefix once in `__init__` via `o.path.rstrip("/")`.
- Strip trailing slashes in `get_graphql_prefix` before the final concat.

Full absolute URLs (pagination next/last links) and the redirect path are unaffected because they take `o.path` for non-leading-slash URLs.

### Tests

`tests/Requester.py::testMakeAbsoluteUrl` was a no-op: it used `assert "<literal>", value`, where the literal is always truthy and the real value becomes the assertion message, so nothing was checked. Converted to real `assertEqual` comparisons covering the default base_url, a trailing-slash base_url, a GHE `/api/v3/` base_url, a custom-prefix base_url, and full-URL passthrough.

Verified the corrected test fails on `main` (`'/request' != '//request'`) and passes with this fix. `tests/Requester.py` is green (32 passed, 605 subtests). black/isort clean on the changed lines.


EnricoMi (approved):
LGTM!

PR #3539, opened 2026-06-26
feat: repositories-scoped app installation tokens

guillp:
# Add support for Repositories-scoped App Installation tokens.

This adds a `repositories` optional parameter to `GitHubIntegration.get_access_token()`, which accepts a list of repository names.
This list will be passed as `repositories` parameter to the GH API call, so that the resulting token only has access to those specific repositories instead of all repositories the Installation has access to.
This allows least-privilege tokens.

As documented here:
https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app

### DIFF
--- a/github/GithubIntegration.py
+++ b/github/GithubIntegration.py
@@ -262,7 +262,7 @@ class GithubIntegration:
         return self.auth.create_jwt(expiration)
 
     def get_access_token(
-        self, installation_id: int, permissions: dict[str, str] | None = None
+        self, installation_id: int, permissions: dict[str, str] | None = None, repositories: list[str] | None = None
     ) -> InstallationAuthorization:
         """
         :calls: `POST /app/installations/{installation_id}/access_tokens <https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app>`
@@ -274,6 +274,11 @@ class GithubIntegration:
             raise GithubException(status=400, data={"message": "Invalid permissions"}, headers=None)
 
         body = {"permissions": permissions}
+        if repositories is not None:
+            if not isinstance(repositories, list):
+                raise GithubException(status=400, data={"message": "Invalid repositories"}, headers=None)
+            
+            body["repositories"] = repositories
         headers, response = self.__requester.requestJsonAndCheck(
             "POST",
             f"/app/installations/{installation_id}/access_tokens",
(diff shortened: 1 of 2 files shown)


tests/GithubIntegration.py
```diff
@@ -343,6 +343,14 @@ def testGetAccessTokenWithInvalidPermissions(self):
         self.assertEqual(raisedexp.exception.message, "The permissions requested are not granted to this installation.")
         self.assertEqual(raisedexp.exception.status, 422)
 
+    def testGetAccessTokenWithInvalidRepositories(self):
```
EnricoMi, line 346: Can you record the necessary replay data for this test as well?

dbermuehler:
Hi @guillp, thank you for your contribution! I would be also really interested in this being added as a feature to PyGithub. Did you have a chance to look at @EnricoMi's comment?

tests/GithubIntegration.py
EnricoMi: Looks like this one is only testing with invalid value for `repositories`. Probably no test data needed here.

Which brings me to: can you please add a test with valid repository values and provide test data for that test?

github/GithubIntegration.py
```diff
@@ -262,7 +262,7 @@ def create_jwt(self, expiration: int | None = None) -> str:
         return self.auth.create_jwt(expiration)
 
     def get_access_token(
-        self, installation_id: int, permissions: dict[str, str] | None = None
+        self, installation_id: int, permissions: dict[str, str] | None = None, repositories: list[str] | None = None
```
EnricoMi, line 265:
```suggestion
        self, installation_id: int, permissions: dict[str, str] | None = None, repositories: list[str | Repository] | None = None
```
needs `from github.Repository import Repository` above.

github/GithubIntegration.py
```diff
@@ -274,6 +274,11 @@ def get_access_token(
             raise GithubException(status=400, data={"message": "Invalid permissions"}, headers=None)
 
         body = {"permissions": permissions}
+        if repositories is not None:
+            if not isinstance(repositories, list):
+                raise GithubException(status=400, data={"message": "Invalid repositories"}, headers=None)
+            
+            body["repositories"] = repositories
```
EnricoMi, line 281:
```suggestion
        if repositories is not None:
            if not is_list(repositories, int | Repository):
                raise GithubException(status=400, data={"message": "Invalid repositories"}, headers=None)
            repositories = [repo.id if isinstance(repo, Repository) else repo for repo in repositories]
            body["repositories"] = repositories
```
needs `from github.GithubObject import is_list` above.

EnricoMi (changes requested):

PR #3540, opened 2026-06-27
Add `max_rate_limit_wait` to `GithubRetry` to cap rate limit backoff

### LINKED ISSUE #3473: Allow capping X-RateLimit-Reset header values
Currently there are a couple paths for retries:
- if `Retry-After` response header is present it uses the default urllib3 behavior which allows some configuration via `respect_retry_after_header` and `retry_after_max` params to `GithubRetry`
- if `X-RateLimit-Reset` response header is present, `GithubRetry` strictly honors the field with no customization

The problem I'm seeing is that sometimes `X-RateLimit-Reset` sets values really far into the future which we do not want to wait for. It would be nice and consistent with the urllib3 behavior if we can configure a cap for this value as well. I tried subclassing `GithubRetry` but the way this logic is structured in the `increment()` method makes it difficult to cleanly override only this behavior.

tanistheta:
## Problem

GithubRetry currently waits until X-RateLimit-Reset (primary rate limit) or
secondary_rate_wait (secondary rate limit) before retrying, with no way to cap
that wait. If GitHub reports a reset far in the future, callers with hard
timeouts (e.g. Celery tasks) end up blocked well past their own deadline.

## Approach

Adds an optional `max_rate_limit_wait` parameter to `GithubRetry.__init__`.
When the computed backoff for either a primary or secondary rate limit error
would exceed this cap, `GithubRetry.increment()` raises a new
`RateLimitExceededExceedsMaxWait` exception (subclass of
`RateLimitExceededException`) instead of waiting. The exception exposes
`.wait` (the backoff that would have been required) and the usual
`.status` / `.data` / `.headers` from `GithubException`, so callers can
inspect the raw rate-limit response and decide how to handle it themselves.

Defaults to `None` (no cap), so existing behavior is unchanged unless a
caller opts in.

## Tests

Added three tests in `tests/GithubRetry.py`:
- `test_primary_rate_error_exceeds_max_wait` — primary rate limit backoff
  exceeding the cap raises with the correct `.wait`/`.status`/`.data`
- `test_primary_rate_error_within_max_wait` — backoff under the cap still
  retries normally
- `test_secondary_rate_error_exceeds_max_wait` — same cap behavior on the
  secondary rate limit path

Fixes #3473

### DIFF
--- a/github/GithubRetry.py
+++ b/github/GithubRetry.py
@@ -45,7 +45,7 @@ from urllib3.connectionpool import ConnectionPool
 from urllib3.exceptions import MaxRetryError
 from urllib3.response import HTTPResponse
 
-from github.GithubException import GithubException
+from github.GithubException import GithubException, RateLimitExceededExceedsMaxWait
 from github.Requester import Requester
 
 DEFAULT_SECONDARY_RATE_WAIT: int = 60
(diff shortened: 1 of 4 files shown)


github/GithubRetry.py
```diff
@@ -72,12 +72,22 @@ class GithubRetry(Retry):
...
+        secondary_rate_wait: float = DEFAULT_SECONDARY_RATE_WAIT,
+        max_rate_limit_wait: float | None = None,
+        **kwargs: Any,
+    ) -> None:
         """
         :param secondary_rate_wait: seconds to wait before retrying secondary rate limit errors
+        :param max_rate_limit_wait: maximum seconds to wait for a primary or secondary rate limit reset;
+            if the required backoff would exceed this, raises
+            :class:`~github.GithubException.RateLimitExceededExceedsMaxWait` instead of waiting.
+            ``None`` (the default) means no cap is applied.
```
EnricoMi, line 87: Should be explicit here that if max_rate_limit_wait < secondary_rate_wait, any secondary rate limit error will raise an exception and never wait.

EnricoMi (changes requested):
Thanks for raising this, I think this is very useful given `retry-after` headers can be an hour in the future. Minor documentation comment.

github/GithubRetry.py
EnricoMi, line 87:
```suggestion
            Note that ``max_rate_limit_wait < secondary_rate_wait`` causes any secondary rate limit error to raise an exception and never wait.
            ``None`` (the default) means no cap is applied.
```

EnricoMi (approved):
LGTM!

djsigmann:
Thanks for merging this :), when might we see a new release with these changes?

PR #3543, opened 2026-07-03
docs: add exception handling example

zain-asif-dev:
## Problem

#874 collects requests for more usage examples, and more than one comment there
(e.g. from @NicoWeio) specifically asked for an example of catching PyGithub
exceptions, such as `github.GithubException.UnknownObjectException`. The
`utilities` page documents the exception classes via their docstrings, but
there was no example showing how to actually catch them in practice.

## Change

Adds `doc/examples/Exceptions.rst` with:

- catching the base `GithubException` when you don't need to distinguish
  between failure types
- catching the most common subclasses separately (`UnknownObjectException`,
  `BadCredentialsException`, `RateLimitExceededException`)
- reading `status`/`data`/`headers` off a caught exception, e.g. to build a
  custom error message

Registered the new page in `doc/examples.rst`.

## Testing

Built the docs locally with `sphinx -b html doc <out>`; the new page renders
correctly (cross-references to the exception classes resolve) and introduces
no new warnings compared to a build without this change.


EnricoMi (approved):
LGTM! Thanks!

PR #3544, opened 2026-07-03
fix: totalCount capped to 1000 for large search results

### LINKED ISSUE #1309: search_issues() result has totalCount maxed out to 1000
The paginated list returned from a search_issues() will report a totalCount that maxes out to 1000 when a search query results in more than 1000 items. The totalCount() method in PaginatedList() will get headers and data from result of a requester.requestJsonAndCheck() using the first URL. The logic then tests to see if 'link' is not in headers which in my case there is a 'link' so it falls to the else side and parses the lastUrl line for 'page' which is 1000.

Problem though is that in my search, the original 'data' structure has a valid 'total_count' field of 3041 so I'm not sure why the logic tries to derive something from the lastUrl which is generating an incorrect value?

zain-asif-dev:
## Problem

Fixes #1309.

When `PaginatedList.totalCount` is accessed *before* any page has been fetched, it's computed in `_fetchFirstElement()` by fetching a single item (`per_page=1`) and deriving the count from the page number in the `last` Link header. For search endpoints, GitHub only lets you page through the first 1000 results of a search, so this derived count is capped at 1000 even when the search actually matched far more items (as reported in this issue, and reproduced by other commenters, e.g. `repos.totalCount` showing `1000` before iterating, then the real total after).

The response body's `total_count` field (or whichever field `total_count_item` points at, e.g. `total_files`/`total_commits` for `Commit.files`/`Comparison.commits`) already carries the real, uncapped total. This is exactly what `PaginatedList._getPage()` uses once any page has actually been fetched — so depending on whether you read `totalCount` before or after iterating, you could silently get two different answers (the wrong capped one, then the correct one).

## Fix

In `_fetchFirstElement()`, prefer the count from the response body (`total_count_item`) over the one derived from the `last` Link header. This matches the logic already used in `_getPage()`, so `totalCount` is now consistent and correct regardless of access order. The Link-header-derived count is still used as a fallback for endpoints that don't return a body total (plain list endpoints only return a bare JSON array).

## Testing

Added `testTotalCountWithSearchResultsCappedAtMax`, with a fixture whose `last` link is capped at page 1000 while the body's `total_count` is 3041. Verified this test fails with `1000 != 3041` on the old code and passes with the fix.

- `pytest tests/` — 1124 passed (+ 649 subtests), no regressions
- `ruff check`, `black --check` (pinned 23.12.0), `mypy` — all clean on the changed files


EnricoMi:
Thanks for looking into this! What are we going to about these two semantics? Either number might be useful for users. See #3452.

KeenHero:
Thanks, I think #3544 and #3452 can be separated without changing the meaning of `totalCount` in this patch.

The stable choice for `totalCount` is the API response field `total_count`. That matches what `_getPage` already records, so callers get the same value whether they inspect `totalCount` before or after iteration. Keeping the Link header derived value would preserve the 1000 cap, but it would also preserve the current access order dependent behavior.

I would treat `retrievableCount` as a separate optional property in #3452. It should not be computed as `min(totalCount, 1000)`, because retrieval limits vary by endpoint and query. It can remain `None` until pagination exhaustion or an endpoint specific cap makes the value knowable.

I checked out `f599eb36` and ran the `TotalCount` tests: 5 passed. Ruff and mypy are clean on the changed files. On that basis, #3544 looks sound as the `totalCount` consistency fix, with `retrievableCount` following separately.

PR #3548, opened 2026-07-09
Allow creating custom property with `values_editable_by` param

### LINKED ISSUE #3416: Creation of Custom Property not possible when "values_editable_by" is set
Hi there,

Is there any reason why the `create_custom_property` function does not allow the creation of a custom property with the `values_editable_by` parameter being set?

As a workaround the `create_custom_properties` function can be used, without this limitation.

https://github.com/PyGithub/PyGithub/blame/e79d9bc1e5d786f4b76df0c0e1304829fd6a3aa1/github/Organization.py#L1677

chawlajay9:
The `create_custom_property ()` was not allowing the creation of the property when the `values_editable_by ` parameter was set. 

This fix removed the line which checks  if  `values_editable_by ` Param is `NotSet`. This is blocking the creation of the custom property with other options of  `values_editable_by ` Param. 

Added the extra line  values_editable_by="org_actors" in testcases and replay text file for testing purpose. 

Fix: #3416.


EnricoMi (approved):
LGTM!

PR #3550, opened 2026-07-21
Fix PullRequest.is_merged() swallowing real HTTP errors

### LINKED ISSUE #2760: PullRequest.is_merged does not handle token expiration

comNayaK:
Fixes #2760

## Problem
`PullRequest.is_merged()` currently calls `requestJsonAndCheck` and treats any exception from that call as "not merged," returning `False`. In practice, the GitHub API returns:
- `204 No Content` if the PR is merged
- `404 Not Found` if the PR is not merged
- Any other status (e.g. `500`, `502`, rate-limit responses, etc.) on a genuine error

Because `requestJsonAndCheck` only accepts one "success" status code, both the legitimate `404` and real server/network errors raise a `GithubException`, and `is_merged()` catches all of them the same way, silently reporting `False` even when something has actually gone wrong.

## Fix
- Added `Requester.requestJsonAndValidateStatus()`, which behaves like `requestJsonAndCheck` but accepts a set of acceptable status codes instead of just one, and raises `GithubException` only if the response status isn't in that set.
- Updated `PullRequest.is_merged()` to call the new method with `valid_codes={204, 404}`, then return `True`/`False` based on which of those two codes was actually returned, instead of based on whether an exception was raised.

This keeps the intended behavior for the two expected codes, but now lets any other error propagate normally instead of being masked as "not merged."

## Notes
- This follows the fix direction suggested by a maintainer in #2760; I wrote a fresh implementation rather than reviving #2763, which has been open and merge-conflicted since 2023.
- Ran `pytest tests/PullRequest.py -k is_merged` and the full `tests/PullRequest.py` suite locally — all tests passed.
- No new test was added for the "real error now raises" branch, since that would require a new recorded HTTP fixture for a non-204/404 response. I can add one if maintainers want it.


EnricoMi:
Hi @comNayaK, thanks for the fix.

There should be more places where status is checked this way. Can you move those to the new pattern as well?

Please record a non-204/404 response as suggested.

KeenHero:
Thanks for working on this. I reviewed this together with #3542, since both address the same error handling family.

The broader coverage requested here already exists in #3542, which updates 34 call sites and adds failure path tests. I checked out that branch and ran the targeted tests: 13 passed. Ruff and mypy are clean there.

I think the strongest path is to keep the generic validator design from this PR, then use the call site inventory and tests from #3542. A validator that accepts explicit status sets is safer than a universal boolean helper, because not every endpoint gives 404 the same meaning.

The two current CI failures in this PR are mechanical. Mypy reports that `valid_codes` is declared as `set[int]` but defaults to `frozenset[int]`. The pre commit job reports only the blank line that docformatter adds before the closing docstring quotes.

Choosing one PR as the implementation and using the other as supporting inventory would avoid parallel work for the two contributors.

comNayaK:
Hi @EnricoMi  and @KeenHero , apologies for the delay in getting back to this! I've updated the PR and marked it as ready for review.

Following your advice, I kept the requestJsonAndValidateStatus design and migrated the other boolean status check methods to use it (borrowing the inventory from #3542 as suggested). I also just pushed a commit to fix the mechanical mypy and docformatter CI failures.

Let me know if there is anything else needed before we can get this merged!

PR #3551, opened 2026-07-26
Add support for Python 3.15 and drop EOL 3.9

hugovk:
The Python 3.15 beta is out and release candidate is nearly here! 🚀 

The 3.15 release manager (👋 hi, that's me!) [said](https://discuss.python.org/t/python-3-15-0-beta-4-is-here):

> We ***strongly encourage*** maintainers of third-party Python projects to ***test with 3.15*** during the beta phase and report issues found to [the Python bug tracker](https://github.com/python/cpython/issues) as soon as possible. While the release is planned to be feature-complete entering the beta phase, it is possible that features may be modified or, in rare cases, removed up until the start of the release candidate phase (2026-08-04). Our goal is to have ***no ABI changes*** after beta 4 and as few code changes as possible after the first release candidate. To achieve that, it will be ***extremely important*** to get as much exposure for 3.15 as possible during the beta phase.
> 
> This includes creating pre-release wheels for 3.15, as it helps other projects to do their own testing. However, we recommend that your regular production releases wait until 3.15.0rc1, to avoid the risk of ABI breaks.

Also drop EOL 3.9: https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         # keep in sync with tox.ini [gh-actions] section
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.15"]
         os: ["ubuntu-latest"]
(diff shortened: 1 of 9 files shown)


EnricoMi (approved):
LGTM! Thanks for the upgrade!

PR #3557, opened 2026-08-13
Migrate from reqests to niquests

EnricoMi:
The `requests` package is central to this project to communicate with the Github REST API. That package is under permanent feature freeze, meaning there are only bug and security fixes but no new features enhancing performance or future compatibility:

- https://github.com/psf/requests/issues/6664
- https://3.python-requests.org/dev/todo/#feature-freeze

This migrates to PyGithub to niquests, a fully compatible fork that moved beyond requests feature-wise:
https://github.com/jawah/niquests

This reuses the migration bit from #3464.

### DIFF
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -1433,6 +1433,7 @@ class WithRequester(Generic[T]):
     def requester(self) -> Requester:
         return self.__requester
 
-    def withRequester(self, requester: Any) -> WithRequester[T]:
+    def withRequester(self, requester: Requester) -> WithRequester[T]:
+        assert isinstance(requester, Requester), requester
         self.__requester = requester
         return self
--- a/github/Requester.py
+++ b/github/Requester.py
@@ -98,14 +98,13 @@ import time
 import urllib
 import urllib.parse
 from collections import deque
-from collections.abc import Callable, ItemsView, Iterator
+from collections.abc import Callable, Iterator
 from datetime import datetime, timezone
 from io import IOBase
 from typing import TYPE_CHECKING, Any, BinaryIO, Deque, Generic, TypeVar
 
-import requests
-import requests.adapters
-from urllib3 import Retry
+import niquests
+import niquests.adapters
 
 import github.Consts as Consts
 import github.GithubException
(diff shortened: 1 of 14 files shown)


EnricoMi:
@Ousret I have extracted your migration from `requests` to `niquests` from your PR. Does this look alright?

pyproject.toml
```diff
@@ -30,10 +30,9 @@ dynamic = ["version"]
 
 dependencies = [
   "pynacl>=1.4.0",
-  "requests>=2.14.0",
+  "niquests>=3,<4",
```
Ousret:
```suggestion
  "niquests>=3.12,<4",
```

some API exposed are available since 3.12, so it would be safe to assume this as the lowest compatible. (was my own mistake before, forgot to review the lower bound)

github/Requester.py
```diff
@@ -1442,7 +1433,6 @@ def __init__(self) -> None:
     def requester(self) -> Requester:
         return self.__requester
 
-    def withRequester(self, requester: Requester) -> WithRequester[T]:
-        assert isinstance(requester, Requester), requester
+    def withRequester(self, requester: Any) -> WithRequester[T]:
```
Ousret: I believe this is unrelated to the patch purpose. Could be shipped independently of that PR.

Ousret: A protocol based Requester to avoid the circle dependency in async.

Ousret (changes requested):
Took the time to check the diff, overall, LGTM. Two comments left.

Regards,

github/Requester.py
EnricoMi: Yeah, that was not intentional. Well spotted!

pyproject.toml
EnricoMi: Thanks for the suggestion!

github/Requester.py
EnricoMi:
```suggestion
    def withRequester(self, requester: Requester) -> WithRequester[T]:
        assert isinstance(requester, Requester), requester
```

Ousret (approved):
👌

PR #3558, opened 2026-08-16
Align get_global_advisories parameters with the API

ckarnell:
This brings `get_global_advisories` in line with the `GET /advisories` API.

`per_page` and `keywords` were both in the signature with no plumbing behind them, so passing either sent the same request as passing nothing. So this removes them both. Page size is set on the `Github` instance for every `PaginatedList` in the project, and `keywords` isn't in the API at all.

Two go the other way. `epss_percentage` and `epss_percentile` are documented on `GET /advisories` and were missing. Both are `Opt[str]`, since the API takes range syntax like `>=0.5` in the same style as `published` and `updated`.

Both have been unplumbed since #3158, which added them to the signature and stopped there.

Test is `testGetGlobalAdvisoriesByEPSS`, reusing the existing replay file with the query string updated. Suite goes 1123 to 1124. Reverting only `MainClass.py` fails that one test on `unexpected keyword argument`, so the new parameters are the reason it passes. No token here to re-record replay data. `black` 23.12.0 and `mypy` are clean.

The response body in that replay file is the one already in the repo, and only the request line changed.


EnricoMi:
The `per_page` option in this method should be removed. In this project, `per_page` is controlled for all `PaginatedList`s when creating the `Github` instance:

```python
g = github.Github(per_page=100)
```
https://pygithub.readthedocs.io/en/stable/utilities.html#pagination

You are right, parameter `keywords` is not documented and should be removed as well.

Options `epss_percentage` and `epss_percentile` are documented but missing from the implementation. Maybe you can add then while you are touching this code are.

ckarnell:
That makes sense, and it's a better fix than mine. Pushed a rewrite.

`per_page` and `keywords` are both gone from the signature. `get_global_advisories` was the only method returning a `PaginatedList` that took a per-call `per_page`, so it was the odd one out, not the one with a missing feature.

Added `epss_percentage` and `epss_percentile`. Both are `Opt[str]` and not a numeric type, since the API takes range syntax like `>=0.5` in the same style as `published` and `updated`.

Renamed the test to `testGetGlobalAdvisoriesByEPSS` and reused the replay file. Suite is 1124 passing. Reverting just `MainClass.py` fails that one test on `unexpected keyword argument`, so the new parameters are the reason it passes. `black` 23.12.0 and `mypy` are clean.

Dropping the two parameters does turn `get_global_advisories(per_page=5)` from a silent no-op into a `TypeError`. Nobody could have depended on either doing anything, so I haven't added a deprecation cycle, but I couldn't find precedent in the repo for removing a parameter outright and you may want one.

PR #3565, opened 2026-08-25
Make niquests dependency optional

EnricoMi:
This allows to choose between running PyGithub with `requests` or `niquests`. This provides full control over the HTTP library dependency used.

**This is a breaking change.** Installing `PyGithub` does not install `requests` as a dependency any more.

To restore this behaviour, instead of

    pygithub

use

    pygithub[requests]

in your `requirements.txt` or dependencies list. For instance, instead use

    pip install pygithub[requests]

Co-authored-by: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,17 +19,20 @@ jobs:
 
   test:
     runs-on: ${{ matrix.os }}
-    name: test (Python ${{ matrix.python-version }} on ${{ matrix.os-label }})
+    name: test (${{ matrix.request-lib }} and Python ${{ matrix.python-version }} on ${{ matrix.os-label }})
     strategy:
       fail-fast: false
       matrix:
-        # keep in sync with tox.ini [gh-actions] section
+        # keep in sync with tox.ini [gh-actions] / [gh-actions:env] sections
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14", "3.15"]
+        request-lib: ["requests", "niquests"]
         os: ["ubuntu-latest"]
         os-label: ["Ubuntu"]
         include:
-          - {python-version: "3.10", os: "windows-latest", os-label: "Windows"}
-          - {python-version: "3.10", os: "macos-latest", os-label: "macOS"}
+          - {python-version: "3.10", request-lib: "requests", os: "windows-latest", os-label: "Windows"}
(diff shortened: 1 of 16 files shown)


EnricoMi:
@Ousret What do you think? This allows to use either `requests` or `niquests`. Is the `github/requestlib/__init__.py` sane?

EnricoMi:
Rational: The PyGithub package might be used in an application that itself depends on `requests`, so moving from `requests` to `niquests` for good might break user code. This allows a smoother transition.

Ousret:
> What do you think? This allows to use either requests or niquests. Is the github/requestlib/__init__.py sane?

It does answer your concern about the potential breaking change.

A couple of notes I have for you:

- Do not make your compat module public, declare it private by all means. People will eventually accidentally rely on it. And believe me, you don't want them to.
- Requests is almost impossible to dodge in any environment due to his wide adoption from the early days. Even if one actually never use it. So the Niquests path will be very difficult to leverage. If not impossible for some.
- Maybe the compat module is a bit overengineered. Prefer a mere try: import niquests except: import requests. I believe it will yield the smallest patch forward and avoid an additional module to maintain.

I think at first we need to carefully identify what breaking changes this would introduce (the Niquests sole path) in order to better protect end users. What API surface currently exposed can be problematic when one use explicitly Requests right now? What deprecation can the project foresee to guide users into not feel for the single http client choice?

Regards,

Ousret:
> The PyGithub package might be used in an application that itself depends on requests, so moving from requests to niquests for good might break user code. This allows a smoother transition.

Yeah, for the dep tree, yes, but in practical way, we almost all depends on httpx, requests and aiohttp and there is nothing we can do about it. The fragmentation is there and will stay for a long time. I believe PyGithub does a pretty good job at hiding what http engine it have.

Right now, If I am not mistaken, Requests is directly depended upon in Requester and Retry logic.
And none of them expose a public arg that allows people to inject there own Requests instances.

Regards,

Ousret:
Finally, Niquests is so close to the Requests APIs, and knowing that PyGithub doesn't allow injecting Requests instances plus doesn't leak exceptions from below, this would be as transparent as it can be. But that's only a naive way to perceive it right away. Maybe a deeper investigation will yield something.

regards,

edit: trust your instinct more on this, you are actually the best knowledgeable person to take care of pygithub end users. And given the "_elephant in the room_ **cough** maintain Niquests" I am most certainly not the most objective person here.

PR #3568, opened 2026-08-31
Add `signature` parameter to `Repository.create_git_commit`

j-flat:
`POST /repos/{owner}/{repo}/git/commits` accepts an optional `signature` field, which Github stores in the `gpgsig` header of the commit it creates and uses to verify it. `Repository.create_git_commit` did not expose that field, so commits created through PyGithub could not be signed and were always reported as unverified.

cf. https://docs.github.com/en/rest/git/commits#create-a-commit

Added an optional `signature: Opt[str]` argument to `Repository.create_git_commit`, handled the same way as the existing `author` and `committer` arguments: validated with `is_optional` and only added to the post parameters when defined. The argument comes last, so existing callers are unaffected.

Also added an example to `doc/examples/Commit.rst`. Building the payload to sign is the part that is easy to get wrong, so the example spells out the canonical commit format and notes that the author and committer in that payload must match the ones sent to the API, otherwise Github considers the signature invalid.

`Repository.testCreateGitCommitWithSignature` creates a commit with an armored signature and asserts the sha and the returned verification. Removing the two lines that send the parameter makes it fail on the recorded request body, so the test covers the change rather than just the happy path.

### DIFF
--- a/doc/examples/Commit.rst
+++ b/doc/examples/Commit.rst
@@ -26,3 +26,74 @@ Get commit date
     2018-10-11 03:04:52
     >>> print(commit.commit.committer.date)
     2018-10-11 03:04:52
+
+Create a signed commit
+----------------------
+
+:meth:`github.Repository.Repository.create_git_commit` takes an optional ``signature`` argument,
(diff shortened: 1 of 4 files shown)


j-flat:
Related issue #3569

PR #3571, opened 2026-09-06
fix: list installation repositories with user authentication

### LINKED ISSUE #3546: `Installation.get_repos` needs to differentiate between IAT access and UAT access
`get_repos` calls [`/installation/repositories`](https://docs.github.com/en/rest/apps/installations?apiVersion=2026-03-10#list-repositories-accessible-to-the-app-installation). This is the correct endpoint when the `Installation` was created from an IAT (or an app, which pygithub handles and upgrades internally).

However if the Installation was obtained via a user access token e.g. `AuthenticatedUser.get_installations`, then this triggers an HTTP 403 error, the correct endpoint is [`/user/installations/{installation_id}/repositories`](https://docs.github.com/en/rest/apps/installations?apiVersion=2026-03-10#list-repositories-accessible-to-the-user-access-token).

Sadly one can not even use the `repositories_url` which is returned as part of the installation object, because it's apparently hardcoded to `/installation/repositories` (note that it is *not* currently used by the code anyway, `get_repos` hard-codes that URL separately).

I can see two places to handle this, both by checking if the auth is `AppUserAuth`:

- lazily in `get_repos`, and switch to the user version
- or eagerly in `__init__`, after checking for `AppAuth` check for `AppUserAuth`, and update the `_repositories_url` (if necessary, in case github eventually fixes this?), and update `get_repos` to access `self.repostories_url`

hansu650:
## Summary

Fixes #3546.

`Installation.get_repos()` always calls `/installation/repositories`. That endpoint requires installation authentication, so it is unsuitable for an installation obtained through `Github(auth=AppUserAuth(...)).get_user().get_installations()`.

Select `/user/installations/{installation_id}/repositories` when the requester uses `AppUserAuth`, as specified by [GitHub's user-access-token endpoint](https://docs.github.com/en/rest/apps/installations#list-repositories-accessible-to-the-user-access-token). Keep the existing [installation-access-token endpoint](https://docs.github.com/en/rest/apps/installations#list-repositories-accessible-to-the-app-installation) for other authentication types, including the existing `AppAuth` to `AppInstallationAuth` conversion.

This is the lazy endpoint-selection option described in the issue. It does not modify the API-provided `repositories_url` attribute, inspect token strings, or add a public API or dependency. An opaque `Auth.Token` retains its existing behavior; user-access-token routing is selected with `AppUserAuth`.

## Regression coverage

Three new HTTP-mocked cases cover:

- User authentication through `get_user().get_installations()` on both GitHub.com and a custom Enterprise-style API URL with a non-default port.
- Lazy repository retrieval, two-page iteration, `totalCount`, and preservation of authentication, preview headers, and the selected API version.
- An empty repository list when an installation token is supplied through `Auth.Token`.

The existing installation-authentication tests remain unchanged. With the final tests against unmodified upstream `5cdf6ab395c52efb652d0bc1f28fdf2e452b1b41`, the module reports **2 failures and 5 passes**: the two user-authentication cases request the wrong endpoint and receive the mocked 403 response. With the fix, **all 7 pass**. These are offline mocks, not newly recorded live API responses; no real credentials were used.

## Validation

Windows 11 x64, dedicated Conda Python 3.13.15 environment.

- `python -m pytest tests --cov=github`: **1,132 passed, 1 skipped, 649 subtests passed**. The skip is the existing Windows `GitReleaseAsset.testDownloadAssetFile` case.
- `python -m mypy --show-column-numbers github tests`: passed, 321 source files.
- Black 23.12.0 and Ruff 0.1.7: passed on all 329 tracked Python files.
- The repository-pinned docformatter, codespell, pyupgrade, and file-format hooks: passed for the changed files.
- OpenAPI index / verb checking and the `Installation` class-sort dry run: passed.
- `python -m build --no-isolation`, `python -m twine check dist/*`, `python -m pip check`, and `git diff --check`: passed.

Strict Sphinx HTML builds (`-W --keep-going`) generated the documentation but both the patch and unmodified upstream reported 26 warnings. Their warning messages match after normalizing source paths, and none refer to `Installation.py`; a warning-free documentation build is not claimed. Other Python/OS combinations, live GitHub App authentication, and the separate OpenAPI CLI test suite were not run locally.

AI assistance: this patch and its tests were prepared with Codex. The verification results above are from actual local runs.


EnricoMi:
@hansu650 please consider merging https://github.com/hansu650/PyGithub/pull/1 into your branch.

PR #3572, opened 2026-09-08
Respect application logging configuration by default

### LINKED ISSUE #3561: Logging does not inherit the root logger's configuration
The following lines do not allow the `github` logger to let the root logger propagate all of its configuration to it.
They also cause messages to be directly printed to stderr, when one might not necessarily want that behaviour (e.g. letting `rich` format log messages before printing to stderr).
https://github.com/PyGithub/PyGithub/blob/5cdf6ab395c52efb652d0bc1f28fdf2e452b1b41/github/__init__.py#L75-L78

This can be worked around by doing the following:
```python
import logging

import github


logging.getLogger('github').handlers.clear()
logging.getLogger('github').setLevel(logging.NOTSET)
```

vzer200:
Importing PyGithub currently sets the `github` logger to INFO and attaches a stderr handler. That suppresses DEBUG messages requested by the application and can emit duplicate, unformatted messages even when the root logger is configured differently.

Use a `NullHandler` without overriding the logger level, so application logging settings control PyGithub output. Keep `enable_console_debug_logging()` working by adding its stream handler explicitly, and reuse that handler on repeated calls.

Add isolated-process regression tests for configuration before and after import, explicit logger settings, silent unconfigured logging, and repeated console-debug setup. Document the default behavior and correct the helper's output stream description.

Validation on Python 3.12.14:
- `pytest tests -q`: 1,134 passed, 1 skipped, 651 subtests passed.
- `mypy github tests`: no issues in 321 source files.
- All pre-commit hooks for the changed files and `git diff --check` pass.

Fixes #3561.


EnricoMi (approved):
LGTM!

